PR #27, opened 2015-03-07
Add support for ALPN for server and client side + support for NPN client...

normanmaurer:
... side

Motivation:

There is currently no support for ALPN, even now that openssl 1.0.2 supports it. This make it impossible to use it for example with HTTP2.
NPN was only supported when using in server mode, while we support client mode in general now.

Modification:

- Add support for ALPN support when openssl >= 1.0.2 (server and client)
- Add support for NPN when in client mode
- Respect failure behaviour

Result:

It's now possible to use ALPN / NPN on server and client side

### DIFF
--- a/src/main/c/ssl.c
+++ b/src/main/c/ssl.c
@@ -1394,7 +1394,7 @@ TCN_IMPLEMENT_CALL(jstring, SSL, getNextProtoNegotiated)(TCN_STDARGS,
 
     if (ssl_ == NULL) {
         tcn_ThrowException(e, "ssl is null");
-        return 0;
+        return NULL;
     }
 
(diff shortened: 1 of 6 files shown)


normanmaurer:
@trustin please review... @jpinner may also be interesting for you guys at twitter :)

normanmaurer:
Will open another PR which depends on this one that make use of it in netty

trustin:
Looks good to me.

normanmaurer:
@trustin thanks.. cherry-picked as 9720e5350216ecff0166a25e13b4e8b966e7758b

Scottmitch:
+1 good work @normanmaurer

PR #28, opened 2015-03-07
Fix memory leak due not release global reference when using a verifier

normanmaurer:
Motivation:

We miss to delete the global reference to the verifier when free the ssl context.
This leads to a small memory leak,

Modifications:

- Store the reference to the verifier in th tcn_ssl_ctxt_t structure
- Correctly call DeleteGlobalRef when the ssl context is destroyed

Result:

No more memory leak when a verifier is used and the ssl context is destroyed.

### DIFF
--- a/src/main/c/ssl_private.h
+++ b/src/main/c/ssl_private.h
@@ -268,6 +268,10 @@ struct tcn_ssl_ctxt_t {
 
     unsigned char   *next_proto_data;
     unsigned int    next_proto_len;
+
+    /* certificate verifier callback */
+    jobject verifier;
+    jmethodID verifier_method;
 };
 
   
--- a/src/main/c/sslcontext.c
+++ b/src/main/c/sslcontext.c
@@ -34,6 +34,8 @@ static jclass byteArrayClass;
 static apr_status_t ssl_context_cleanup(void *data)
 {
     tcn_ssl_ctxt_t *c = (tcn_ssl_ctxt_t *)data;
+    JNIEnv *e;
+
     if (c) {
         int i;
         if (c->crl)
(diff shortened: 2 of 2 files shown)


normanmaurer:
@trustin please review

src/main/c/sslcontext.c
```diff
@@ -1204,11 +1203,10 @@ TCN_IMPLEMENT_CALL(void, SSLContext, setCertVerifyCallback)(TCN_STDARGS, jlong c
         if (method == NULL) {
             return;
         }
-        ver = malloc(sizeof(struct cert_verifier));
-        ver->verifier = (*e)->NewGlobalRef(e, verifier);
-        ver->method = method;
+        c->verifier = (*e)->NewGlobalRef(e, verifier);
+        c->verifier_method = method;
```
trustin: What happens if a user calls `setCertVerifyCallback()` multiple times? Shouldn't we release the old callback?

src/main/c/ssl_private.h
```diff
@@ -268,6 +268,10 @@ struct tcn_ssl_ctxt_t {
 
     unsigned char   *next_proto_data;
     unsigned int    next_proto_len;
+
+    /* verifier callback */
```
trustin: certificate verifier callback?

src/main/c/sslcontext.c
normanmaurer: Good catch.

normanmaurer:
@trustin thanks for the review.. Addressed comments and cherry-picked 4c52cc917b64e336fe2258baf29067bf309e9cbe

PR #30, opened 2015-03-13
Remove duplicated definition

normanmaurer:
Motivation:

By mistake we duplicated the definition of a function.

Modifications:

Remove duplicate and so allow the compilation to sucess.

Result:

Be able to compile if HAVE_OPENSSL is not set


normanmaurer:
Cherry-picked as 89792a5457177bd9b2138913731db1ff1a9655ab

Issue #31, opened 2015-03-14
NPN/ALPN protocol selection preference reversed


Scottmitch:
The selection preference for the ALPN/NPN protocol negotiation process done by [selectProtocols](https://github.com/netty/netty-tcnative/blob/1.1.32/src/main/c/sslutils.c#L766) should be select the first local "supported" protocols that the peer advertises per [alpn spec](http://tools.ietf.org/html/rfc7301#section-3.2) and [npn tech spec - Protocol Negotiation Section](https://technotes.googlecode.com/git/nextprotoneg.html).  The iteration order is currently reversed.

Scottmitch:
@normanmaurer - Oops...sry :)

Scottmitch:
See PR https://github.com/netty/netty-tcnative/pull/32

normanmaurer:
Fixed by #32 (f5937b6b7f3473643c18653181447a0fc43e8d4e)

PR #32, opened 2015-03-14
ALPN/NPN Selection Order Reversed

Scottmitch:
Motiviation:
The peer doing the selection for ALPN/NPN should choose the first protocol that it prefers instead of the first protocol its peer prefers.  See http://tools.ietf.org/html/rfc7301#section-3.2 and https://technotes.googlecode.com/git/nextprotoneg.html.

Modifications:
- Reverse the looping order in sslutils::select_next_proto.

Result:
Application protocol selection is compliant with the specs.


Scottmitch:
@normanmaurer FYI. Fixes https://github.com/netty/netty-tcnative/issues/31.

normanmaurer:
@Scottmitch good catch... Cherry-picked as f5937b6b7f3473643c18653181447a0fc43e8d4e

Issue #24, opened 2015-02-13
Detect correct JNI library at runtime instead of build time


cconroy:
Rather than try to detect the OS at build time to determine which JNI library to place in META-INF/native, it would be better to package all of them and then let the NativeResourceLoader detect the OS/arch at runtime in order to load the appropriate library.

Detecting at build time is problematic because the build machine may not share the same OS/arch as the production environment. Also, this allows one artifact to be shared across multiple architectures.

What do you think?

normanmaurer:
@cconroy so just to be sure I understand it right... you propose to build all of them, package them together in the jar (and so just produce one jar). Right ?

cconroy:
@normanmaurer exactly. If that sounds good to you, I can try to spin a patch.

normanmaurer:
@cconroy I think this sounds great! I'm just not sure how we could package this stuff correctly while mvn release:perform. As we would need to compile on different platforms before we could build the jar. Any idea?

cconroy:
I figure trying to wrangle cross compilation is not worth the effort / wouldn't work for all platforms.

We could make a separate artifact for each native library. The build machines for those artifacts would need to match their target. 

Then, netty-tcnative would depend on all of those native artifacts to package them all up.

normanmaurer:
@cconroy so you mean we would have some kind of an "uber-jar" that depends on all of the different jars (for different platforms etc) and so pull everything in ?

cconroy:
@normanmaurer yep!

my plan is: the io.netty:netty-tcnative artifact will depend on the arch-specific artifacts. We can drop all native libs into META-INF/native and then update the native resource loader to examine the platform at runtime and pull in the right library.

trustin:
Even if you have built your application on a different platform than your production platform, your application built on a non-production platform should run without a problem on a production platform.

You can just use the netty-tcnative JAR file with the proper classifier as the production dependency, and your problem is gone.

If I understood how your application is deployed, I could have understood your proposal better. Could you elaborate so we learn why it is worth changing how we distribute our native libraries?

cconroy:
@trustin the use case here is that I'd like to share the same artifact across multiple platforms. we have a mix of architectures that might build an artifact and it then might run on yet a different mix of platforms. For example, a developer might want to build an artifact locally to debug (e.g. on OSX), but then also deploy the same artifact to a staging machine in a datacenter (linux).

without such a change I would need to either 
- bundle all of them in a custom/forked way
[...]

PR #33, opened 2015-03-17
Shade platform specific libraries into a single jar

cconroy:
- Split architecture specific libraries into their own artifacts with
  os-arch classifiers.
- Store libs under META-INF/native/${os.classifier}/
- Creates a "java" classifier for the Java JNI bindings
- Shade all supported platforms together with Java source for the main
  artifact.

https://github.com/netty/netty-tcnative/issues/24

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,14 @@
   <name>Netty/TomcatNative</name>
   <url>https://github.com/netty/netty-tcnative/</url>
   <description>
-    A Mavenized fork of Tomcat Native which incorporates various patches
+    A Mavenized fork of Tomcat Native which incorporates various patches.
+
+    Building this artifact produces a native library with a classifier matching the
+    os/arch of the build machine. The Java JNI bindings are exported with the 'java' classifier.
+
+    Prefer specifying a dependency on the shaded artifact (no classifier) which contains
+    the java bindings and prebuilt native libraries for all supported platforms.
+
   </description>
 
   <scm>
@@ -135,16 +142,41 @@
             </goals>
             <configuration>
               <target>
-                <copy todir="${nativeJarWorkdir}">
-                  <zipfileset src="${defaultJarFile}" />
-                </copy>
                 <copy todir="${nativeJarWorkdir}" includeEmptyDirs="false">
                   <zipfileset dir="${nativeLibOnlyDir}/META-INF/native" />
-                  <regexpmapper handledirsep="yes" from="^(?:[^/]+/)*([^/]+)$" to="META-INF/native/\1" />
+                  <regexpmapper handledirsep="yes" from="^(?:[^/]+/)*([^/]+)$" to="META-INF/native/${os.detected.classifier}/\1" />
                 </copy>
-                <jar destfile="${nativeJarFile}" manifest="${nativeJarWorkdir}/META-INF/MANIFEST.MF" basedir="${nativeJarWorkdir}" index="true" excludes="META-INF/MANIFEST.MF,META-INF/INDEX.LIST" />
+                <jar destfile="${nativeJarFile}" manifest="${project.build.directory}/classes/META-INF/MANIFEST.MF" basedir="${nativeJarWorkdir}" index="true" excludes="META-INF/INDEX.LIST"/>
                 <attachartifact file="${nativeJarFile}" classifier="${os.detected.classifier}" type="jar" />
               </target>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+
+      <plugin>
+        <artifactId>maven-jar-plugin</artifactId>
+        <executions>
+          <execution>
+            <!-- Produce a jar with a platfrom-specific classifier. -->
+            <id>default-jar</id>
+            <goals>
+              <goal>jar</goal>
+            </goals>
+            <phase>package</phase>
+            <configuration>
+              <classifier>${os.detected.classifier}</classifier>
+            </configuration>
+          </execution>
+          <execution>
+            <!-- Also produce a jar with 'java' classifier. This can be built on any platform. -->
+            <id>java-jar</id>
+            <goals>
+              <goal>jar</goal>
+            </goals>
+            <phase>package</phase>
+            <configuration>
+              <classifier>java</classifier>
             </configuration>
           </execution>
         </executions>
--- a/shaded/pom.xml
+++ b/shaded/pom.xml
@@ -1,0 +1,116 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright 2015 The Netty Project
+  ~
+  ~ The Netty Project licenses this file to you under the Apache License,
+  ~ version 2.0 (the "License"); you may not use this file except in compliance
+  ~ with the License. You may obtain a copy of the License at:
+  ~
+  ~   http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+  ~ WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+  ~ License for the specific language governing permissions and limitations
+  ~ under the License.
+  -->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>io.netty</groupId>
+    <artifactId>netty-parent</artifactId>
+    <version>4.0.18.Final</version>
+  </parent>
+  <artifactId>netty-tcnative</artifactId>
+  <version>1.1.32.Fork2-SNAPSHOT</version>
+  <packaging>jar</packaging>
+  <properties>
+    <netty-tcnative.version>1.1.32.Fork2-SNAPSHOT</netty-tcnative.version>
+  </properties>
+
+  <name>Netty/TomcatNative</name>
+  <url>https://github.com/netty/netty-tcnative/</url>
+  <description>
+    A Mavenized fork of Tomcat Native which incorporates various patches.
+
+    This shaded jar contains the JNI Java bindings and the prebuilt native libraries for
+    all supported platforms.
+  </description>
+
+  <scm>
+    <url>https://github.com/netty/netty-tcnative</url>
+    <connection>scm:git:git://github.com/netty/netty-tcnative.git</connection>
+    <developerConnection>scm:git:ssh://git@github.com/netty/netty-tcnative.git</developerConnection>
+    <tag>HEAD</tag>
+  </scm>
+
+  <dependencies>
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-tcnative</artifactId>
+      <classifier>java</classifier>
+      <version>${netty-tcnative.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-tcnative</artifactId>
+      <classifier>osx-x86_64</classifier>
+      <version>${netty-tcnative.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-tcnative</artifactId>
+      <classifier>linux-x86_64</classifier>
+      <version>${netty-tcnative.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-tcnative</artifactId>
+      <classifier>windows-x86_64</classifier>
+      <version>${netty-tcnative.version}</version>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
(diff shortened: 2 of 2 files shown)


cconroy:
Companion netty PR: https://github.com/netty/netty/pull/3502

pom.xml
```diff
@@ -28,7 +28,14 @@
   <name>Netty/TomcatNative</name>
   <url>https://github.com/netty/netty-tcnative/</url>
   <description>
-    A Mavenized fork of Tomcat Native which incorporates various patches
+    A Mavenized fork of Tomcat Native which incorporates various patches.
+
+    Building this artifact produces a native library with a classifier matching the
+    os/arch of the build machine. (The Java JNI bindings are exported with the 'java' classifier)
```
normanmaurer: move the (...) stuff before the `.`

pom.xml
```diff
@@ -150,6 +154,35 @@
         </executions>
       </plugin>
 
+      <plugin>
+        <artifactId>maven-jar-plugin</artifactId>
+        <executions>
+          <execution>
+            <!-- Produce a jar with a platfrom-specific classifier-->
```
normanmaurer: missing space before `-->`

normanmaurer: Missing . on EOL

pom.xml
```diff
@@ -150,6 +154,35 @@
...
+            <goals>
+              <goal>jar</goal>
+            </goals>
+            <phase>package</phase>
+            <configuration>
+              <classifier>${os.detected.classifier}</classifier>
+            </configuration>
+          </execution>
+          <execution>
+            <!-- Also produce a jar with platfrom-specific classifier. This can be built on any platform -->
```
normanmaurer: Missing `.` on EOL

shaded/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+  ~ WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+  ~ License for the specific language governing permissions and limitations
+  ~ under the License.
+  -->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>io.netty</groupId>
+    <artifactId>netty-parent</artifactId>
+    <version>4.0.18.Final</version>
```
normanmaurer: I wonder if we want to pump this to 4.0.26.Final ?

shaded/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+  <version>1.1.32.Fork2-SNAPSHOT</version>
+  <packaging>jar</packaging>
+  <properties>
+    <netty-tcnative.version>1.1.32.Fork2-SNAPSHOT</netty-tcnative.version>
+  </properties>
+
+  <name>Netty/TomcatNative</name>
+  <url>https://github.com/netty/netty-tcnative/</url>
+  <description>
+    A Mavenized fork of Tomcat Native which incorporates various patches
```
normanmaurer: Missing `.` on EOL

shaded/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+              <goal>shade</goal>
+            </goals>
+            <configuration>
+              <createSourcesJar>true</createSourcesJar>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+
+
```
normanmaurer: Kill two empty lines

shaded/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+
+      <!-- Override the parent POM's configuration -->
```
normanmaurer: Missing `.` on EOL

shaded/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+      <plugin>
+        <artifactId>maven-release-plugin</artifactId>
+        <configuration>
+          <arguments>-P restricted-release,sonatype-oss-release</arguments>
+          <tagNameFormat>@{project.artifactId}-@{project.version}</tagNameFormat>
+        </configuration>
+      </plugin>
+    </plugins>
+
```
normanmaurer: Kill empty line

pom.xml
```diff
@@ -150,6 +154,35 @@
...
+            <id>java-jar</id>
+            <goals>
+              <goal>jar</goal>
+            </goals>
+            <phase>package</phase>
+            <configuration>
+              <classifier>java</classifier>
+            </configuration>
+          </execution>
```
normanmaurer: Kill empty line

normanmaurer:
@cconroy please address my comments

pom.xml
cconroy: actually, i'll just remove the parens

pom.xml
cconroy: fixed. and also updated wording (this is the 'java' classifier here)

shaded/pom.xml
cconroy: i just copied the existing parent version. let me know what you want here

shaded/pom.xml
cconroy: fixed

shaded/pom.xml
cconroy: fixed

shaded/pom.xml
cconroy: fixed

pom.xml
cconroy: fixed

cconroy:
@trustin @normanmaurer one thing I meant to call out: In our discussion on the issue @trustin noted that it would make sense to keep the java classes in each architecture specific artifact to support the non-shaded use case (e.g. if a user only wants to pull in one platform). 

In this PR, I have split java out into a separate artifact classifier that is built on any platform. This should address that use case while avoiding duplicate class files.

pom.xml
```diff
@@ -150,6 +154,34 @@
...
+            <goals>
+              <goal>jar</goal>
+            </goals>
+            <phase>package</phase>
+            <configuration>
+              <classifier>${os.detected.classifier}</classifier>
+            </configuration>
+          </execution>
+          <execution>
+            <!-- Also produce a jar with 'java' classifier. This can be built on any platform. -->
```
trustin: I thought we build the JAR without native libraries with no classifier already? Could you explain why we need this?

pom.xml
```diff
@@ -150,6 +154,34 @@
         </executions>
       </plugin>
 
+      <plugin>
+        <artifactId>maven-jar-plugin</artifactId>
+        <executions>
+          <execution>
+            <!-- Produce a jar with a platfrom-specific classifier. -->
```
trustin: We actually generate and attach the JAR with the native library using `maven-antrun-plugin` above. I guess we don't need it.

trustin:
I like the idea of providing a shaded JAR that contains all native libraries. How about:

* Make the `netty-tcnative` jar with no classifier (the default jar) include all supported native libraries. i.e. most users can just pull the default JAR and that should be all.
* Keep the JARs with the OS classifiers as before.

One problem I see with the `shade/pom.xml` is that the maven release plugin will not update the version number of `shade/pom.xml`. How would you fix this problem?

pom.xml
cconroy: The antrun plugin generates the jar, but this execution attaches the os classifier. I should probably update the comment to read "Tag the jar with a platform-specific classifier" ?

cconroy: More accurately: "Make the default target the platform-specific jar: only the shade plugin should produce a package without a classifier"

pom.xml
cconroy: Under this model, the non-classified jar is produced by the shade pom to produce a jar that can be used on any platform. Here, the antrun plugin produces the platform-specific jar. This extra execution tells maven to package up the normal target results as a 'java' package.

cconroy:
@trustin: i'm not too familiar with the maven release plugin. Unfortunately, maven does not let me use a property as a version tag. I'll dig around to see what I can find... At the very least we could have a clear README with release steps or perhaps a shell script to automate?

cconroy:
"Make the netty-tcnative jar with no classifier (the default jar) include all supported native libraries. i.e. most users can just pull the default JAR and that should be all."

Yep! This is what happens when you run the shaded target. I encourage you to try it out (you can test locally by commenting out the other platforms to just shade your local platform after building the platform-specific libs).

cconroy:
"Keep the JARs with the OS classifiers as before."

IMHO it's cleaner to separate the java and platform-specific artifacts for those users that don't want the shaded jar. I would personally prefer to add a second dep rather than get duplicate class files. 

Historically I've seen build failures and other problems (e.g. with IDEs, other build integration tools) occur  with duplicate classes. Most users will probably not care about a slightly larger jar and will just use the shaded one anyways. We should add documentation to clarify how to depend on a single platform.

If you're okay with keeping them separate, I can update the readme to reflect. If not, I can try to spin a version with duplicate classes.

cconroy:
@trustin: realistically, i don't think you probably want to update the shade and platform specific numbers at the same time. You'll be releasing platform specific libs from the same platform machines. Once all of those are updated, you'll want to do a shade release. I don't see a path for doing this automatically with the release plugin.

trustin:
@cconroy Agreed. The build process of the all-platform JAR will be more like fetching & merging all JARs of the already-deployed JARs. I guess we don't need a pom for this. Just a shell script should do the job.  Let apply your changes partially and write one.

netkins:
**AUTOMATED MESSAGE for ADMINS**:<br>Please verify and accept the pull request.<br>To accept, say `@netkins accept`<br>To build again, say `@netkins build`<br>To whitelist the author, say `@netkins whitelist`

cconroy:
This should be good to merge?

cconroy:
@trustin @normanmaurer ping

cconroy:
@trustin @normanmaurer can we please merge this?

normanmaurer:
Sorry will check tomorrow... Busy as hell :/

trustin:
@cconroy @normanmaurer I'll take care of this as well as the `NativeLibraryLoader` changes in Netty. I don't have time at the moment, either, but I wanna resolve this by myself.  Sorry for the delay and thanks for your patience.

cconroy:
@trustin @normanmaurer ping

shaded/pom.xml
```diff
@@ -0,0 +1,116 @@
...
+      <version>${netty-tcnative.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-tcnative</artifactId>
+      <classifier>osx-x86_64</classifier>
+      <version>${netty-tcnative.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.netty</groupId>qq
```
nmittler: Remove "qq"

nmittler:
@trustin is there anything else to be done? can we cherry-pick this?

normanmaurer:
@nathan I'm busy this week but let me finish this stuff next week

nmittler:
@normanmaurer SGTM ... thanks!

btw ... wrong Nathan :)

shaded/pom.xml
cconroy: fixed! (and pushed with a squash rebase)

cconroy:
pinging again...

cconroy:
bueller?

nmittler:
where bueller = @normanmaurer :)

slandelle:
@normanmaurer @trustin Is there anything I can do here to make this happen?

slandelle:
At the moment, it's impossible to extend something in Netty in order to have this because of the loading strategy is hardcoded in `io.netty.handler.ssl.OpenSsl` and that `io.netty.util.internal.NativeLibraryLoader` always load from `META-INF/native/` where it would be great to be able to pass a subdirecty to isolate native libs for each OS.

normanmaurer:
I will try to fig into this again next week or so... Sorry super busy atm

slandelle:
Great! Let me know if I can help.

louiscryan:
Now that we have the statically linked boring ssl artifacts I'd like to bring up what we think the 'default' artifact should be. We theoretically have

<platforms> X [openssl, libre, boring] X [dynamic, static]

Based on the current plan a mega jar with boring-static is the library most likely to work out of the box for all features for developers when compared to a shared openssl one. Switching to this however can't be done without a certain amount of messaging. 

In particular the lifecycle around managing security patches to openssl interacts with this decision as sys admins may have the latent expectation that updating shared openssl is their primary tool for resolving security issues. I suppose most admins have become used to updating Java to resolve issues too so this may be less of a concern if we were to switch to boring-static as the default but its worth mentioning.

nmittler:
Hey guys,
Based on the recent work for creating statically-linked libraries, I've created a similar PR: #113.

Let's discuss.

trustin:
@nmittler Thank you so much for taking a stab on this!

johnou:
@cconroy now that @nmittler has broken the ice how would you go about adapting this pull request to support other libraries such as openssl dynamic fedora etc. I have an open PR on the netty side here https://github.com/netty/netty/pull/4925

cconroy:
@johnou This pull request is over a year old despite having general agreement on the approach so any other changes should probably be done in a subsequent PR

@normanmaurer  any updates here?

normanmaurer:
@nmittler I think we can close this now as we have the boringssl static uberjar. WDYT ?

nmittler:
Yeah sgtm ... closing.

PR #35, opened 2015-04-08
Use weak definitions to make it usuable with openssl 1.0.1 and openssl 1...

normanmaurer:
....0.2

Motivation:

To be able to support alpn when its avaible in the used openssl version we should use weak definitions.

Modifications:

Use weak definations for the needed openssl calls

Results:

Can now use alpn if supported on the openssl version while still work correctly if not avaible.

### DIFF
--- a/src/main/c/ssl.c
+++ b/src/main/c/ssl.c
@@ -1409,8 +1409,18 @@ TCN_IMPLEMENT_CALL(jstring, SSL, getNextProtoNegotiated)(TCN_STDARGS,
 
 TCN_IMPLEMENT_CALL(jstring, SSL, getAlpnSelected)(TCN_STDARGS,
                                                          jlong ssl /* SSL * */) {
-    // Only supported with openssl >= 1.0.2
-    #if OPENSSL_VERSION_NUMBER >= 0x10002000L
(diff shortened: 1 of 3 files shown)


normanmaurer:
@trustin please review... With this we can support different openssl versions at the same time on systems where we use GCC (unix/osx/bsd/linux) by just set LD_LIBRARY_PATH correctly before run the app. On windows we need to ensure to compile with the correct openssl version though as virtual studio not support weak stuff. I think this is not a beak deal though.

WDYT ?

normanmaurer:
@Scottmitch FYI :+1:

Scottmitch:
good old pre-compiler flags to the rescue...typically not a fan but in this case I think it is the right move +1.

Scottmitch:
great work BTW....linking/compiling across different systems with libraries out of your control is rarely enjoyable :(

normanmaurer:
@Scottmitch ha yeah... all the holes in my desk from banging :100:

normanmaurer:
Cherry-picked into 1.1.32 (45c4648072f7adfc178c19409c95508e827becaf) and 1.1.33 (2d291d24a803ffbabd9ed0f9d0113d94816be3e0)

Issue #36, opened 2015-04-27
NoClassDefFoundError when building tcnative on Ubuntu


doom369:
Hi,

I was following instructions from wiki

```
sudo apt-get install autoconf automake libtool make tar libapr1-dev libssl-dev
git clone https://github.com/netty/netty-tcnative.git
cd netty-tcnative
git checkout netty-tcnative-1.1.32.Fork1
mvn clean install
```
And getting :
```
[INFO] os.detected.name: linux
[INFO] os.detected.arch: x86_64
[INFO] os.detected.classifier: linux-x86_64


Exception in thread "main" java.lang.NoClassDefFoundError: org/eclipse/aether/RepositorySystemSession
	at kr.motd.maven.os.DetectExtension.injectSession(DetectExtension.java:115)
	at kr.motd.maven.os.DetectExtension.afterProjectsRead(DetectExtension.java:92)
	at org.apache.maven.DefaultMaven.doExecute(DefaultMaven.java:274)
	at org.apache.maven.DefaultMaven.execute(DefaultMaven.java:156)
	at org.apache.maven.cli.MavenCli.execute(MavenCli.java:537)
	at org.apache.maven.cli.MavenCli.doMain(MavenCli.java:196)
	at org.apache.maven.cli.MavenCli.main(MavenCli.java:141)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at org.codehaus.plexus.classworlds.launcher.Launcher.launchEnhanced(Launcher.java:289)
	at org.codehaus.plexus.classworlds.launcher.Launcher.launch(Launcher.java:229)
	at org.codehaus.plexus.classworlds.launcher.Launcher.mainWithExitCode(Launcher.java:415)
	at org.codehaus.plexus.classworlds.launcher.Launcher.main(Launcher.java:356)
Caused by: java.lang.ClassNotFoundException: org.eclipse.aether.RepositorySystemSession
	at org.codehaus.plexus.classworlds.strategy.SelfFirstStrategy.loadClass(SelfFirstStrategy.java:50)
[...]

PR #37, opened 2015-04-29
[#36] Enforce usage maven 3.1.1+

normanmaurer:
Motivation:

When using maven < 3.1 you see the follow exception:

Exception in thread "main" java.lang.NoClassDefFoundError: org/eclipse/aether/RepositorySystemSession

Modifications:

Enforce using maven 3.1.1+ so the user not get the "cryptic" exception

Result:

Better expierence for developers that try to build netty-tcnative.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -158,6 +158,24 @@
           <tagNameFormat>@{project.artifactId}-@{project.version}</tagNameFormat>
         </configuration>
       </plugin>
+      <plugin>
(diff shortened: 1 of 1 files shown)


normanmaurer:
Cherry-picked into 1.1.32 (aebc03c819dd09df3de7efa28a6239133cefb1e8) and 1.1.33 (0abce06b39fabf1a90df97f420100414fcaaebb7)

PR #38, opened 2015-05-05
Allow to retrieve the number of renegations for a ssl instance

normanmaurer:
Motivation:

To allow rejection client initiated renegotiations we need a way to obtain the number of previous done renegations.

Modifications:

Add SSL.getRenegations(...).

Result:

It's now possible to obtain the number of renegations.

### DIFF
--- a/src/main/c/ssl.c
+++ b/src/main/c/ssl.c
@@ -1177,6 +1177,16 @@ TCN_IMPLEMENT_CALL(jint, SSL, getLastErrorNumber)(TCN_STDARGS) {
     return ERR_get_error();
 }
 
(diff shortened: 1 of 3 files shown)


normanmaurer:
@trustin please review

trustin:
I guess `ctx->renegotiation` in increased on the first successful handshake as well as renegotiations. Correct?  If so, how about renaming `renegotiation` to `handshakeCount` and `getRenegotiations()` to `getHandshakeCount()`?

normanmaurer:
@trustin yes... Let me do this. And then cherry-pick.

normanmaurer:
Cherry-picked into 1.1.33 as 1cd48327e711a230e325f74d3d09873d70043edd

PR #39, opened 2015-05-06
Allow to correctly retrieve the number of handshakes for a ssl instance

normanmaurer:
Motivation:

In commit 1cd48327e711a230e325f74d3d09873d70043edd SSL.getHandshakeCount(...) was introduced. Unfortunally the implementation was not correct and so the returned number did not match the reality.

Modifications:

- Store the handshakeCount per SSL instance and so fix the implemented

Result:

SSL.getHandshakeCount(...) returns the correct number now.

### DIFF
--- a/src/main/c/ssl.c
+++ b/src/main/c/ssl.c
@@ -719,8 +719,8 @@ TCN_IMPLEMENT_CALL(jint, SSL, initialize)(TCN_STDARGS, jstring engine)
      * low entropy seed.
      */
     SSL_rand_seed(NULL);
-    /* For SSL_get_app_data2() at request time */
-    SSL_init_app_data2_idx();
+    /* For SSL_get_app_data2() and SSL_get_app_data3() at request time */
+    SSL_init_app_data2_3_idx();
 
     SSL_TMP_KEYS_INIT(r);
     if (r) {
@@ -1178,11 +1178,11 @@ TCN_IMPLEMENT_CALL(jint, SSL, getLastErrorNumber)(TCN_STDARGS) {
 }
 
 static void ssl_info_callback(const SSL *ssl, int where, int ret) {
-    tcn_ssl_ctxt_t *c;
+    uint16_t *handshakeCount = NULL;
     if (0 != (where & SSL_CB_HANDSHAKE_START)) {
-        c = SSL_get_app_data2(ssl);
-        if (c != NULL) {
-            ++c->handshakeCount;
+        handshakeCount = (uint16_t*) SSL_get_app_data3(ssl);
+        if (handshakeCount != NULL) {
+            ++(*handshakeCount);
         }
     }
 }
@@ -1191,6 +1191,7 @@ TCN_IMPLEMENT_CALL(jlong /* SSL * */, SSL, newSSL)(TCN_STDARGS,
                                                    jlong ctx /* tcn_ssl_ctxt_t * */,
                                                    jboolean server) {
     tcn_ssl_ctxt_t *c = J2P(ctx, tcn_ssl_ctxt_t *);
+    uint16_t *handshakeCount = malloc(sizeof(uint16_t));
     SSL *ssl;
 
     UNREFERENCED_STDARGS;
@@ -1201,6 +1202,11 @@ TCN_IMPLEMENT_CALL(jlong /* SSL * */, SSL, newSSL)(TCN_STDARGS,
         tcn_ThrowException(e, "cannot create new ssl");
         return 0;
     }
+
+    // Store the handshakeCount in the SSL instance.
+    *handshakeCount = 0;
+    SSL_set_app_data3(ssl, handshakeCount);
+
     // Add callback to keep track of handshakes.
     SSL_CTX_set_info_callback(c->ctx, ssl_info_callback);
 
@@ -1313,7 +1319,14 @@ TCN_IMPLEMENT_CALL(void, SSL, freeSSL)(TCN_STDARGS,
                                        jlong ssl /* SSL * */) {
     UNREFERENCED_STDARGS;
 
-    SSL_free(J2P(ssl, SSL *));
+    uint16_t *handshakeCount = NULL;
+    SSL *ssl_ = J2P(ssl, SSL *);
+
+    handshakeCount = SSL_get_app_data3(ssl_);
+    if (handshakeCount != NULL) {
+        free(handshakeCount);
+    }
+    SSL_free(ssl);
 }
 
 // Make a BIO pair (network and internal) for the provided SSL * and return the network BIO
@@ -1751,7 +1764,7 @@ TCN_IMPLEMENT_CALL(jbyteArray, SSL, getSessionId)(TCN_STDARGS, jlong ssl)
 
 TCN_IMPLEMENT_CALL(jint, SSL, getHandshakeCount)(TCN_STDARGS, jlong ssl)
 {
-    tcn_ssl_ctxt_t *c;
+    uint16_t *handshakeCount = NULL;
     SSL *ssl_ = J2P(ssl, SSL *);
     if (ssl_ == NULL) {
         tcn_ThrowException(e, "ssl is null");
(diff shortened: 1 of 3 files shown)


src/main/c/ssl.c
```diff
@@ -719,8 +719,8 @@ TCN_IMPLEMENT_CALL(jint, SSL, initialize)(TCN_STDARGS, jstring engine)
      * low entropy seed.
      */
     SSL_rand_seed(NULL);
-    /* For SSL_get_app_data2() at request time */
-    SSL_init_app_data2_idx();
+    /* For SSL_get_app_data2() and SSL_get_app_data3() at request time */
```
Scottmitch: Do we comment anywhere what is intended to be stored in either position?  Seems like this case the `data2` and `data3` are intended to be mutually exclusive?

src/main/c/ssl.c
```diff
@@ -1178,11 +1178,11 @@ TCN_IMPLEMENT_CALL(jint, SSL, getLastErrorNumber)(TCN_STDARGS) {
 }
 
 static void ssl_info_callback(const SSL *ssl, int where, int ret) {
-    tcn_ssl_ctxt_t *c;
+    int *handshakeCount;
```
Scottmitch: `= NULL`?

src/main/c/ssl.c
normanmaurer: actually data2 is used by tcnative and data3 I just added. Let me add a comment in the code

src/main/c/ssl.c
```diff
@@ -1191,6 +1191,7 @@ TCN_IMPLEMENT_CALL(jlong /* SSL * */, SSL, newSSL)(TCN_STDARGS,
                                                    jlong ctx /* tcn_ssl_ctxt_t * */,
                                                    jboolean server) {
     tcn_ssl_ctxt_t *c = J2P(ctx, tcn_ssl_ctxt_t *);
+    int * handshakeCount = malloc(sizeof(int));
```
normanmaurer: will remove space between * and handshakeCount

src/main/c/ssl.c
normanmaurer: I don't think we need to init with NULL. Do we ?

src/main/c/ssl.c
```diff
@@ -1178,11 +1178,11 @@ TCN_IMPLEMENT_CALL(jint, SSL, getLastErrorNumber)(TCN_STDARGS) {
...
 
 static void ssl_info_callback(const SSL *ssl, int where, int ret) {
-    tcn_ssl_ctxt_t *c;
+    int *handshakeCount;
     if (0 != (where & SSL_CB_HANDSHAKE_START)) {
-        c = SSL_get_app_data2(ssl);
-        if (c != NULL) {
-            ++c->handshakeCount;
+        handshakeCount = (int*) SSL_get_app_data3(ssl);
+        if (handshakeCount != NULL) {
```
normanmaurer: @Scottmitch btw do you prefer `if (handshakeCount != NULL)` or `if (handshakeCount)` ?

src/main/c/ssl.c
Scottmitch: Is `uint32_t` (or `uint16_t`) a better choice for a count?  Are we intending to use negative numbers?

src/main/c/ssl.c
Scottmitch: `!= NULL`. Lets not confuse pointers with boolean expressions :)

src/main/c/ssl.c
Scottmitch: Don't need to no.  But if you were to use it with out "really" initializing it you would get a quick stack trace....as opposed to accepting what ever gets put in there and potentially dereferencing who knows what.  This block of code is good as it is initialized before it is used...but more just good practice (declaration is initialization).

src/main/c/ssl.c
normanmaurer: uint16_t may be a good choice… Good idea!

normanmaurer:
@Scottmitch ok I think I addressed all of your comments buddy :) Anything else ?

src/main/c/sslutils.c
```diff
@@ -64,13 +64,22 @@ void SSL_init_app_data2_idx(void)
         return;
     }
 
-    /* we _do_ need to call this twice */
-    for (i = 0; i <= 1; i++) {
+    /* we _do_ need to call this two times */
```
Scottmitch: 0, 1, 2 = three times?

normanmaurer: ups... the comment is correct but the change in the next line not... fixed

src/main/c/sslutils.c
```diff
@@ -84,6 +93,17 @@ void SSL_set_app_data2(SSL *ssl, void *arg)
     return;
 }
 
+void *SSL_get_app_data3(SSL *ssl)
+{
+    return (void *)SSL_get_ex_data(ssl, SSL_app_data3_idx);
```
Scottmitch: Do we need this (void *) cast?

src/main/c/sslutils.c
```diff
@@ -84,6 +93,17 @@ void SSL_set_app_data2(SSL *ssl, void *arg)
...
+}
+
+void SSL_set_app_data3(SSL *ssl, void *arg)
+{
+    SSL_set_ex_data(ssl, SSL_app_data3_idx, arg);
+    return;
```
Scottmitch: remove this return?

src/main/c/ssl.c
```diff
@@ -1178,11 +1178,11 @@ TCN_IMPLEMENT_CALL(jint, SSL, getLastErrorNumber)(TCN_STDARGS) {
...
 static void ssl_info_callback(const SSL *ssl, int where, int ret) {
-    tcn_ssl_ctxt_t *c;
+    uint16_t *handshakeCount = NULL;
     if (0 != (where & SSL_CB_HANDSHAKE_START)) {
-        c = SSL_get_app_data2(ssl);
-        if (c != NULL) {
-            ++c->handshakeCount;
+        handshakeCount = (uint16_t*) SSL_get_app_data3(ssl);
+        if (handshakeCount != NULL) {
+            ++(*handshakeCount);
```
normanmaurer: @Scottmitch btw I think I could remove the `()` around it but I kind of think it is easier to read with them. What you prefer ?

src/main/c/sslutils.c
normanmaurer: yep why not... Just copied the `SSL_set_app_data2(...)` method mostly

src/main/c/sslutils.c
normanmaurer: @Scottmitch looking at `man SSL_get_ex_data ` I think no..

src/main/c/ssl.c
Scottmitch: @normanmaurer - You would be fine but I vote leave it. Prefix inc and dereference are in the same precedence group, but parsing is from right to left...but I think what you have here prevents people from having to double check operator precedence every time they look at it ;)

normanmaurer: ok I think then I'm done :)

src/main/c/ssl.c
```diff
@@ -1201,6 +1202,11 @@ TCN_IMPLEMENT_CALL(jlong /* SSL * */, SSL, newSSL)(TCN_STDARGS,
         tcn_ThrowException(e, "cannot create new ssl");
         return 0;
     }
+
+    // Store the handshakeCount in the SSL instance.
+    *handshakeCount = 0;
```
Scottmitch: nit: do we need to check for a memory allocation failure....we would be pretty hosed at that point anyways but WDYT?

Scottmitch:
@normanmaurer - Just 1 more question then ship it.

src/main/c/ssl.c
normanmaurer: @Scottmitch Hmm just leave it as it is for now. Maybe we should create another issue and go over the whole code and check for these. Then when once is detected throw an exception or so. WDYT ?

Scottmitch: Yes I think that is a good idea in general...like I said if we can't allocate a `uint16_t` we are likely near death anyways :)

normanmaurer: cool :) Let me squash and cherry-pick + open an issue

normanmaurer:
Cherry-picked as 2c17755b6330a6b6f2aa0dcb0ced5bca0ab476b3

PR #42, opened 2015-06-10
Ensure SSLContext.make(...) not fails when using libressl

normanmaurer:
Motivation:

libressl dropped support for export ciphers for which SSL_CTX_set_tmp_rsa_callback is set in SSLContext.make(...). This cause an error when using libressl.

Modifications:

Not call SSL_CTX_set_tmp_rsa_callback when using libressl.

Result:

SSLContext.make(...) does work as expected if libressl is used.


normanmaurer:
@trustin please check... easy as well :)

trustin:
LGTM

normanmaurer:
Cherry-picked as 117d8a3d15acd6e6eae9fcb33dc0cef002393f03

PR #43, opened 2015-06-16
Add SSL.clearError() which allows to clear all pending errors from th…

normanmaurer:
…e threads error queue.

Motivation:

ERR_clear_error() needs to be called before a SSL operation if a failure was detected before. Otherwise it may report incorrect results when calling SSL_get_error(...)

Modifications:

Add SSL.clearError()

Result:

It's now possible to clear the error queue.


normanmaurer:
@trustin please check... should be an easy one

normanmaurer:
@trustin as this was very easy fix I just cherry-picked as 7d603d9b1f322d37a6b00eb8c9213d13fcd7d776

trustin:
LGTM.

PR #48, opened 2015-07-03
Fix segfault when SSL.getSessionId() is called and no SESSION exists.

normanmaurer:
Motivation:

Due a missing NULL check a segfault was produced when try to call SSL.getSessionId() while not SESSION eists.

Modifications:

Add missing NULL check.

Result:

No more segfault possible.


normanmaurer:
@trustin please check

normanmaurer:
Cherry-picked into 1.1.33 as e16ba976214dc61c4b5149473eb8034172f84fb6

Issue #45, opened 2015-06-18
Linux .so isn't stripped


ejona86:
netty-tcnative-1.1.33.Fork2-linux-x86_64.jar is 422.7K, whereas osx-x86_64 and windows-x86_64 are ~125K.

libnetty-tcnative.so is actually 996K in size. Stripping it reduced it to 163K and the JAR itself to 129K.

normanmaurer:
@ejona86 you are talking about gcc -s ?

ejona86:
I was referring to the strip command. But yes, I think gcc -s does the same thing.

Hmm... `strip -g` reduced the .so to 214K. Was it accidentally compiled with `gcc -g`?

normanmaurer:
@ejona86 I will check what our autoconf stuff does here.

ejona86:
I was thinking it was more likely that you had some environment variables set on the build machine. If you had a CXXFLAGS=-g or similar then it would cause that.

normanmaurer:
@ejona86 I will check on a fresh linux install how it looks like

normanmaurer:
@ejona86 I checked and "CFLAGS" includes "-g" in the Makefile which was generated by autoconf. Checking now how to fix this. Stay tuned.

normanmaurer:
@ejona86 ok I think I know how I can fix this... will need to spend some time on it. That said I'm currently a bit tight on time so I guess not before next week.

ejona86:
@normanmaurer, no rush. It isn't hurting much right now. We will want it eventually, since it is a noticeable decrease in dependency weight, but it doesn't block anything.

normanmaurer:
@ejona86 see https://github.com/netty/netty-tcnative/pull/49

normanmaurer:
Fixed by https://github.com/netty/netty-tcnative/pull/49

PR #49, opened 2015-07-06
[#45] Set correct GCC CFLAGS when compile

normanmaurer:
Motivation:

By default autotools use -g when compile via gcc and so add global debug symbols. This makes the compiled file very big. Beside this no GCC optimization settings are used.

Modifications:

Use our own configure.ac file which allows us to not use -g flag and setting -O3 for optimized code.

Result:

Smaller file size due not include debug symbols and faster code due optimizations done by GCC.

### DIFF
--- a/src/main/native-package/configure.ac
+++ b/src/main/native-package/configure.ac
@@ -58,7 +58,7 @@ AC_SUBST(LDFLAGS)
 ## -----------------------------------------------------
 ## Generate the files
 ## -----------------------------------------------------
-AM_INIT_AUTOMAKE([subdir-objects no-dependencies -Wall -Werror foreign])
+AM_INIT_AUTOMAKE([subdir-objects no-dependencies -Wall foreign])
 AC_CONFIG_FILES([Makefile])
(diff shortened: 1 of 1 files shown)


normanmaurer:
@trustin please check... will do some benchmarks in the meantime

nmittler:
yikes :)

ejona86:
-O3. Good idea! :-)

Note that -Werror is frequently looked down upon because different compiler versions have different warnings. So if the user is using a newer or older version of the compiler they frequently will be unable to compile, especially with -Wall. Different header files can also cause unexpected trouble. The general recommendation is that you specify -Werror in your own personal CFLAGS.

trustin:
LGTM. Please address @ejona86's comment before cherry-picking though.

normanmaurer:
Addressed comment and cherry-picked as 14037dec78dd8bb8f7b1261d495f7505bbaa4560 into 1.1.33. 

Did a few benchmarks and found it slightly faster, but not much.

PR #50, opened 2015-07-07
BoringSSL Support

Scottmitch:
Motivation:
BoringSSL is largely API compatible with OpenSSL. BoringSSL may be preferred to OpenSSL in some situations.

Modifications:
- Modify codebase to support BoringSSL while retaining OpenSSL compatibility.

Result:
We can now link against BoringSSL.

### DIFF
--- a/src/main/c/ssl.c
+++ b/src/main/c/ssl.c
@@ -451,7 +451,6 @@ static struct CRYPTO_dynlock_value *ssl_dyn_create_function(const char *file,
 /*
  * Dynamic locking and unlocking function
  */
-
 static void ssl_dyn_lock_function(int mode, struct CRYPTO_dynlock_value *l,
                            const char *file, int line)
 {
@@ -531,8 +530,11 @@ static int ssl_rand_load_file(const char *file)
         file = ssl_global_rand_file;
     if (file && (strcmp(file, "builtin") == 0))
         return -1;
+// BoringSsl doesn't support RAND_file_name, but RAND_status() returns 1 anyways
+#ifndef OPENSSL_IS_BORINGSSL
     if (file == NULL)
         file = RAND_file_name(buffer, sizeof(buffer));
+#endif
     if (file) {
 #ifdef HAVE_SSL_RAND_EGD
         if (strncmp(file, "egd:", 4) == 0) {
@@ -555,9 +557,9 @@ static int ssl_rand_load_file(const char *file)
  */
 static int ssl_rand_save_file(const char *file)
 {
+#ifndef OPENSSL_IS_BORINGSSL
     char buffer[APR_PATH_MAX];
     int n;
-
     if (file == NULL)
         file = RAND_file_name(buffer, sizeof(buffer));
     else if ((n = RAND_egd(file)) > 0) {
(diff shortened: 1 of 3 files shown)


normanmaurer:
@Scottmitch LGTM. Ship it :beers:

Scottmitch:
@nmittler @ejona86 @nicolasnoble - Would you mind taking a peek?  BoringSSL changed the RAND interfaces a bit.  Be good to get someone who is familiar with boring to make sure I did the right thing here.

Scottmitch:
@nicolasnoble - You were referred by @buchgr :)

nicolasnoble:
Most of the time when BoringSSL is not implementing something from OpenSSL, it's because it's a deprecated or dangerous feature. I don't know exactly the scope of the changes, but wouldn't it be worth just killing the features that BoringSSL doesn't support instead ?

Scottmitch:
@nicolasnoble - The intention was to `#ifdef` out all behavior that is no longer needed and return the appropriate return status (similar approach that BoringSSL takes for the peripheral `RAND_*` methods that we are using).  The public interface exposes `ssl_rand_save_file` (which uses a bunch of stuff BoringSSL doesn't support) and is essentially now a NOOP.  The main questions I have for you revolve around the changes in this method and `ssl_rand_load_file`. The question is have I preserved the `RAND_*` functionality given BoringSSL's drop  in support for these `RAND_*file*` methods.

ejona86:
I don't have anything to say about it myself, as I'm not very familiar with BoringSSL (or OpenSSL) from an API perspective. I have started asking around a bit internally to see if someone would be interested taking a glance.

Scottmitch:
@ejona86 - Thanks! I am also not on expert on either and so that is why I want someone to take a closer look...especially since `RAND*` functions are involved :)

ejona86:
I got a response internally. Things look good. Paraphrased: `ssl_rand_save_file` returning 0 makes a lot of sense. `ssl_rand_load_file` is harder to reason about but seems to be doing the best that can be done.

Woo!

normanmaurer:
@ejona86 so this means ready to merge ?

ejona86:
It's fine from my perspective. @Scottmitch may make tweaks based on agl's comments.

I also received a later comment questioning if tcnative would really support BoringSSL after this change. As an example, it seemed odd that `EVP_read_pw_string` was still being used, because that isn't present in BoringSSL.

Scottmitch:
> it seemed odd that EVP_read_pw_string was still being used, because that isn't present in BoringSSL.

@ejona86 - Thanks for bringing this up!  I have since `ifdef`ed out this method call for boringssl.  This patch was used to dynamically link and run an internal application. This application only uses a subset of methods and I have not exhaustively checked every method used by tcnative to see if boringssl still supports it.  If there are other methods you are aware of please bring them up, otherwise I think we can handle those in followup PRs.

normanmaurer:
@Scottmitch ship it

Scottmitch:
@nicolasnoble @agl @ejona86 @normanmaurer - Thanks for review!

Cherry-picked (29b59e9)

PR #53, opened 2015-07-20
Allow to set certificate and certificate chain via BIO.

normanmaurer:
Motiviation:

It's useful to be able to set the certificate and certificate chain via a BIO and not be limited by the need to have it stored in the fs.

Modification:

- Add methods so its possible to set the certificate and certifcate chain via a BIO
- Add method which allows to create a in-memory BIO

Result:

It's now possible to also configure SSLContext via BIOs.

### DIFF
--- a/src/main/c/sslcontext.c
+++ b/src/main/c/sslcontext.c
@@ -463,6 +463,23 @@ TCN_IMPLEMENT_CALL(jboolean, SSLContext, setCertificateChainFile)(TCN_STDARGS, j
     return rv;
 }
 
+TCN_IMPLEMENT_CALL(jboolean, SSLContext, setCertificateChainBio)(TCN_STDARGS, jlong ctx,
+                                                                  jlong chain,
+                                                                  jboolean skipfirst)
+{
+    tcn_ssl_ctxt_t *c = J2P(ctx, tcn_ssl_ctxt_t *);
+    BIO *b = J2P(chain, BIO *);
+
+    UNREFERENCED(o);
+    TCN_ASSERT(ctx != 0);
+    if (b == NULL)
+        return JNI_FALSE;
+    if (SSL_CTX_use_certificate_chain_bio(c->ctx, b, skipfirst) > 0)  {
+        return JNI_TRUE;
+    }
+    return JNI_FALSE;
+}
+
 TCN_IMPLEMENT_CALL(jboolean, SSLContext, setCACertificate)(TCN_STDARGS,
                                                            jlong ctx,
                                                            jstring file,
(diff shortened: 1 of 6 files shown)


normanmaurer:
@trustin please check... 

There is one important thing to note which I'm not sure yet how much of a big deal it is... When you create a new in-memory BIO it is not located in the "secure heap" as this is not supported (yet) in openssl. It might be supported in the next stable release (it's already included in the master branch as https://github.com/openssl/openssl/commit/74924dcb3802640d7e2ae2e80ca6515d0a53de7a). 

We destroy the BIO before we start using the SslContext though...

src/main/java/org/apache/tomcat/jni/SSL.java
```diff
@@ -319,6 +319,12 @@ public static native long newBIO(long pool, BIOCallback callback)
             throws Exception;
 
     /**
+     * Initialize new in-memory BIO that is located in the secure heap.
+     * @return New BIO handle
```
trustin: a new BIO handle? (no need to capitalize 'new')

src/main/java/org/apache/tomcat/jni/SSLContext.java
```diff
@@ -186,6 +186,28 @@ public static native boolean setCARevocation(long ctx, String file,
...
+     * coupled RSA+DSA certificate pair, this will work only if actually both
+     * certificates use the same certificate chain. Otherwsie the browsers will be
+     * confused in this situation.
+     * @param ctx Server or Client context to use.
+     * @param bio BIO of PEM-encoded Server CA Certificates.
+     * @param skipfirst Skip first certificate if chain file is inside
+     *                  certificate file.
+     */
+    public static native boolean setCertificateChainBio(long ctx, long bio,
+                                                         boolean skipfirst);
```
trustin: You could probably merge the two lines into one.

src/main/java/org/apache/tomcat/jni/SSLContext.java
```diff
@@ -215,6 +237,32 @@ public static native boolean setCertificate(long ctx, String cert,
...
+     * @param ctx Server or Client context to use.
+     * @param certBio Certificate BIO.
+     * @param keyBio Private Key BIO to use if not in cert.
+     * @param password Certificate password. If null and certificate
+     *                 is encrypted, password prompt will be displayed.
+     * @param idx Certificate index SSL_AIDX_RSA or SSL_AIDX_DSA.
+     */
+    public static native boolean setCertificateBio(long ctx, long certBio,
+                                                long keyBio, String password,
+                                                int idx) throws Exception;
```
trustin: Indentation?

trustin:
@normanmaurer LGTM. Please fix the cosmetic issues and cherry-pick. We could use some `#if` preprocessor tricks (or dynamic/weak binding?) later when it is released.

normanmaurer:
@trustin so you think its not a big deal to not use memory out of the secure heap ?

trustin:
@normanmaurer Is there a better option for us? I guess we are fine if what JSSE does isn't any better than what we currently do.. We could look for some BSD-licensed secure memory implementation and fork it alternatively.

normanmaurer:
The alternative would be to not merge this ;)

trustin:
@normanmaurer Ha! But didn't netty/netty#4006 depend on this?

src/main/c/sslcontext.c
```diff
@@ -870,6 +926,85 @@ TCN_IMPLEMENT_CALL(jboolean, SSLContext, setCertificate)(TCN_STDARGS, jlong ctx,
...
+        ERR_error_string(ERR_get_error(), err);
+        ERR_clear_error();
+
+        tcn_Throw(e, "Private key does not match the certificate public key (%s)",
+                  err);
+        rv = JNI_FALSE;
+        goto cleanup;
+    }
+cleanup:
+    TCN_FREE_CSTRING(password);
```
trustin: Should we memset `password` with zeroes?

trustin:
I looked into openssl source code to understand what happens when BIO is freed, and it seems to fill the BIO with 0 before freeing it. (See `BUF_MEM_free`)  It's not as safe as secmem, but it seems to me that it is as safe as loading a certificate into Java heap.

Keep in mind that I'm not a security expert, but I think we are fine unless we have to keep it for long period of time.

normanmaurer:
@trustin yep it depends on it.. :)

I agree with you and think we are ok with just doing it like here. Let me polish it and cherry-pick

src/main/c/sslcontext.c
normanmaurer: we not need this.

normanmaurer:
Cherry-picked as 1b38119a57d98572e18f47547f3aefbf88db6ff0

PR #55, opened 2015-07-31
Fix double freeing of BIO

normanmaurer:
Motivation:

Due an failure we called BIO_free(...) multiple times on a BIO on a failure.

Modifications:

Remove BIO_free(...) call.

Result:

No double free.


normanmaurer:
Cherry-picked as 4e1db18cf5fff531ee22b31b6bd5dfec162bd255

PR #56, opened 2015-08-12
BoringSSL Updates

Scottmitch:
BoringSSL Updates

Motivation:
Some NOOP methods have been added to BoringSSL to reduce the need for conditional compile macros. SSL_CIPHER_get_kx_name should also be used instead of manually contructing the names.

Modifications:
- Use SSL_CIPHER_get_kx_name
- Remove conditional compile time macros where no longer needed

Result:
More correct code with less ifdefs.

### DIFF
--- a/src/main/c/ssl.c
+++ b/src/main/c/ssl.c
@@ -665,7 +665,7 @@ TCN_IMPLEMENT_CALL(jint, SSL, initialize)(TCN_STDARGS, jstring engine)
         TCN_FREE_CSTRING(engine);
         return (jint)APR_SUCCESS;
     }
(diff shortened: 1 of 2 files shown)


Scottmitch:
@agl - FYI. I still have to wrap `CRYPTO_malloc_init` in a conditional block.  That may be OK though...thoughts?

normanmaurer:
@Scottmitch looks good to me... but will leave it to @agl to give the go for merge ;)

Scottmitch:
Cherry-picked as be811ae.

@agl - If you find anything let me know.

PR #57, opened 2015-08-27
Upgrade os-maven-plugin

normanmaurer:
Motivation:

We need to use the new os-maven-plugin to be able to release for different flavors of linux at the same time.

Modifications:

Upgrade plugin.

Result:

Be able to release for multiple linux flavors.


normanmaurer:
Cherry-picked into 1.1.33 as d9eea538f93d7b09607240c76276fc5e702d34a0

PR #59, opened 2015-09-08
Use OpenSSL methods for version rather than macros

### LINKED ISSUE #58: OpenSSL version detection broken
Calling `OpenSsl.version()` seems to result in the version that it was compiled against. I tried building openssl and setting `LD_LIBRARY_PATH`, but it kept returning the same version (1.1.33.Fork6 seems to be built against 1.0.1e). The logic for `OpenSsl.isAlpnSupported()` is based on version, so if it's broken, ALPN support is broken as well.

It seems that tc_native is using a macro to get the value: https://github.com/netty/netty-tcnative/blob/be811aeb61eb989724be54f1cb7d9a9969302e27/src/main/c/ssl.c#L273.  I tried switching to using methods:

```c
// From my local ssl.c...
TCN_IMPLEMENT_CALL(jint, SSL, version)(TCN_STDARGS)
{
    UNREFERENCED_STDARGS;
    fprintf(stderr, "NM: in new code");
    return (jint) SSLeay();
}

TCN_IMPLEMENT_CALL(jstring, SSL, versionString)(TCN_STDARGS)
{
    UNREFERENCED(o);
    fprintf(stderr, "NM: in new code");
    return AJP_TO_JSTRING(SSLeay_version(SSLEAY_VERSION));
}
```
However I still see the same problem.  I had considered that the method may have been inlined, but it appears that the definition is in a c file (https://github.com/openssl/openssl/blob/6218a1f57e7e25a6b9a798f00cf5f0e56a02ff31/crypto/cversion.c#L104), so that should work.

Any ideas?

@Scottmitch @normanmaurer

nmittler:
Motivation:

Netty's test for ALPN support in OpenSSL is dependent on
the `version()` method of tcnative.  However, this method
is currently using the macro OPENSSL_VERSION_NUMBER, which will
always return the version of OpenSSL that tcnative was built
against.  Need to change this so that version indicates the
runtime version of OpenSSL.

Modifications:

Changed the `version` and `versionString` methods to use the
OpenSSL methods `SSLeay` and `SSLeay_version`, respectively.

Result:

tcnative now indicates the OpenSSL version used at runtime.

Fixes #58

### DIFF
--- a/src/main/c/ssl.c
+++ b/src/main/c/ssl.c
@@ -270,13 +270,13 @@ static int ssl_tmp_key_init_dh(int bits, int idx)
 TCN_IMPLEMENT_CALL(jint, SSL, version)(TCN_STDARGS)
 {
     UNREFERENCED_STDARGS;
-    return OPENSSL_VERSION_NUMBER;
+    return (jint) SSLeay();
 }
 
 TCN_IMPLEMENT_CALL(jstring, SSL, versionString)(TCN_STDARGS)
 {
     UNREFERENCED(o);
-    return AJP_TO_JSTRING(OPENSSL_VERSION_TEXT);
+    return AJP_TO_JSTRING(SSLeay_version(SSLEAY_VERSION));
 }
 
 /*


nmittler:
@normanmaurer PTAL

Scottmitch:
lgtm. can you update commit message to follow template?

normanmaurer:
@nmittler what @Scottmitch said :) Once it's done I will pull in.

nmittler:
@Scottmitch @normanmaurer PTAL

normanmaurer:
@nmittler looks good... 3, 2, 1 -> cherry-picking in the process ;)

normanmaurer:
Cherry-picked as 405ebb4fc14b49f8bbd3a412a9922ea99e4ca33f

nmittler:
Woot! Thanks!

PR #62, opened 2015-09-25
Correctly validate session keys

normanmaurer:
Motivation:

If the user passed in an array that has the wrong length for session keys we failed and exist the program. This makes no sense as this will just crash the JVM. We should better validate and throw an IllegalArgumentException in this case.

Modifications:

- Move the validation to the java code
- throw IllegalArgumentException on invalid array length.

Result:

Not crashing the JVM when an array with invalid length is given.

### DIFF
--- a/src/main/c/sslcontext.c
+++ b/src/main/c/sslcontext.c
@@ -1331,7 +1331,7 @@ static int ssl_tlsext_ticket_key_cb(SSL *s, unsigned char key_name[16], unsigned
     }
 }
 
-TCN_IMPLEMENT_CALL(void, SSLContext, setSessionTicketKeys)(TCN_STDARGS, jlong ctx, jbyteArray keys)
+TCN_IMPLEMENT_CALL(void, SSLContext, setSessionTicketKeys0)(TCN_STDARGS, jlong ctx, jbyteArray keys)
 {
     tcn_ssl_ctxt_t *c = J2P(ctx, tcn_ssl_ctxt_t *);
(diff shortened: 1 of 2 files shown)


src/main/c/sslcontext.c
```diff
@@ -1340,16 +1340,6 @@ TCN_IMPLEMENT_CALL(void, SSLContext, setSessionTicketKeys)(TCN_STDARGS, jlong ct
     int i;
     int cnt;
 
-    if (((*e)->GetArrayLength(e, keys) % SSL_SESSION_TICKET_KEY_SIZE) != 0) {
```
Scottmitch: This method does a few direct memory operations (memcpy) based upon the assumption that this conditional holds. Should this method be responsible for validating this conditional instead of hoisting the responsibility on the caller? Seems like unexpected results may happen (that may be tough to track down) if the caller forgets to validate this.

normanmaurer: I think it is completely fine to only check in the caller as it is only done by our internal stuff. We do the same kind of thing in other native methods in netty.

Scottmitch: ok sgtm. C code and JNI bridge are assumed to be tightly coupled.

normanmaurer: exactly :)

normanmaurer:
Cherry-picked as ecdc3dad16ef2750803665c608191d65dc0ee67e

PR #64, opened 2015-09-28
des_read_pw_string not supported by BoringSSL

Scottmitch:
Motivation:
des_read_pw_string is not supported by BoringSSL and explicitly used by an ifdef is BoringSSL is being used.

Modifications:
- The SSL_password_prompt should not call des_read_pw_string when BoringSSL is used, and should return failure

Result:
des_read_pw_string is not used for BoringSSL.
Fixes netty#63


Scottmitch:
@normanmaurer @nmittler - PTAL

normanmaurer:
@Scottmitch go for it

nmittler:
LGTM

Scottmitch:
Cherry-picked fe5915b

Issue #65, opened 2015-09-28
Methods do not exist in BoringSSL


nmittler:
PKCS12_verify_mac (sslcontext.c)
CONF_modules_unload (ssl.c)

nmittler:
@Scottmitch For [CONF_modules_unload](https://github.com/netty/netty-tcnative/blob/1.1.33/src/main/c/ssl.c#L310), it looks like it should simply not be called for BoringSSL. It's intended to cleanup after calling [OPENSSL_load_builtin_modules](https://github.com/netty/netty-tcnative/blob/1.1.33/src/main/c/ssl.c#L691), which is not called for BoringSSL.

Scottmitch:
+1. was then starting to look into PKCS12_verify_mac (p12_mult.c in openssl)

nmittler:
Cool ... I'm pinging Adam internally to see what his thoughts are.

nmittler:
Ok, submitted the fix for CONF_modules_unload ... now just need to fix PKCS12_verify_mac

Scottmitch:
@nmittler - Is this 1 remaining method going to be supported by boringssl?

nmittler:
This method has been added to BoringSSL: https://code.google.com/p/chromium/issues/detail?id=536939

Closing this issue.

Scottmitch:
thanks @nmittler !

PR #66, opened 2015-09-28
Do not call CONF_modules_unload for BoringSSL

nmittler:
Motivation:

For CONF_modules_unload, it looks like it should simply not be called for BoringSSL. It's intended to cleanup after calling OPENSSL_load_builtin_modules, which is not called for BoringSSL.

Modifications:

Modified ssl.c to not call CONF_modules_unload if BoringSSL

Result:

Addresses part of #65


nmittler:
@Scottmitch PTAL

Scottmitch:
@nmittler - LGTM.

Make sure the issue isn't auto closed when this PR is closed.

PR #68, opened 2015-09-29
Fixing build issues

nmittler:
Motivation:

Building against openssl and boringssl is yielding several errors. They can be supressed, but it would be better to address them.

Modifications:

Various changes to ssl.c, sslutils.c, and sslcontext.c.

Result:

Builds cleanly for both openssl and boringssl.

### DIFF
--- a/src/main/c/ssl.c
+++ b/src/main/c/ssl.c
@@ -523,7 +523,6 @@ static int ssl_rand_choosenum(int l, int h)
 
 static int ssl_rand_load_file(const char *file)
 {
-    char buffer[APR_PATH_MAX];
     int n;
 
     if (file == NULL)
@@ -532,8 +531,10 @@ static int ssl_rand_load_file(const char *file)
         return -1;
 // BoringSsl doesn't support RAND_file_name, but RAND_status() returns 1 anyways
 #ifndef OPENSSL_IS_BORINGSSL
-    if (file == NULL)
+    if (file == NULL) {
+        char buffer[APR_PATH_MAX];
         file = RAND_file_name(buffer, sizeof(buffer));
+    }
 #endif
     if (file) {
 #ifdef HAVE_SSL_RAND_EGD
@@ -1210,7 +1211,7 @@ TCN_IMPLEMENT_CALL(jint, SSL, getLastErrorNumber)(TCN_STDARGS) {
 static void ssl_info_callback(const SSL *ssl, int where, int ret) {
     int *handshakeCount = NULL;
     if (0 != (where & SSL_CB_HANDSHAKE_START)) {
-        handshakeCount = (int*) SSL_get_app_data3(ssl);
+        handshakeCount = (int*) SSL_get_app_data3((SSL*) ssl);
         if (handshakeCount != NULL) {
             ++(*handshakeCount);
         }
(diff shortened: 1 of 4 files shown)


nmittler:
@Scottmitch @normanmaurer PTAL

src/main/c/sslcontext.c
```diff
@@ -1396,70 +1396,13 @@ TCN_IMPLEMENT_CALL(void, SSLContext, setSessionTicketKeys0)(TCN_STDARGS, jlong c
 
 /* OpenSSL end */
 
-/*
- * Adapted from Android:
- * https://android.googlesource.com/platform/external/openssl/+/master/patches/0003-jsse.patch
- */
-const char* SSL_CIPHER_authentication_method(const SSL_CIPHER* cipher){
```
normanmaurer: @nmittler why is this removed ? I'm a bit confused how this will still work if not boringssl is used.

nmittler: @normanmaurer non-boringssl should use SSL_authentication_method (which is not defined in boring).

normanmaurer: @nmittler are you sure "SSL_authentication_method " does the same as the method here ?

normanmaurer: @nmittler actually I'm very confused by this change and I think this is not correct.

nmittler: This method was only used by our own `SSL_authentication_method` defined just below it.  I've renamed our `SSL_authentication_method` to `authentication_method` in order to avoid the naming conflict with OpenSSL's version. I then got rid of `SSL_CIPHER_authentication_method` which also conflicts with an OpenSSL method, and we don't need anymore anyway.

src/main/c/sslcontext.c
```diff
@@ -1124,7 +1124,7 @@ TCN_IMPLEMENT_CALL(void, SSLContext, setAlpnProtos)(TCN_STDARGS, jlong ctx, jobj
         TCN_ASSERT(ctx != 0);
         UNREFERENCED(o);
 
-        if (initProtocols(e, c->pool, &c->alpn_proto_data, &c->alpn_proto_len, alpn_protos) == 0) {
+        if (initProtocols(e, c, &c->alpn_proto_data, &c->alpn_proto_len, alpn_protos) == 0) {
```
Scottmitch: this is a bit scary....maybe we should make compiler warning into errors...

Luckily it looks like this parameter wasn't used in `initProtocols`...should we just remove it from the signature?

normanmaurer: +1 let us remove it

nmittler: done

src/main/c/ssl.c
```diff
@@ -1616,7 +1617,7 @@ TCN_IMPLEMENT_CALL(jlong, SSL, getTime)(TCN_STDARGS, jlong ssl)
 
     UNREFERENCED(o);
 
-    return SSL_get_time(ssl_);
+    return SSL_get_time(ssl_->session);
```
Scottmitch: thanks @nmittler! see https://github.com/netty/netty-tcnative/issues/69.  This is used for every single `OpenSslSession` in Netty ... https://github.com/netty/netty/blob/master/handler/src/main/java/io/netty/handler/ssl/OpenSslEngine.java#L1298

normanmaurer:
@nmittler if the netty testsuite passes with this I'm fine to cherry-pick it.

normanmaurer:
@nmittler just ensure you squash first

Scottmitch:
@nmittler - LGTM!

Feels like it is time to recommend that everyone update :wink:

nmittler:
@Scottmitch @normanmaurer PTAL.  I've also cleaned up several cases of const vs non-const.

normanmaurer:
@nmittler looks good

nmittler:
@Scottmitch @normanmaurer I've updated the PR to include "-Werror" and got the build running cleanly on osx and ubuntu.

normanmaurer:
@nmittler thanks! Cherry-picked as 9b06af6863bcb8c7eaf123ef7ee6134743c26690

PR #70, opened 2015-09-30
Correctly compare session tickets

normanmaurer:
Motivation:

Due a bug during comparing session tickets, session tickers did not work.

Modifications:

- Correctly compare session tickets
- Only call RAND_bytes(...) if a session ticket is set.

Result:

Support for using session tickets works as aspected.


normanmaurer:
Cherry-picked as 3ad7d683a579f51d04c8696823252d2114bd7e90

PR #71, opened 2015-10-01
Add methods to allow support of renegotiation

normanmaurer:
Motivation:

Expose methods openssl methods to support renegotiation.

Modifications:

Add SSL.renegotiate and SSL.setState.

Result:

It's possible to implement renegotiation with tcnative now.

### DIFF
--- a/src/main/java/org/apache/tomcat/jni/SSL.java
+++ b/src/main/java/org/apache/tomcat/jni/SSL.java
@@ -234,6 +234,9 @@ public final class SSL {
     public static final int SSL_SELECTOR_FAILURE_NO_ADVERTISE = 0;
     public static final int SSL_SELECTOR_FAILURE_CHOOSE_MY_LAST_PROTOCOL = 1;
 
+    public static final int SSL_ST_CONNECT = 0x1000;
+    public static final int SSL_ST_ACCEPT =  0x2000;
+
     /* Return OpenSSL version number */
     public static native int version();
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/tomcat/jni/SSL.java
```diff
@@ -234,6 +234,9 @@
     public static final int SSL_SELECTOR_FAILURE_NO_ADVERTISE = 0;
     public static final int SSL_SELECTOR_FAILURE_CHOOSE_MY_LAST_PROTOCOL = 1;
 
+    public static final int SSL_ST_CONNECT = 0x1000;
```
Scottmitch: would it be worth exposing these defines from c code somehow? so we can use the value as defined by openssl?

normanmaurer: we don't expose the others as well from c so I thought we may just be consistent

normanmaurer:
@Scottmitch once this is merged I will open a pr against netty to use it

src/main/java/org/apache/tomcat/jni/SSL.java
Scottmitch: Ok that is fine for this PR....but should we create another issue to expose defines from C to java? Seems like it would be an improvement to use the "real" definitions.

Scottmitch:
@normanmaurer - K lgtm. just wondering if we should create a followup issue. WDYT? https://github.com/netty/netty-tcnative/pull/71#discussion_r40927121

src/main/java/org/apache/tomcat/jni/SSL.java
normanmaurer: yeah maybe... I just don't want to change too many things as we may get into merge trouble once we change too many things and netty to upgrade to a new tcnative version for our fork

normanmaurer:
@Scottmitch I will create one... not sure yet if we should do or not though. But let me open it so we are aware of it at least.

src/main/java/org/apache/tomcat/jni/SSL.java
Scottmitch: @normanmaurer - that is a good point. I'm not sure if this will be a deal breaker though...lets discuss in new issue :)

normanmaurer:
@Scottmitch https://github.com/netty/netty-tcnative/issues/72

normanmaurer:
Cherry-picked into 1.1.33 as d4ebb73596b928864f7b2ffb9f59082e14e1c77a

PR #73, opened 2015-10-01
Fixing segfault with BoringSSL

nmittler:
Motivation:

BoringSSL's SSL_SESSION_get_time method does not guard against a null session, however OpenSSL's returns 0 if the session is null.  This seems to cause a segfault when starting up the connection:

Stack: [0x00007f877b516000,0x00007f877b617080],  sp=0x00007f877b614ea0,  free space=1019k
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
C  [libnetty-tcnative.so+0xd449c]  SSL_SESSION_get_time+0xc
C  [libnetty-tcnative.so+0x79448]  Java_org_apache_tomcat_jni_SSL_getTime+0x78
j  org.apache.tomcat.jni.SSL.getTime(J)J+0
j  io.netty.handler.ssl.OpenSslEngine$OpenSslSession.<init>(Lio/netty/handler/ssl/OpenSslEngine;JLio/netty/handler/ssl/OpenSslSessionContext;)V+11
j  io.netty.handler.ssl.OpenSslEngine.<init>(JLio/netty/buffer/ByteBufAllocator;ZLio/netty/handler/ssl/OpenSslSessionContext;Lio/netty/handler/ssl/OpenSslApplicationProtocolNegotiator;Lio/netty/handler/ssl/OpenSslEngineMap;ZLjava/lang/String;I[Ljava/security/cert/Certificate;Lio/netty/handler/ssl/ClientAuth;)V+123
j  io.netty.handler.ssl.OpenSslContext.newEngine(Lio/netty/buffer/ByteBufAllocator;Ljava/lang/String;I)Ljavax/net/ssl/SSLEngine;+39
j  io.grpc.netty.ProtocolNegotiators$2$1.handlerAdded(Lio/netty/channel/ChannelHandlerContext;)V+24

Modifications:

Modified the getTime method in ssl.c to guard against a null session.

Result:

tcnative works with BoringSSL.

### DIFF
--- a/src/main/c/ssl.c
+++ b/src/main/c/ssl.c
@@ -1615,6 +1615,12 @@ TCN_IMPLEMENT_CALL(jlong, SSL, getTime)(TCN_STDARGS, jlong ssl)
         return 0;
     }
 
+    if (ssl_->session == NULL) {
+        // BoringSSL does not protect against a NULL session. OpenSSL
+        // returns 0 if the session is NULL, so do that here.
+        return 0;
+    }
+
     UNREFERENCED(o);
 
     return SSL_get_time(ssl_->session);


nmittler:
@normanmaurer PTAL

normanmaurer:
@nmittler please cherry-pick

src/main/c/ssl.c
```diff
@@ -1615,6 +1615,12 @@ TCN_IMPLEMENT_CALL(jlong, SSL, getTime)(TCN_STDARGS, jlong ssl)
         return 0;
     }
 
+    if (ssl_->session == NULL) {
```
Scottmitch: Maybe we should file a bug against BoringSSL?

https://www.openssl.org/docs/manmaster/ssl/SSL_SESSION_get_time.html
> If any of the function is passed the NULL pointer for the session s, 0 is returned.

nmittler:
@normanmaurer why do you think the session is null at the point that this is called?  Is that expected?

src/main/c/ssl.c
Scottmitch: If we need a quick work around then consider wrapping this in a `#ifdef OPENSSL_IS_BORINGSSL` with a comment to the bug report?

normanmaurer: +1. @nmittler please do and add a reference to the comment.

normanmaurer:
@nmittler nope... working on a fix as we speak ;)

nmittler:
@normanmaurer awesome!

nmittler:
Cherry-picked as d8897e9

nmittler:
@Scottmitch @normanmaurer FYI, I've filed a bug against BoringSSL: https://code.google.com/p/chromium/issues/detail?id=538292&thanks=538292&ts=1443727165

src/main/c/ssl.c
nmittler: @Scottmitch I filed a ticket ... see the main thread for this PR.

normanmaurer:
@nmittler no perms to access it :(

nmittler:
ah bummer ... not sure why the issues aren't public :/

PR #74, opened 2015-10-02
Add methods to get and set the timout of a SSL session.

normanmaurer:
Motivation:

To be able to fully implement SSLSession we need to be able to get and set the timeout of a SSL Session.

Modifications:

Add SSL.getTimeout(...) and SSL.setTimeout(...)

Result:

Be able to manipulate and get informations about the SSL Session timeout.

### DIFF
--- a/src/main/c/ssl.c
+++ b/src/main/c/ssl.c
@@ -1757,6 +1757,48 @@ TCN_IMPLEMENT_CALL(jlong, SSL, getTime)(TCN_STDARGS, jlong ssl)
     return SSL_get_time(ssl_->session);
 }
 
+
+TCN_IMPLEMENT_CALL(jlong, SSL, getTimeout)(TCN_STDARGS, jlong ssl)
+{
+    SSL *ssl_ = J2P(ssl, SSL *);
+
+    if (ssl_ == NULL) {
+        tcn_ThrowException(e, "ssl is null");
+        return 0;
+    }
+
+    if (ssl_->session == NULL) {
+        // BoringSSL does not protect against a NULL session. OpenSSL
+        // returns 0 if the session is NULL, so do that here.
+        return 0;
+    }
+
+    UNREFERENCED(o);
+
+    return SSL_get_timeout(ssl_->session);
+}
+
+
+TCN_IMPLEMENT_CALL(jlong, SSL, setTimeout)(TCN_STDARGS, jlong ssl, jlong seconds)
+{
+    SSL *ssl_ = J2P(ssl, SSL *);
+
+    if (ssl_ == NULL) {
+        tcn_ThrowException(e, "ssl is null");
+        return 0;
+    }
+    if (ssl_->session == NULL) {
+        // BoringSSL does not protect against a NULL session. OpenSSL
+        // returns 0 if the session is NULL, so do that here.
+        return 0;
+    }
+
+    UNREFERENCED(o);
+
+    return SSL_set_timeout(ssl_->session, seconds);
+}
+
+
 TCN_IMPLEMENT_CALL(void, SSL, setVerify)(TCN_STDARGS, jlong ssl,
                                                 jint level, jint depth)
 {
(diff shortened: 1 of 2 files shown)


normanmaurer:
@Scottmitch @nmittler please check

src/main/c/ssl.c
```diff
@@ -1626,6 +1626,37 @@ TCN_IMPLEMENT_CALL(jlong, SSL, getTime)(TCN_STDARGS, jlong ssl)
...
+    SSL *ssl_ = J2P(ssl, SSL *);
+
+    if (ssl_ == NULL) {
+        tcn_ThrowException(e, "ssl is null");
+        return 0;
+    }
+
+    UNREFERENCED(o);
+
+    return SSL_set_timeout(ssl_->session, seconds);
```
normanmaurer: @nmittler does boring handle a null session here ?

src/main/c/ssl.c
```diff
@@ -1626,6 +1626,37 @@ TCN_IMPLEMENT_CALL(jlong, SSL, getTime)(TCN_STDARGS, jlong ssl)
...
+    SSL *ssl_ = J2P(ssl, SSL *);
+
+    if (ssl_ == NULL) {
+        tcn_ThrowException(e, "ssl is null");
+        return 0;
+    }
+
+    UNREFERENCED(o);
+
+    return SSL_get_timeout(ssl_->session);
```
normanmaurer: @nmittler does boring handle a null session here ?

nmittler: nope

src/main/c/ssl.c
nmittler: nope

src/main/java/org/apache/tomcat/jni/SSL.java
```diff
@@ -612,6 +612,21 @@ public static native long newBIO(long pool, BIOCallback callback)
     public static native long getTime(long ssl);
 
     /**
+     * SSL_get_timeout
+     * @param ssl the SSL instance (SSL *)
+     * @return returns the timeout for the session ssl The time is given in seconds since the Epoch
+     */
+    public static native long getTimeout(long ssl);
```
nmittler: Are these just for completeness or are you planning on using them in Netty?

normanmaurer: I'm using them in netty... PR is coming ;)

src/main/c/ssl.c
normanmaurer: ok let me add guards then

normanmaurer:
@nmittler please check again

normanmaurer:
Used by https://github.com/netty/netty/pull/4307

src/main/c/ssl.c
```diff
@@ -1626,6 +1626,49 @@ TCN_IMPLEMENT_CALL(jlong, SSL, getTime)(TCN_STDARGS, jlong ssl)
...
+TCN_IMPLEMENT_CALL(jlong, SSL, getTimeout)(TCN_STDARGS, jlong ssl)
+{
+    SSL *ssl_ = J2P(ssl, SSL *);
+
+    if (ssl_ == NULL) {
+        tcn_ThrowException(e, "ssl is null");
+        return 0;
+    }
+
+    UNREFERENCED(o);
```
nmittler: What's the difference between this and `UNREFERENCED_STDARGS`?

src/main/c/ssl.c
```diff
@@ -1626,6 +1626,49 @@ TCN_IMPLEMENT_CALL(jlong, SSL, getTime)(TCN_STDARGS, jlong ssl)
...
+
+    if (ssl_->session == NULL) {
```
nmittler: Move this up to just below the other check?

nmittler:
@normanmaurer LGTM ... just a couple questions

src/main/c/ssl.c
normanmaurer: the difference is that UNREFERENCED_STDARGS will unreferenced JNIEnv and jobject (which I think would be fine here as well). That said UNREFERENCED(o) is used in getTime(...) so maybe just leave it. WDYT ?

nmittler: Yeah, it's fine.  Thanks.

nmittler:
@normanmaurer fire away :)

normanmaurer:
@nmittler you may want to add another bugreport for boringssl or extend the scope of the existing

nmittler:
@normanmaurer I'm not sure how much they care to be honest.  They look at this as: "maybe a segfault is the right thing".  But for JNI it's probably not.  I think we (in tcnative) should just guard against NULLs even if the OpenSSL docs state that they should handle it.

normanmaurer:
@nmittler well if you want to be API compatible and the api docs state that null is fine I would say it is clear what should be done ;)

nmittler:
@normanmaurer BoringSSL does not make the claim that they are API compatible.  Boring has their own docs which do not make the same claim about NULL parameters.

normanmaurer:
@nmittler oh I thought they are... I just hope we will not get into too much trouble at some point to support boringssl if they change behaviour compared to openssl.

nmittler:
@normanmaurer I think we're ok in general. I see handling NULL pointers as a one-off ... many C apis would crash for this sort of thing, which is reasonable behavior.  We'll be using boring internally so we have motivation for keeping tcnative working with it :)

normanmaurer:
@nmittler fair enough...

normanmaurer:
Cherry-picked as 094163498157927fe198e0b11968dcfdf8c9e46b

PR #75, opened 2015-10-02
Fix compile error if HAS_OPENSSL is not set

normanmaurer:
Motivation:

If HAS_OPENSSL is not set a compile is raised.

Modifications:

Fix compile error.

Result:

No compile error if HAS_OPENSSL is not set.


nmittler:
LGTM

normanmaurer:
Cherry-picked as 1f3d4180749d82cc5a651244c68c8c12e2bfc886

PR #76, opened 2015-10-02
Handling null arguments in ssl.c

nmittler:
Motivation:

There are several methods in ssl.c that do not verify the parameters before using them, which has been the cause of segfaults with openssl variants (i.e. BoringSSL).

Modifications:

Tightening up the API methods in ssl.c to properly handle null arguments.

Result:

ssl.c should do better at guarding against segfaults

### DIFF
--- a/src/main/c/ssl.c
+++ b/src/main/c/ssl.c
@@ -1543,7 +1543,7 @@ TCN_IMPLEMENT_CALL(jstring, SSL, getVersion)(TCN_STDARGS,
 
     if (ssl_ == NULL) {
         tcn_ThrowException(e, "ssl is null");
-        return AJP_TO_JSTRING("");
+        return NULL;
     }
 
     UNREFERENCED_STDARGS;
--- a/src/main/c/ssl.c
+++ b/src/main/c/ssl.c
@@ -1753,7 +1753,6 @@ TCN_IMPLEMENT_CALL(jlong, SSL, getTime)(TCN_STDARGS, jlong ssl)
     if (ssl_->session == NULL) {
         // BoringSSL does not protect against a NULL session. OpenSSL
         // returns 0 if the session is NULL, so do that here.
-        // TODO(nmittler): should we throw here?
         return 0;
     }
 
--- a/src/main/c/ssl.c
+++ b/src/main/c/ssl.c
@@ -1114,8 +1114,13 @@ init_failed:
 TCN_IMPLEMENT_CALL(jint, SSL, closeBIO)(TCN_STDARGS, jlong bio)
 {
     BIO *b = J2P(bio, BIO *);
+
     UNREFERENCED_STDARGS;
-    SSL_BIO_close(b);
+
+    if (b != NULL) {
+        SSL_BIO_close(b);
+    }
+
     return APR_SUCCESS;
 }
 
@@ -1225,9 +1230,17 @@ TCN_IMPLEMENT_CALL(jlong /* SSL * */, SSL, newSSL)(TCN_STDARGS,
     int *handshakeCount = malloc(sizeof(int));
     SSL *ssl;
 
+    if (c == NULL) {
+        tcn_ThrowException(e, "ssl ctx is null");
+        return 0;
+    }
+    if (c->ctx == NULL) {
+        tcn_ThrowException(e, "ctx is null");
+        return 0;
+    }
+
     UNREFERENCED_STDARGS;
 
-    TCN_ASSERT(ctx != 0);
     ssl = SSL_new(c->ctx);
     if (ssl == NULL) {
         tcn_ThrowException(e, "cannot create new ssl");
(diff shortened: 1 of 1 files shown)


nmittler:
@normanmaurer @Scottmitch PTAL

src/main/c/ssl.c
```diff
@@ -1618,6 +1753,7 @@ TCN_IMPLEMENT_CALL(jlong, SSL, getTime)(TCN_STDARGS, jlong ssl)
     if (ssl_->session == NULL) {
         // BoringSSL does not protect against a NULL session. OpenSSL
         // returns 0 if the session is NULL, so do that here.
+        // TODO(nmittler): should we throw here?
```
normanmaurer: nope we should not throw

normanmaurer:
@nmittler one comment.

src/main/c/ssl.c
nmittler: @normanmaurer Just to be clear, is your concern of using tcnative with BoringSSL and an old version of netty? With the new version of Netty, throwing would be the right thing to do, since it would expose a programming error.

normanmaurer: @nmittler nope... we still may call this method if the session is not created yet even within netty. So I think we should not throw here otherwise it will be just harder to guard

nmittler: sgtm ... done.

nmittler:
@normanmaurer shall I cherry-pick?

normanmaurer:
@nmittler let @Scottmitch have a look as well.

src/main/c/ssl.c
```diff
@@ -1292,9 +1337,21 @@ TCN_IMPLEMENT_CALL(jint /* status */, SSL, writeToBIO)(TCN_STDARGS,
...
                                                        jint wlen /* sizeof(wbuf) */) {
+    BIO *b = J2P(bio, BIO *);
+    void *w = J2P(wbuf, void *);
+
+    if (b == NULL) {
+        tcn_ThrowException(e, "bio is null");
+        return 0;
+    }
+    if (w == NULL) {
+        tcn_ThrowException(e, "buffer is null");
```
normanmaurer: wbuf is null ?

src/main/c/ssl.c
```diff
@@ -1303,9 +1360,21 @@ TCN_IMPLEMENT_CALL(jint /* status */, SSL, readFromBIO)(TCN_STDARGS,
...
                                                         jint rlen /* sizeof(rbuf) - 1 */) {
+    BIO *b = J2P(bio, BIO *);
+    void *r = J2P(rbuf, void *);
+
+    if (b == NULL) {
+        tcn_ThrowException(e, "bio is null");
+        return 0;
+    }
+    if (r == NULL) {
+        tcn_ThrowException(e, "buffer is null");
```
normanmaurer: rbuf is null ?

src/main/c/ssl.c
```diff
@@ -1323,32 +1404,62 @@ TCN_IMPLEMENT_CALL(jint /* status */, SSL, readFromSSL)(TCN_STDARGS,
...
                                                         jint rlen /* sizeof(rbuf) - 1 */) {
+    SSL *ssl_ = J2P(ssl, SSL *);
+    void *r = J2P(rbuf, void *);
+
+    if (ssl_ == NULL) {
+        tcn_ThrowException(e, "ssl is null");
+        return 0;
+    }
+    if (r == NULL) {
+        tcn_ThrowException(e, "buffer is null");
```
normanmaurer: rbuf is null

src/main/c/ssl.c
nmittler: done.

src/main/c/ssl.c
nmittler: done.

src/main/c/ssl.c
nmittler: done.

nmittler:
@Scottmitch do you want to have a look?

Scottmitch:
@nmittler - Will look this afternoon.

src/main/c/ssl.c
```diff
@@ -1225,9 +1230,17 @@ TCN_IMPLEMENT_CALL(jlong /* SSL * */, SSL, newSSL)(TCN_STDARGS,
     int *handshakeCount = malloc(sizeof(int));
```
Scottmitch: potential memory leak. do this after you check pre conditions.

src/main/c/ssl.c
```diff
@@ -1366,59 +1477,82 @@ TCN_IMPLEMENT_CALL(jlong, SSL, makeNetworkBIO)(TCN_STDARGS,
...
- fail:
-    return 0;
 }
 
 // Free a BIO * (typically, the network BIO)
 TCN_IMPLEMENT_CALL(void, SSL, freeBIO)(TCN_STDARGS,
                                        jlong bio /* BIO * */) {
-    BIO *bio_;
+    BIO *bio_ = J2P(bio, BIO *);
+    if (bio_ == NULL) {
```
Scottmitch: nit: New line consistency? other functions have a new line in between declarations/initialization and preconditions.  I would prefer no-newline...but just as long as we are consistent.

src/main/c/ssl.c
```diff
@@ -1752,14 +1887,13 @@ TCN_IMPLEMENT_CALL(jboolean, SSL, setCipherSuites)(TCN_STDARGS, jlong ssl,
     SSL *ssl_ = J2P(ssl, SSL *);
     TCN_ALLOC_CSTRING(ciphers);
```
Scottmitch: potential memory leak? move until after pre condition check.

Scottmitch:
@nmittler - code looks good.

Does Netty ever call these functions where it is OK to not throw exceptions? I guess you will say those are netty bugs and we should fix them there ... but then do we need to check in both places? Seems like in some instances we will have 3 layers all checking the same thing (netty, tcnative, openssl).

normanmaurer:
The only place where this can happen (in terms of netty) is where we use ssl_->session. The session may be NULL

src/main/c/ssl.c
nmittler: This seems unrelated to this PR ... or am I missing something?

nmittler: Oh nevermind ... haven't had my coffee yet :)

nmittler: done.

src/main/c/ssl.c
nmittler: done.

src/main/c/ssl.c
nmittler: done.

nmittler:
@Scottmitch PTAL

src/main/c/ssl.c
```diff
@@ -1260,31 +1272,63 @@ TCN_IMPLEMENT_CALL(void, SSL, setBIO)(TCN_STDARGS,
...
+    if (ssl_ == NULL) {
+        tcn_ThrowException(e, "ssl is null");
+        return;
+    }
+
     UNREFERENCED_STDARGS;
-    SSL_set_bio(J2P(ssl, SSL *), J2P(rbio, BIO *), J2P(wbio, BIO *));
+
+    SSL_set_bio(ssl_, r, w);
     return;
```
Scottmitch: cleanup: remove?

src/main/c/ssl.c
```diff
@@ -1366,59 +1476,83 @@ TCN_IMPLEMENT_CALL(jlong, SSL, makeNetworkBIO)(TCN_STDARGS,
...
 }
 
 // Free a BIO * (typically, the network BIO)
 TCN_IMPLEMENT_CALL(void, SSL, freeBIO)(TCN_STDARGS,
                                        jlong bio /* BIO * */) {
-    BIO *bio_;
+    BIO *bio_ = J2P(bio, BIO *);
+
+    if (bio_ == NULL) {
+        tcn_ThrowException(e, "bio is null");
```
Scottmitch: For `closeBIO` we didn't throw an exception but instead we just did the close if the pointer was not null. Should we do something similar here?

src/main/c/ssl.c
```diff
@@ -1366,59 +1476,83 @@ TCN_IMPLEMENT_CALL(jlong, SSL, makeNetworkBIO)(TCN_STDARGS,
...
 
 // Read which cipher was negotiated for the given SSL *.
 TCN_IMPLEMENT_CALL(jstring, SSL, getCipherForSSL)(TCN_STDARGS,
                                                   jlong ssl /* SSL * */)
 {
+    SSL *ssl_ = J2P(ssl, SSL *);
+
+    if (ssl_ == NULL) {
+        tcn_ThrowException(e, "ssl is null");
+        return AJP_TO_JSTRING("");
```
Scottmitch: not sure if signatures lines up ... but can we just return null here? We shouldn't get to this line anyways right?

Scottmitch:
@nmittler - just a few questions. LGTM after these are worked out.

src/main/c/ssl.c
nmittler: done.

src/main/c/ssl.c
nmittler: done.

src/main/c/ssl.c
nmittler: @normanmaurer is `NULL` ok here?

normanmaurer: @nmittler NULL is ok as we throw the exception anyway and it is just to return early.

nmittler:
@Scottmitch @normanmaurer anything else before I cherry-pick?

normanmaurer:
Fine by me.. @Scottmitch?

nmittler:
Cherry-picked as 83b3036

Scottmitch:
thanks @nmittler !

normanmaurer:
@nmittler :heart_eyes:

nmittler:
Ha!  I feel the love :)

PR #77, opened 2015-10-10
Allow to compile tcnative against libressl 2.3.0

normanmaurer:
Motivation:

libressl 2.3.0 removed a few methods, so we need to guard against this via ifdefs.

Modifications:

Added pre-processor rules to be able to compile against libressl 2.3.0

Result:

tcnative can be compiled again libressl >= 2.3.0

### DIFF
--- a/src/main/c/ssl.c
+++ b/src/main/c/ssl.c
@@ -561,10 +561,12 @@ static int ssl_rand_save_file(const char *file)
 #ifndef OPENSSL_IS_BORINGSSL
     char buffer[APR_PATH_MAX];
     int n;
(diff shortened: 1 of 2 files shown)


normanmaurer:
@Scottmitch @nmittler FYI

nmittler:
@normanmaurer LGTM ... should you add docs to the wiki for building with libressl?

normanmaurer:
@nmittler yep will do :)

normanmaurer:
Cherry-picked as cad514a1961b5061cd51f3994ffa131fe87d5349

PR #78, opened 2015-10-13
Allow to get the release against which tcnative was compiled.

normanmaurer:
Motivation:

It's useful to detect against which openssl compatible project / release tcnative was compiled as some features may only work with one or the other.

Modifications:

Add method to SSL to detect it.

Result:

Possible to get informations about against which release it was compiled.

### DIFF
--- a/src/main/c/ssl.c
+++ b/src/main/c/ssl.c
@@ -2036,6 +2036,17 @@ TCN_IMPLEMENT_CALL(void, SSL, setState)(TCN_STDARGS,
     SSL_set_state(ssl_, state);
 }
 
+TCN_IMPLEMENT_CALL(jint, SSL, release0)(TCN_STDARGS)
+{
+    UNREFERENCED_STDARGS;
+    #ifdef LIBRESSL_VERSION_NUMBER
+        return 1;
+    #elif defined(OPENSSL_IS_BORINGSSL)
+        return 2;
+    #else
+        return 0;
+    #endif
+}
 
 /*** End Apple API Additions ***/
 
@@ -2451,5 +2462,12 @@ TCN_IMPLEMENT_CALL(void, SSL, setState)(TCN_STDARGS, jlong ssl, jint state) {
   UNREFERENCED(state);
   tcn_ThrowException(e, "Not implemented");
 }
+
+TCN_IMPLEMENT_CALL(jint, SSL, release0)(TCN_STDARGS)
+{
+    UNREFERENCED_STDARGS;
+    tcn_ThrowException(e, "Not implemented");
+}
+
 /*** End Apple API Additions ***/
 #endif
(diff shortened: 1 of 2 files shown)


normanmaurer:
@Scottmitch @nmittler please check

src/main/java/org/apache/tomcat/jni/SSL.java
```diff
@@ -237,6 +237,15 @@
     public static final int SSL_ST_CONNECT = 0x1000;
     public static final int SSL_ST_ACCEPT =  0x2000;
 
+    /**
+     * The actual release that is used and implement the OPENSSL API / ABI
+     */
+    public enum Release {
```
normanmaurer: @nmittler @Scottmitch not sure if Release is a good name... suggestions welcome :)

src/main/c/ssl.c
```diff
@@ -2036,6 +2036,17 @@ TCN_IMPLEMENT_CALL(void, SSL, setState)(TCN_STDARGS,
     SSL_set_state(ssl_, state);
 }
 
+TCN_IMPLEMENT_CALL(jint, SSL, release0)(TCN_STDARGS)
```
Scottmitch: Will something like the following be sufficient? With the current system we may be masking useful information related to the specific version within each implementation of the OpenSSL interface.

```c
TCN_IMPLEMENT_CALL(jint, SSL, versionAsCompiled)(TCN_STDARGS) {
  return SSLEAY_VERSION;
}
```

normanmaurer: is it guaranteed that SSLEAY_VERSION will return different things for libressl / openssl / boringssl ? Also where would the user find the values for each of the different releases ?

Scottmitch: No sry...what I gave will just print something that identifies 1.0.2.  What I was thinking of would be represented by our current `versionString` method (compile time version string).

I was originally assuming this was just for informational purposes ... but we actually plan on varying behavior (in java code?) depending on the compile time library?

normanmaurer: yeah... for example server side renegotiation will not work with boringssl. So it may be useful for a user to be able to see what can be used and what not

nmittler: Maybe we should have:

``` java
enum OpenSslVariant {
    BORINGSSL,
    LIBRESSL,
    OPENSSL
}

interface OpenSslInfo {
    OpenSslVariant variant();
    int version();
    String versionString();
}

class OpenSsl {
 ...
    OpenSslInfo getCompileInfo();

    // We can pobably do some hacks to determine the variant at runtime
    // (e.g. if version().contains("BoringSSL"))
    OpenSslInfo getRuntimeInfo();
}
```

WDYT?

Scottmitch: @nmittler - Yes this makes sense.

normanmaurer: Yep makes sense... Let me update the PR :)

PR #79, opened 2015-10-14
Fix build on windows

normanmaurer:
Motivation:

While doing a bit of cleanup we broke compiling on windows.

Modifications:

Ensure we can also compile on windows.

Result:

Building on windows works again as well.

### DIFF
--- a/src/main/c/ssl.c
+++ b/src/main/c/ssl.c
@@ -1228,6 +1228,8 @@ static void ssl_info_callback(const SSL *ssl, int where, int ret) {
 TCN_IMPLEMENT_CALL(jlong /* SSL * */, SSL, newSSL)(TCN_STDARGS,
                                                    jlong ctx /* tcn_ssl_ctxt_t * */,
                                                    jboolean server) {
+    SSL *ssl = NULL;
+    int *handshakeCount = NULL;
     tcn_ssl_ctxt_t *c = J2P(ctx, tcn_ssl_ctxt_t *);
 
     if (c == NULL) {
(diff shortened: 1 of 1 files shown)


normanmaurer:
@nmittler please review as you are guilty =P

src/main/c/ssl.c
```diff
@@ -1228,6 +1228,8 @@ static void ssl_info_callback(const SSL *ssl, int where, int ret) {
 TCN_IMPLEMENT_CALL(jlong /* SSL * */, SSL, newSSL)(TCN_STDARGS,
                                                    jlong ctx /* tcn_ssl_ctxt_t * */,
                                                    jboolean server) {
+    SSL *ssl = NULL;
```
nmittler: Hmm ... what was wrong with the way it was before?

normanmaurer: Stuff needs to be declared on the beginning of the method

Scottmitch: C89/ANSI requires declarations to be up front ... didn't' know we were compatible back to c89.

normanmaurer: We may be able to fix this in the windows build file... Anyway let us look into it after this was merged

nmittler: It's fine ... I don't think we should change the build file.

nmittler:
@normanmaurer  LGTM

normanmaurer:
Cherry-picked into 1.1.33 as 2aa47be27783ec31086ca9881402f845543de4e6

PR #82, opened 2016-01-11
Adding profile for BoringSSL

nmittler:
Motivation:

A major pain point for deployments  is that we require OpenSSL be installed on the system, which is not automatically the case on many systems (e.g. windows, osx).

We should provide a way to easily build and statically link netty-tcnative against BoringSSL. This can eventually lead to official distributions as well.

Modifications:

Added a profile to the pom.xml to build against BoringSSL. When this profile is activated, the artifactId becomes netty-tcnative-boringssl.

Result:

Users can now choose to build a statically linked version of netty-tcnative.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -21,7 +21,7 @@
     <artifactId>netty-parent</artifactId>
     <version>4.0.18.Final</version>
   </parent>
-  <artifactId>netty-tcnative</artifactId>
+  <artifactId>${artifactIdToUse}</artifactId>
   <version>1.1.33.Fork11-SNAPSHOT</version>
   <packaging>jar</packaging>
 
@@ -39,6 +39,7 @@
   </scm>
 
   <properties>
+    <artifactIdToUse>netty-tcnative</artifactIdToUse>
     <checkstyle.skip>true</checkstyle.skip>
     <animal.sniffer.skip>true</animal.sniffer.skip>
     <forceAutogen>true</forceAutogen>
@@ -47,6 +48,7 @@
     <nativeJarFile>${project.build.directory}/${project.build.finalName}-${os.detected.classifier}.jar</nativeJarFile>
     <nativeLibOnlyDir>${project.build.directory}/native-lib-only</nativeLibOnlyDir>
     <nativeJarWorkdir>${project.build.directory}/native-jar-work</nativeJarWorkdir>
+    <maven-hawtjni-plugin-version>1.11</maven-hawtjni-plugin-version>
   </properties>
 
   <build>
@@ -103,7 +105,7 @@
       <plugin>
         <groupId>org.fusesource.hawtjni</groupId>
         <artifactId>maven-hawtjni-plugin</artifactId>
-        <version>1.10</version>
+        <version>${maven-hawtjni-plugin-version}</version>
         <executions>
           <execution>
             <id>build-native-lib</id>
@@ -123,10 +125,29 @@
         </executions>
       </plugin>
 
-      <!-- Build the additional JAR that contains the native library. -->
       <plugin>
         <artifactId>maven-antrun-plugin</artifactId>
         <executions>
+          <!--
+            Fedora-based systems use a different soname for OpenSSL than other linux distributions.
+            Use a custom classifier ending in "-fedora" when building on fedora-based systems.
+          -->
+          <execution>
+            <phase>initialize</phase>
+            <configuration>
+              <exportAntProperties>true</exportAntProperties>
+              <target>
+                <condition property="classifier_to_use" value="${os.detected.classifier}-fedora" else="${os.detected.classifier}">
+                  <isset property="os.detected.release.like.fedora" />
+                </condition>
+              </target>
+            </configuration>
+            <goals>
+              <goal>run</goal>
+            </goals>
+          </execution>
+
+          <!-- Build the additional JAR that contains the native library. -->
           <execution>
             <id>native-jar</id>
             <phase>package</phase>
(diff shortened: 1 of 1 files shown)


nmittler:
FYI @normanmaurer @Scottmitch @ejona86 @louiscryan

pom.xml
```diff
@@ -176,35 +196,101 @@
...
+    <!-- Build distribution statically linked to BoringSSL -->
+    <profile>
+      <id>boringssl</id>
+
+      <properties>
+        <artifactIdToUse>netty-tcnative-boringssl</artifactIdToUse>
+        <boringSslHome>${project.build.directory}/boringssl</boringSslHome>
+        <checkoutDirectory>${boringSslHome}</checkoutDirectory>
+        <boringSslBuildDir>${boringSslHome}/build</boringSslBuildDir>
+        <boringSslVersion>7b8b9c17db93ea5287575b437c77fb36eeb81b31</boringSslVersion>
```
normanmaurer: I guess this is the sha for the latest boring version that was used in the latest chromium release ? Is there any day to automate this to update on a new release ?

pom.xml
```diff
@@ -176,35 +196,101 @@
...
+            <executions>
+              <execution>
+                <id>build-native-lib</id>
+                <configuration>
+                  <nativeSourceDirectory>${project.basedir}/src/main/c</nativeSourceDirectory>
+                  <libDirectory>${nativeLibOnlyDir}</libDirectory>
+                  <forceAutogen>${forceAutogen}</forceAutogen>
+                  <forceConfigure>${forceConfigure}</forceConfigure>
+                  <windowsBuildTool>msbuild</windowsBuildTool>
+                  <verbose>true</verbose>
```
normanmaurer: remove ?

pom.xml
```diff
@@ -176,35 +196,101 @@
...
+                </configuration>
+              </execution>
+            </executions>
+          </plugin>
+
+          <!-- Configure the distribution statically linked against BoringSSL -->
+          <plugin>
+            <groupId>org.fusesource.hawtjni</groupId>
+            <artifactId>maven-hawtjni-plugin</artifactId>
+            <version>1.11</version>
```
normanmaurer: use a property for the version and use the property in both places where you specified the plugin ?

pom.xml
nmittler: done.

pom.xml
nmittler: done.

pom.xml
nmittler: @agl any thoughts here?

agl: What would work for you? I tag the current tip-of-tree for Cocoapods when GRPC ask me to. We could have a branch called chromium-stable that we try to update, but there could be API tweaks that would break you when that's updated. Other projects periodically sync BoringSSL, test the integration and then roll in the new version.

normanmaurer: a branch sounds good. If it breaks we will adjust the code as needed.

nmittler:
@agl  I tried building boringssl on osx (both with make and ninja) and it fails seemingly due to redefinition of macros.  Have you tried recently? Any thoughts?

nmittler:
@Scottmitch @normanmaurer I haven't had much luck figuring out how to statically link to apr.  Looking at `tcnative.m4`, it defines:

```
  APR_LIBS="`$apr_config --link-libtool --libs`"
```
The options seem to be defined here: https://svn.apache.org/repos/asf/apr/apr/trunk/apr-config.in

I had tried playing with this, but no matter what `APR_LIBS` ends up being, we still seem to dynamically link to APR.

I had also tried passing "-Wl,-Bstatic" in the  `LDFLAGS` passed in by the plugin, but no dice.

Any ideas?  Perhaps we punt on this for now and iterate?

normanmaurer:
@nmittler no ideas atm.. but I think iterate is good enough. I mean this is not something we do for a release yet anyway.

pom.xml
nmittler: @agl `chromium-stable` SGTM.  If you're ok with that, I'll hold off on this PR for now, until you've created the branch.

agl: A chromium-stable branch now exists in BoringSSL.

nmittler: Great, thanks @agl!

nmittler:
@Scottmitch @normanmaurer now using chromium-stable branch.  Anything else before we cherry-pick?

src/main/native-package/m4/tcnative.m4
```diff
@@ -281,7 +281,7 @@ dnl
 dnl Configure for the detected openssl toolkit installation, giving
 dnl preference to "--with-ssl=<path>" if it was specified.
 dnl
-AC_DEFUN(TCN_CHECK_SSL_TOOLKIT,[
+AC_DEFUN([TCN_CHECK_SSL_TOOLKIT],[
 OPENSSL_WARNING=
 AC_MSG_CHECKING(for OpenSSL library)
 AC_ARG_WITH(ssl,
```
normanmaurer: @nmittler are the changes in this file needed ? I don't thin so.

nmittler: not needed, but they do fix pre-existing warnings (underquoted names) that you don't see unless you turn on verbose logging.

normanmaurer: @nmittler let us fix this in another PR as you will also need to fix the 'bootstrap' branch

normanmaurer: after that squash and commit

nmittler:
Reverted change to tcnative.m4.

cherry-picked as 96e545e

PR #83, opened 2016-01-14
Not blindly cast to jlong when return a pointer to make it work on 32bit

normanmaurer:
Motivation:

We blindly cast to jlong in one case when return a pointer and so it will not build on 32bit. Related to https://github.com/netty/netty/issues/4694.

Modifications:

Just return 0 on error and not try to cast to jlong.

Result:

Also builds on 32bit.


normanmaurer:
@nmittler @Scottmitch @trustin  PTAL

normanmaurer:
Cherry-picked into 1.1.33 as 0a77b63525fba0b91e494270e237c83472f8bb91

PR #86, opened 2016-01-20
Add method to clear options set by SSL_set_options

blucas:
Motivation:

We currently provide a way to set SSL options via ```SSL_set_options``` but we do not provide a way to clear out those options. Related to https://github.com/netty/netty/issues/4736

Modifications:

Add clearOptions function which executes ```SSL_clear_options```

Result:

Allow users to clear out options which were set using ```SSL_set_options```


blucas:
@normanmaurer PTAL

Scottmitch:
Cherry-picked 4edc0db

Thanks @blucas !

normanmaurer:
@blucas thanks!

PR #87, opened 2016-01-21
Don't use deprecated RSA_generate_key.

agl:
OpenSSL has deprecated RSA_generate_key and replaced it with
RSA_generate_key_ex. This change switches to use the new function.

RSA_generate_key was already deprecated in OpenSSL 0.9.8 so using the
new function should not break anything. Additionally, it means that the
code can drop an #ifdef that skipped this call with OpenSSL 1.1.0,
presumably because the deprecated function was causing compiler issues.

### DIFF
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -227,26 +227,35 @@ static const jint supported_ssl_opts = 0
 
 static int ssl_tmp_key_init_rsa(int bits, int idx)
 {
-#if (OPENSSL_VERSION_NUMBER < 0x10100000L) || defined(OPENSSL_USE_DEPRECATED)
-    if (!(SSL_temp_keys[idx] =
-          RSA_generate_key(bits, RSA_F4, NULL, NULL))) {
-#ifdef OPENSSL_FIPS
-        /**
-         * With FIPS mode short RSA keys cannot be
-         * generated.
-         */
-        if (bits < 1024)
-            return 0;
-        else
-#endif
-        return 1;
-    }
-    else {
+#if defined(OPENSSL_FIPS)
+    /**
+     * Short RSA keys cannot be generated in FIPS mode.
+     */
+    if (bits < 1024) {
+        SSL_temp_keys[idx] = NULL;
         return 0;
     }
-#else
-    return 0;
 #endif
+
+    BIGNUM *e = BN_new();
+    RSA *rsa = RSA_new();
+    int ret = 1;
+
+    if (e == NULL ||
+        rsa == NULL ||
+        !BN_set_word(e, RSA_F4) ||
+        RSA_generate_key_ex(rsa, bits, e, NULL) != 1) {
+        goto err;
+    }
+
+    SSL_temp_keys[idx] = rsa;
+    rsa = NULL;
+    ret = 0;
+
+err:
+    BN_free(e);
+    RSA_free(rsa);
+    return ret;
 }
 
 static int ssl_tmp_key_init_dh(int bits, int idx)
@@ -744,6 +753,7 @@ TCN_IMPLEMENT_CALL(jint, SSL, initialize)(TCN_STDARGS, jstring engine)
 
     SSL_TMP_KEYS_INIT(r);
     if (r) {
+        ERR_clear_error();
         TCN_FREE_CSTRING(engine);
         ssl_init_cleanup(NULL);
         tcn_ThrowAPRException(e, APR_ENOTIMPL);


agl:
Please note: I've only checked that this compiles because I'm not sure how to test it.

This came as a result of the deprecated `RSA_generate_key` not being available in Chromium. This whole area of the code appears to be related to export ciphersuites and could probably be dropped. If you think that's a change that you would be interested in having, please let me know. The referenced patch is the parsimonious approach, which should be appropriate in either case.

Scottmitch:
Thanks for the patch @agl! Have you signed the [ICLA](https://docs.google.com/forms/d/1GKuaUw10DfPzPy_XBnp3Wk_OLYnA4-Kg16wSiUkO3-A/viewform?formkey=dHBjc1YzdWhsZERUQnhlSklsbG1KT1E6MQ)? Also would you mind reformatting your commit message to the [format used in Netty](http://netty.io/wiki/writing-a-commit-message.html)?

src/main/c/ssl.c
```diff
@@ -231,26 +231,35 @@ static const jint supported_ssl_opts = 0
...
-    return 0;
 #endif
+
+    BIGNUM *e = BN_new();
+    RSA *rsa = RSA_new();
+    int ret = 1;
+
+    if (e == NULL ||
+        rsa == NULL ||
+        !BN_set_word(e, RSA_F4) ||
```
Scottmitch: is there a reason for using boolean type check here and then `!= 1` check below? [BN_set_word](https://www.openssl.org/docs/manmaster/crypto/BN_zero.html) is defined as `0 = failure 1 = success`.

src/main/c/ssl.c
```diff
@@ -231,26 +231,35 @@ static const jint supported_ssl_opts = 0
...
+    }
+
+    SSL_temp_keys[idx] = rsa;
+    rsa = NULL;
+    ret = 0;
+
+err:
+    BN_free(e);
+    RSA_free(rsa);
+    return ret;
```
Scottmitch: should we also clean up the error queue ([old RSA_generate_key docs say error queue may be used](https://www.openssl.org/docs/manmaster/crypto/RSA_generate_key.html)? This method is called externally by `generateRSATempKey` (although we currently don't use it Netty) but is also called internally in a few places. Maybe create an an internal version which atomically cleans up error queue, and another method that does not (designed to be called from JNI)?

src/main/c/ssl.c
```diff
@@ -231,26 +231,35 @@ static const jint supported_ssl_opts = 0
...
-         * generated.
-         */
-        if (bits < 1024)
-            return 0;
-        else
-#endif
-        return 1;
-    }
-    else {
+#if defined(OPENSSL_FIPS)
```
Scottmitch: I don't have much context here but I see in the 1_0_2 and 1_0_1 branch version of `RSA_generate_key_ex` also has some `OPENSSL_FIPS` macros (see [rsa_gen.c](https://github.com/openssl/openssl/blob/OpenSSL_1_0_2-stable/crypto/rsa/rsa_gen.c#L88)). Do we still need this special case? It looks like 1_0_0 doesn't have these defines though ... so maybe safer to keep these defines (if we are sure they are correct).

src/main/c/ssl.c
agl: OpenSSL isn't consistent about return values and sometimes returns negative values too.

I would generally use the boolean-style everywhere except that RSA_generate_key[_ex] is a function that can be redirected to a different method, so I fear that such code might end up returning negative numbers even though the default code does not. It would also be a little hard to criticise such code since the return value for _ex is [not documented](https://www.openssl.org/docs/manmaster/crypto/RSA_generate_key.html).

src/main/c/ssl.c
agl: The original version of this function didn't clear the queue.

As you note, it's called from several places so clearing the error queue uniformly is not obviously correct. Rather I've added a call to `ERR_clear_error` in `initialize` to take care of this.

src/main/c/ssl.c
agl: This Netty code is in the first commit (5404025e0e675bbd4762ef2e0636985532272f53). Grepping for `1024` in OpenSSL's `crypto/rsa` only shows that it's the default size so I'm not actually sure where this 1024-bit limit comes from in the OpenSSL code.

agl:
(I've signed the CLA and have updated the commit message.)

src/main/c/ssl.c
Scottmitch: @normanmaurer - Looks like `ssl_tmp_key_init_rsa` (and `generateRSATempKey`) have been removed from tc-native as of 1.2.0. Since we don't use it perhaps we should remove it...

Scottmitch: `ssl_tmp_key_init_dh` was also removed ... as suggested by @agl https://github.com/netty/netty-tcnative/pull/87#issuecomment-173722497 maybe we should just remove all related methods as was done in tcnative upstream.

agl: Certainly either path works for me :) This change is, at least, more parsimonious for now.

johnou:
@agl would you please rebase?

normanmaurer:
@agl no need to rebase.

johnou:
@normanmaurer figured it would save you some time :+1:

Scottmitch:
@normanmaurer - should we pull this in?

normanmaurer:
@Scottmitch if you are satisfied we can I guess...

Scottmitch:
@agl - can you rebase and we can pull this in?

agl:
(I've done a simple rebase on the original change.)

Scottmitch:
1.1.33 (73cb21a)

Thanks @agl !

PR #89, opened 2016-01-24
Make boringssl profile statically link APR

nmittler:
Motivation:

The boingssl profile statically links boringssl, but dynamically links APR, which means that the user still may have an extra step in installing APR onto the system before being able to use tcnative.

In addition, there was a bug with the boringssl profile in that the JNI lib was created with the name "netty-tcnative-boringssl", since it uses the artifactId by default. This prevents successful loading of the library in netty, since it's hard-coded to load "netty-tcnative".

Modifications:

Modified the pom.xml to download and build APR static libraries only.

Also manually setting the JNI library name to "netty-tcnative", regardless of artifactId.

Result:

BoringSSL profile is now completely stand-alone.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -110,6 +110,7 @@
           <execution>
             <id>build-native-lib</id>
             <configuration>
+              <name>netty-tcnative</name>
               <nativeSourceDirectory>${project.basedir}/src/main/c</nativeSourceDirectory>
               <libDirectory>${nativeLibOnlyDir}</libDirectory>
               <forceAutogen>${forceAutogen}</forceAutogen>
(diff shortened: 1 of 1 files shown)


nmittler:
FYI @Scottmitch @normanmaurer @ejona86 @louiscryan ... it works now!!! :)

normanmaurer:
Awesome! What you think about provide the same for openssl and libressl? I can take care 👍

louiscryan:
Awesome!

nmittler:
@normanmaurer +1 for openssl/libressl ... +2 for you taking care of it :stuck_out_tongue_winking_eye:

normanmaurer:
Cherry-picked as 20515acf6a5949f5d988e70cbcb003dbd46fa956

nmittler:
woot! :)

Scottmitch:
good work @nmittler !

PR #90, opened 2016-01-25
Fix m4 warnings

nmittler:
Motivation:

When running with verbose=true, there are a few warnings from the tcnative.m4 due to underquoted method definitions.

Modifications:

Adding proper quoting to method definitions in tcnative.m4.

Result:

No more warnings :)


nmittler:
@normanmaurer FYI

normanmaurer:
Cherry-Picked as da346d51331b89318e87398d64ff6c25def8f3c7 into 1.1.33

PR #91, opened 2016-01-27
Use SSL_get_session to access the SSL_SESSION for a SSL pointer so it…

normanmaurer:
… works also on LibreSSL

Motivation:

We use ssl->_session to access the SSL_SESSION, which is not a standard API and not works one LibreSSL

Modifications:

Use SSL_get_session to access the SSL_SESSION.

Result:

Code also works on LibreSSL


normanmaurer:
Cherry-picked into 1.1.33 as fd68c837b156ddb4b054e03d99a401e93068b34d

PR #92, opened 2016-01-27
Add profiles to staticially link against libressl and openssl

normanmaurer:
Motivation:

Sometimes it is useful to be able to statically link against libressl and openssl (the same as we provide a profile to do this against boringssl).

Modifications:

- Rename boringssl profile to boringssl-static
- Add libressl-static profile which will build a new tcnative jar that is statically linked against libressl
- Add openssl-static profile which will build a new tcnative jar that is statically linked against openssl

Result:

Easier building of static linked tcnative.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,6 +49,11 @@
     <nativeLibOnlyDir>${project.build.directory}/native-lib-only</nativeLibOnlyDir>
     <nativeJarWorkdir>${project.build.directory}/native-jar-work</nativeJarWorkdir>
     <maven-hawtjni-plugin-version>1.11</maven-hawtjni-plugin-version>
+    <maven-scm-plugin-version>1.9.4</maven-scm-plugin-version>
+    <ant-commons-net-version>1.9.6</ant-commons-net-version>
+    <aprHome>${project.build.directory}/apr</aprHome>
+    <aprTag>1.5.2</aprTag>
+    <aprDeveloperConnectionUrl>scm:svn:http://svn.apache.org/repos/asf/apr/apr</aprDeveloperConnectionUrl>
   </properties>
 
   <build>
@@ -204,22 +209,20 @@
   <profiles>
     <!-- Build distribution statically linked to BoringSSL -->
     <profile>
-      <id>boringssl</id>
+      <id>boringssl-static</id>
 
       <properties>
-        <artifactIdToUse>netty-tcnative-boringssl</artifactIdToUse>
+        <artifactIdToUse>netty-tcnative-boringssl-static</artifactIdToUse>
         <boringsslHome>${project.build.directory}/boringssl</boringsslHome>
         <boringsslBuildDir>${boringsslHome}/build</boringsslBuildDir>
         <boringsslBranch>chromium-stable</boringsslBranch>
-        <aprHome>${project.build.directory}/apr</aprHome>
-        <aprTag>1.5.2</aprTag>
       </properties>
 
       <build>
         <plugins>
           <plugin>
             <artifactId>maven-scm-plugin</artifactId>
-            <version>1.9.4</version>
+            <version>${maven-scm-plugin-version}</version>
             <executions>
               <!-- Download the BoringSSL source -->
               <execution>
@@ -246,7 +249,7 @@
                 <configuration>
                   <checkoutDirectory>${aprHome}</checkoutDirectory>
                   <connectionType>developerConnection</connectionType>
-                  <developerConnectionUrl>scm:svn:http://svn.apache.org/repos/asf/apr/apr</developerConnectionUrl>
+                  <developerConnectionUrl>${aprDeveloperConnectionUrl}</developerConnectionUrl>
                   <scmVersion>${aprTag}</scmVersion>
                   <scmVersionType>tag</scmVersionType>
                 </configuration>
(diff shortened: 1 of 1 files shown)


normanmaurer:
@trustin @Scottmitch @nmittler PTAL. I had some discussion about this with @Scottmitch and we were wondering if we should better not include the version of libressl and openssl in the pom.xml and let the user specify it on the cmdline (via '-D...'). This way we would clearly signal to the user that it's his/her responsibility to ensure no libressl or openssl version is used which has a know vuln.

Also these profiles should not be used to publish any releases and are only in the pom.xml to make it easy for a user to build a static linked version of netty-tcnative. Also once this this is merged I plan to setup a CI job to basically run:

```
mvn clean install -Pboringssl
mvn clean install -Popenssl
mvn clean install -Plibressl
```
After this is done we should also run the netty test suite with all of these. This will ensure our code works the same way with these different ssl flavours.

WDYT ?

pom.xml
```diff
@@ -204,22 +209,20 @@
   <profiles>
     <!-- Build distribution statically linked to BoringSSL -->
     <profile>
-      <id>boringssl</id>
+      <id>boringssl-static</id>
```
nmittler: I assume you're adding "-static" with the idea that we might eventually add a profile to build dynamically against Boring/Libre/Open?

Given that the artifact name is going to be `netty-tcnative-boringssl`, we should probably be consistent with the use of "-static" in the name for both the artifact and the profile name. WDYT?

normanmaurer: Yeah... Sounds good

nmittler:
@normanmaurer Looks great!

> I had some discussion about this with @Scottmitch and we were wondering if we should better not include the version of libressl and openssl in the pom.xml and let the user specify it on the cmdline (via '-D...'). This way we would clearly signal to the user that it's his/her responsibility to ensure no libressl or openssl version is used which has a know vuln.

If you specify the version as a property in the pom.xml, the user can still override it from the command-line, correct? That seems like a win-win, no?

> Also these profiles should not be used to publish any releases

I disagree ... we definitely need to publish these artifacts and should probably just do so as a part of the normal tcnative release cycle.  If they're not available on Maven central, then it's still going to be a major pain point for users.  We'll just need to specify in the release notes the version of XXXssl and APR that were used.  We don't have to publish all of them, necessarily ... I'd be happy just picking one that we all agree on.  WDYT?

Scottmitch:
> If you specify the version as a property in the pom.xml, the user can still override it from the command-line, correct? That seems like a win-win, no?

The concern is if we provide a version number in our code, we are responsible for making sure this version stays current and is not susceptible to any known vulnerabilities. We don't really have a reliable process (immediate/early access to disclosures, mailing lists?, etc...) setup for OpenSSL and LibreSSL to effectively handle this. Perhaps if we have these then it would be more manageable.

>  disagree ... we definitely need to publish these artifacts and should probably just do so as a part of the normal tcnative release cycle.

I think the motivation for not publishing these artifacts is who can effectively manage the versions. I think we can publish for BoringSSL because you and @louiscryan's story around managing disclosures (you are our inside men :wink:) and vested interest to maintain this.

> We'll just need to specify in the release notes the version of XXXssl and APR that were used.

+1 even for BoringSSL.

nmittler:
>> If you specify the version as a property in the pom.xml, the user can still override it from the command-line, correct? That seems like a win-win, no?

> The concern is if we provide a version number in our code, we are responsible for making sure this version stays current and is not susceptible to any known vulnerabilities. We don't really have a reliable process (immediate/early access to disclosures, mailing lists?, etc...) setup for OpenSSL and LibreSSL to effectively handle this. Perhaps if we have these then it would be more manageable.

I think documenting the usage of the profile on the wiki should be sufficient for users that are building their own tcnative, along with a disclaimer that they should probably check for the latest APR, XXXssl to get the latest security updates.

WRT the Netty team, maintenance and concern for vulnerabilities really should only be an issue for deployed artifacts (i.e. BoringSSL-static). Here, Google is signing up to make sure that tcnative is updated and new statically linked artifacts are pushed as needed. So I think we can assume that someone (e.g. @agl, @louiscryan, @ejona86, @nmittler) from Google will ping you guys when we need a new binary.  With that said, security patches aren't likely to be all that frequent, so I wouldn't expect this to be a major burden for whomever is pushing jars. Does that sound reasonable?

Scottmitch:
>  Here, Google is signing up to make sure that tcnative is updated and new statically linked artifacts are pushed as needed

Yes as I said I think we are good for BoringSSL, but we don't have the same story for LibreSSL and OpenSSL. There is some code overlap, but it is not necessarily true that a vulnerability in LibreSSL and/or OpenSSL equates to a vulnerability in BoringSSL.

nmittler:
@Scottmitch Yup, agreed.  As discussed offline, we'll require version parameters for OpenSSL and LibreSSL (but not BoringSSL).  Thanks guys!

normanmaurer:
Let me ensure I can build all of these on windows as well before pull in :)

nmittler:
@normanmaurer +1 :) ... I'm fine with a follow-up PR for windows too if you'd prefer.

I suspect BoringSSL should build on windows, since I used build ninja.  APR may be a problem.

normanmaurer:
Cherry-picked into 1.1.33 as 25eced05edcb9104fc79c82854ca2d447bbb72fe

PR #93, opened 2016-01-27
Fix possible segfault when using LibreSSL and not SSL2

normanmaurer:
Motivation:

We used a non API to get the cipher during the certificate verification which worked in OpenSSL, BoringSSL and older versions of LibreSSL. In newer versions of LibreSSL this leads to a segfault.

Modifications:

Use public api to get the cipher.

Result:

No more segfault when using newer versions of LibreSSL.

### DIFF
--- a/src/main/c/sslcontext.c
+++ b/src/main/c/sslcontext.c
@@ -30,6 +30,8 @@
 #ifdef HAVE_OPENSSL
 #include "ssl_private.h"
 
+static const char* UNKNOWN_AUTH_METHOD = "UNKNOWN";
+
 static jclass byteArrayClass;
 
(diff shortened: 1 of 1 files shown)


normanmaurer:
@nmittler PTAL 

@blucas @fbettag I think you reported this to me before as well.

fbettag:
Yes i did like a year ago :D but i'll recheck

normanmaurer:
@fbettag yeah... at this point I thought it would be a libressl bug but I was wrong :(

nmittler:
@normanmaurer LGTM .... ship it!

normanmaurer:
Cherry-picked into 1.1.33 as 5a4b3f636b40cf136f63c448e7ddf5f09935bf91

Issue #100, opened 2016-02-01
SSL set mode support


tanersener:
Hi,

Do you plan to support SSL_CTX_set_mode operation? In some cases enabling SSL_MODE_ENABLE_PARTIAL_WRITE and SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER modes is necessary but as you pointed out in https://github.com/netty/netty-tcnative/issues/81 SSLExt is not supported and there is not any other way to set them.

https://www.openssl.org/docs/manmaster/ssl/SSL_CTX_set_mode.html

Scottmitch:
Are you using this library standalone or with Netty?

tanersener:
I'm using it with netty 4.0.x.

normanmaurer:
Can you tell us why you need to adjust these?

tanersener:
Well, my case is; I'm receiveing "error:1409F07F:SSL routines:SSL3_WRITE_PENDING: bad write retry" error on high load. After some google digging I found the following thread which suggests me to enable these flags. That's why I'm asking. 

http://stackoverflow.com/questions/2997218/why-am-i-getting-error1409f07fssl-routinesssl3-write-pending-bad-write-retr#

normanmaurer:
@tanersener this sound more like a netty bug. Can you share a reproducer or / and the full stack trace ?

normanmaurer:
@tanersener actually I think you are right and we need set `SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER` because if the user writes an heap-buffer the memoryAddress will most likely change between SSL_write(...) operations on retry. This is because in this case we will copy from heap to direct buffer before call SSL_write(...).

https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/OpenSslEngine.java#L287

Let me work on a fix for this.

normanmaurer:
@tanersener done via https://github.com/netty/netty-tcnative/pull/101. Will make use of this in netty once merged and released.

tanersener:
@normanmaurer you are very fast thanks. I just tested your pull request with commit from netty side locally and it worked for me. SSL_write error is gone. 
Any chance of backporting this feature to 4.0.x line?

normanmaurer:
@tanersener awesome!

Yes it will be applied to 4.0 and 4.1

tanersener:
@normanmaurer cool, thanks again

normanmaurer:
[...]

PR #101, opened 2016-02-02
[#100] Allow to call SSL_CTX_set_mode and SSL_CTX_get_mode

normanmaurer:
Motivation:

SSL_CTX_set_mode and SSL_CTX_get_mode may need to be called from the Java layer to adjust various mode options.

Modifications:

Expose native methods.

Result:

It's noew possible to call SSL_CTX_set_mode and SSL_CTX_get_mode.

### DIFF
--- a/src/main/c/sslcontext.c
+++ b/src/main/c/sslcontext.c
@@ -1591,6 +1591,27 @@ TCN_IMPLEMENT_CALL(jboolean, SSLContext, setSessionIdContext)(TCN_STDARGS, jlong
     }
     return JNI_FALSE;
 }
+
+TCN_IMPLEMENT_CALL(jint, SSLContext, setMode)(TCN_STDARGS, jlong ctx, jint mode)
+{
+    tcn_ssl_ctxt_t *c = J2P(ctx, tcn_ssl_ctxt_t *);
+
+    UNREFERENCED(o);
+    TCN_ASSERT(ctx != 0);
+
+    return (jint) SSL_CTX_set_mode(c->ctx, mode);
+}
+
+
+TCN_IMPLEMENT_CALL(jint, SSLContext, getMode)(TCN_STDARGS, jlong ctx)
+{
+    tcn_ssl_ctxt_t *c = J2P(ctx, tcn_ssl_ctxt_t *);
+
+    UNREFERENCED(o);
+    TCN_ASSERT(ctx != 0);
+
+    return (jint) SSL_CTX_get_mode(c->ctx);
+}
 #else
 /* OpenSSL is not supported.
  * Create empty stubs.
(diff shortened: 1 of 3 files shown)


normanmaurer:
@nmittler @Scottmitch @trustin PTAL

nmittler:
LGTM

src/main/c/sslcontext.c
```diff
@@ -1591,6 +1591,27 @@ TCN_IMPLEMENT_CALL(jboolean, SSLContext, setSessionIdContext)(TCN_STDARGS, jlong
     }
     return JNI_FALSE;
 }
+
+TCN_IMPLEMENT_CALL(jint, SSLContext, setMode)(TCN_STDARGS, jlong ctx, jint mode)
+{
+    tcn_ssl_ctxt_t *c = J2P(ctx, tcn_ssl_ctxt_t *);
+
+    UNREFERENCED(o);
```
Scottmitch: Can you explain what this does? In some cases we do this after the null check/return.

https://github.com/netty/netty-tcnative/blob/1.1.33/src/main/c/ssl.c#L1600

normanmaurer: @Scottmitch it's a macro that basically just "unreference" the JNIEnv. Honestly I think it is not really important to call, but "original" tcnative tends to do it.

Scottmitch: @normanmaurer - Not sure I full understand.

```c
#define TCN_STDARGS     JNIEnv *e, jobject o
#define UNREFERENCED(P)      (P) = (P)
```

so wont the `UNREFERENCED` macro expand to:

```c
o = o;
```

where `o` is the `jclass` class argument?

normanmaurer: yeah I think so... So not sure what exactly it is supposed to do as well.

normanmaurer: @Scottmitch I think let us just keep it as it is to be consistent with the rest of tcnative. That said I want to do some more cleanup later on in the whole project. But this is out of scope here.

Scottmitch: @normanmaurer +1. I was more concerned about if it was harmful or not ... but doesn't seem that way(?).

Scottmitch:
lgtm

src/main/c/sslcontext.c
Scottmitch: see https://github.com/netty/netty-tcnative/issues/102

normanmaurer:
Cherry-picked into 1.1.33 as e80b2b3a3784943ee102b9ed3c27736ba7251c47

Scottmitch:
Good work @normanmaurer! Delete branch when satisfied.

PR #106, opened 2016-02-05
sslcontext.make doesn't work if sslv2 and sslv3 are disabled

Scottmitch:
Motivation:
The intialization logic in sslcontext.make does not initialize the SSL_CTX if sslv2 and sslv3 have been disabled during compile time. In particular using SSL_PROTOCOL_ALL as a protocol won't even attempt to call SSL_CTX_new.

Modifications:
- Fix sslcontext.make so the ifdefs and if/else statements correctly attempt to initialize the SSL_CTX object

Result:
sslcontext.make respects the compile time configuration of openssl and still correctly initializes the SSL_CTX object at runtime.


Scottmitch:
@normanmaurer - FYI

normanmaurer:
@Scottmitch Please fix the build errors:

```
[20:23:28][io.netty:netty-tcnative] [INFO] cc1: warnings being treated as errors
[20:23:28][io.netty:netty-tcnative] [INFO] src/sslcontext.c: In function 'Java_org_apache_tomcat_jni_SSLContext_make':
[20:23:28][io.netty:netty-tcnative] [INFO] src/sslcontext.c:115: error: passing argument 1 of 'SSL_CTX_new' makes pointer from integer without a cast
[20:23:28][io.netty:netty-tcnative] [INFO] /usr/include/openssl/ssl.h:1629: note: expected 'const struct SSL_METHOD *' but argument is of type 'int'
[20:23:28][io.netty:netty-tcnative] [INFO] src/sslcontext.c:117: error: passing argument 1 of 'SSL_CTX_new' makes pointer from integer without a cast
[20:23:28][io.netty:netty-tcnative] [INFO] /usr/include/openssl/ssl.h:1629: note: expected 'const struct SSL_METHOD *' but argument is of type 'int'
[20:23:28][io.netty:netty-tcnative] [INFO] src/sslcontext.c:119: error: passing argument 1 of 'SSL_CTX_new' makes pointer from integer without a cast
[20:23:28][io.netty:netty-tcnative] [INFO] /usr/include/openssl/ssl.h:1629: note: expected 'const struct SSL_METHOD *' but argument is of type 'int'
[20:23:28][io.netty:netty-tcnative] [INFO] make: *** [src/sslcontext.lo] Error 1
```

Scottmitch:
https://www.openssl.org/docs/manmaster/ssl/SSL_CTX_new.html
> SSLv23_method(), SSLv23_server_method(), SSLv23_client_method()
Use of these functions is deprecated. They have been replaced with TLS_method(), TLS_server_method() and TLS_client_method() respectively. New code should use those functions instead.

Apparently these new macros/functions are not in 1.0.1

normanmaurer:
We should support 1.0.1 as well

Scottmitch:
@normanmaurer - Yip. I took out the `TLS_*method` references and used the `SSLv23_*method` references that were already in use.

normanmaurer:
@Scottmitch LGTM

normanmaurer:
Cherry-picked into 1.1.33 as c9e7cde8d5bd52e338637a08c76ac284aff44683. 

@Scottmitch thanks!

PR #107, opened 2016-02-11
Use tar command on linux

nmittler:
Motivation:

The current configuration for openssl-static uses the ant untar task, which does not retain file permissions. To get around this, we then run the chmod task to recursively make all files executable in the extracted directory. This seems to be causing issues on mac.

Modifications:

Switch to using the tar command (rather than the ant task) when extracting the openssl source on linux/mac.

Result:

Mac builds :)

### DIFF
--- a/openssl-static/pom.xml
+++ b/openssl-static/pom.xml
@@ -37,43 +37,6 @@
 
   <build>
     <plugins>
-      <plugin>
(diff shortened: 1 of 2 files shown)


nmittler:
@normanmaurer PTAL

normanmaurer:
@nmittler LGTM.

nmittler:
Cherry-picked as 16af398

PR #108, opened 2016-02-11
Adding ssl/apr version to manifest

nmittler:
Motivation:

It would be helpful to be able to inspect the statically-linked jars to determine exactly which version of APR/SSL they were built against.

Modifications:

Adding additional entries via the maven-bundle-plugin for the statically linked jars.

Result:

The jars provide version information for APR and SSL

### DIFF
--- a/boringssl-static/pom.xml
+++ b/boringssl-static/pom.xml
@@ -63,6 +63,36 @@
         </executions>
       </plugin>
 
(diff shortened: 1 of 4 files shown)


nmittler:
@normanmaurer PTAL

normanmaurer:
@nmittler looking good!

nmittler:
cherry-picked as 928246a

PR #109, opened 2016-02-12
Only deploy static jar on centos

nmittler:
Motivation:

Our current deploy process requires that we start on centos first. Since we only generate one boring-static jar for linux, this means that it will be overwritten when we get to debian. Debian has a higher glibc version, which will release a jar that ultimately won't work on old versions of centos.

Modifications:

Modified the boringssl module to only deploy if on centos.

Result:

Our deploy process works :)


normanmaurer:
@nmittler ship it...

PR #113, opened 2016-02-16
Add option to build BoringSSL uber jar

nmittler:
Motivation:

It would be useful to have a single uber jar that contains the native libraries for all the platforms. This would simplify the user's "getting started" process.

Modifications:

Added an uber profile to boringssl-static.

Result:

We can build a uber jar that is statically linked to BoringSSL and APR for osx-64, linux-64, and windows-64.

### DIFF
--- a/boringssl-static/pom.xml
+++ b/boringssl-static/pom.xml
@@ -30,165 +30,321 @@
     to BoringSSL and Apache APR.
   </description>
 
-  <properties>
-    <boringsslCheckoutDir>${project.build.directory}/boringssl-${boringsslBranch}</boringsslCheckoutDir>
-    <boringsslBuildDir>${boringsslCheckoutDir}/build</boringsslBuildDir>
-    <linkStatic>true</linkStatic>
-    <msvcSslIncludeDirs>${boringsslCheckoutDir}/include</msvcSslIncludeDirs>
-    <msvcSslLibDirs>${boringsslBuildDir}/ssl;${boringsslBuildDir}/crypto;${boringsslBuildDir}/decrepit</msvcSslLibDirs>
-    <msvcSslLibs>ssl.lib;crypto.lib;decrepit.lib</msvcSslLibs>
-  </properties>
+  <profiles>
+    <!-- Default profile that builds a platform-specific jar -->
+    <profile>
+      <id>boringssl-static-default</id>
+      <activation>
+        <activeByDefault>true</activeByDefault>
+      </activation>
 
-  <build>
-    <plugins>
-      <!-- Download the BoringSSL source -->
-      <plugin>
-        <artifactId>maven-scm-plugin</artifactId>
-        <executions>
-          <execution>
-            <id>get-boringssl</id>
-            <phase>generate-sources</phase>
-            <goals>
-              <goal>checkout</goal>
-            </goals>
+      <properties>
+        <boringsslCheckoutDir>${project.build.directory}/boringssl-${boringsslBranch}</boringsslCheckoutDir>
+        <boringsslBuildDir>${boringsslCheckoutDir}/build</boringsslBuildDir>
+        <linkStatic>true</linkStatic>
+        <msvcSslIncludeDirs>${boringsslCheckoutDir}/include</msvcSslIncludeDirs>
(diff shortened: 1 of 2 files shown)


nmittler:
@normanmaurer @louiscryan @trustin @cconroy  FYI

normanmaurer:
@nmittler I guess this also will need a change in the NativeLibraryLoader to respect this right ?

normanmaurer:
@nmittler btw why call it `mongo` ? I though this kind of stuff was usually called `uber-jar`

trustin:
@normanmaurer @nmittler I wonder if there's a way to make the artifact without classifier contain the shared objects of all platforms, instead of introducing a uber/mongo artifactId.

trustin:
For example, after deploying all platform-specific artifacts (the ones with classifiers), we could run an additional step to create a uber/mongo artifact from the deployed artifacts and then to re-deploy the mongorized artifact without a classifier.

trustin:
To avoid any chance of making mistakes, we could set up a TeamCity job for performing a release. (i.e. we need the TeamCity agents for all our supported platforms.

nmittler:
@normanmaurer  - yep you're right, it should be "uber" ... I'll change it. And yes, Netty's OpenSSL class will have to change to look for the platform-specific native libraries (should be an easy fix).

@trustin +1 for overwriting the no-classifier jar with the uber jar.  I'll look into it.

nmittler:
@normanmaurer @trustin PTAL. I've moved the uber jar profile into the boringssl-static pom so that it can overwrite the default jar.

nmittler:
@normanmaurer @trustin One thing I'm not sure of is whether we can actually download from the Maven staging repo. If we can, this should be fairly simple... we would just build the uber jar last, pointing it at the staging repo so that it can get all of the platform-specific jars.

If we can't download from the staging repo, we have 2 options that I see:

1. Manually copy the platform-specific jars to the local maven repo, so that we can build and push the uber jar.  This will make releasing complicated and difficult to automate.
2. Push the uber jar after we have released. Not sure if Maven will allow us to overwrite a released jar or not.  If not, we'll have to use a separate artifact-id or classifier in this case.

Thoughts?

nmittler:
It looks like the staging repo is publicly accessible.  I've updated the uber profile to use the staging repo.

nmittler:
@trustin @normanmaurer want to take another look?

boringssl-static/pom.xml
```diff
@@ -30,164 +30,312 @@
...
+            <groupId>org.codehaus.mojo</groupId>
+            <artifactId>buildnumber-maven-plugin</artifactId>
+            <executions>
+              <execution>
+                <phase>generate-sources</phase>
+                <goals>
+                  <goal>create</goal>
+                </goals>
+                <configuration>
+                  <scmDirectory>${boringsslCheckoutDir}</scmDirectory>
```
trustin: Could we also use an alternative property name so it's less confusing? `buildNumber` -> `boringsslBuildNumber` for example. http://www.mojohaus.org/buildnumber-maven-plugin/create-mojo.html#buildNumberPropertyName

trustin:
Just one nit. Great job as usual, @nmittler !

boringssl-static/pom.xml
nmittler: done.

nmittler:
@normanmaurer anything else before I cherry-pick?

Also, just to note that deployment is hard to test ... so we may have to tweak a few things once we get to the point of releasing, but I suspect it shouldn't be too bad.

WRT releasing, should we add the change to Netty to optionally load the platform-specific jars before we release this? WDYT?

nmittler:
@normanmaurer See the required changes to Netty: #4887.  That should get committed first.

normanmaurer:
@nmittler LGTM... Great work buddy!

nmittler:
Cherry-picked as f0bafbc

johnou:
@normanmaurer @trustin @nmittler I would like to see support for an openssl (dynamic) uber jar, however, are there any preferences on how the artifact should be named? uber jar for boring ssl overwrites the no-classifier jar but I guess we may want dynamic / static uber jars..?

nmittler:
I suppose we could just modify the default  `netty-tcnative` jar to be the uber jar, as we did for boringssl-static. With that said, I'm not against a rename to something like `netty-tcnative-dynamic` to make it more immediately obvious which jar you have.

@normanmaurer @trustin WDYT?

trustin:
@nmittler SGTM

PR #117, opened 2016-02-19
Set the correct cause of the verification failure

normanmaurer:
Motivation:

We should support setting the exact cause of the verification failure which may help the remote peer to fix it.

Modifications:

Change CertificateVerifier to allow set the actual cause.

Result:

Give more control about certificate verification


normanmaurer:
@nmittler @Scottmitch PTAL

nmittler:
LGTM

normanmaurer:
Cherry-picked into 1.1.33 as (26fa1429c92238ef28e2e6b2d93112bfa1063ef4)

openssl-dynamic/src/main/java/org/apache/tomcat/jni/CertificateVerifier.java
```diff
@@ -20,6 +20,73 @@
  */
 public interface CertificateVerifier {
 
+    int X509_V_OK = 0;
```
Scottmitch: would be nice to have a link as to where these were derived from ... I suspect some openssl header file though :wink:

Scottmitch:
lgtm

openssl-dynamic/src/main/java/org/apache/tomcat/jni/CertificateVerifier.java
normanmaurer: haha yeah... ;) Let me add a link

normanmaurer: @Scottmitch actually these are from: 
https://www.openssl.org/docs/manmaster/apps/verify.html

Scottmitch: OooO docs ... i guess they would also live in some SSL / TLS rfc somewhere too.

normanmaurer: :+1: I will check and add a follow pr

Issue #104, opened 2016-02-03
Remove unused code


normanmaurer:
I would like to propose removing all code that we not need for netty (like `Socket`, `Multicast` and many more). We are only interested in the parts that are needed for our OpenSslEngine anyway. This will help us to focus on things we care about and not ship things that we are not using.

@nmittler @Scottmitch @trustin WDYT ?

trustin:
+1

nmittler:
+1

Scottmitch:
+1. I think tcnative "upstream" has rebased off of our code anyways, which may make taking updates a bit more complicated but I'm not sure if we plan on taking updates.

normanmaurer:
Ok will take care of this

nmittler:
@normanmaurer did these commits make it into the 1.1.33 branch? I don't see them in the history

normanmaurer:
@nmittler nope... I need to rebase it figure out what I broke ;)

nmittler:
ah ok sgtm

davidben:
I started poking at this myself. (I'd rather you all not hard-code the values of `SSL_OP_*` constants and figured I ought to get dead code out of the way before tackling that.) GitHub's diff viewer is really slow on that patch and I'm not sure how to download it to be sure, but I think the issue is you removed `tcn_ThrowAPRException` which is called a lot in `ssl.c`. That, in turn, requires that `Error.java` stay there.

PR #122, opened 2016-02-25
[#104] Remove unused code

normanmaurer:
Motivation:

We only use a sub-set of the original tcnative code that we forked. We should remove everything we not care about.

Modification:

Remove all stuff that we not need and want to maintain.

Result:

Less code to maintain.


normanmaurer:
@Scottmitch @trustin @nmittler PTAL... After this is merged I will take a look what it will take to remove apr dependency and if make sense.

nmittler:
@normanmaurer Looks like it's literally all deletes.  Assuming it works, ship it!

+1 for less code!! :)

Scottmitch:
Soo much red ... ship it!

trustin:
Awesome :-)

normanmaurer:
seems like I fucked up somehow... this not works :( Let me close and re-open once figured out.

PR #127, opened 2016-03-23
Use latest 6.x centos release for enforcing rule.

normanmaurer:
Motivation:

We currently use centos 6.6 when enforcing the linux version during release. We should use 6.7 which is the latest centos release out of the 6 series.

Modifications:

Enforce centos 6.7

Result:

Be able to release on latest centos version out of the 6 series.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -389,7 +389,7 @@
                       <files>
                         <file>/etc/redhat-release</file>
                       </files>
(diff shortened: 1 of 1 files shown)


nmittler:
LGTM

normanmaurer:
Cherry-picked as 8c247c5132a1bc0d48469027a5848902cce44853

PR #139, opened 2016-05-04
Update openssl to 1.0.2h and libressl to 2.3.4

normanmaurer:
Motivation:

New versions of openssl and libressl were released.

Modifications:

Update versions.

Result:

Use up to date versions.


normanmaurer:
Cherry-picked int 1.1.33 as 5ddbde652ca41f40485d8a55b773862f90394c39

PR #142, opened 2016-05-19
Replacing deprecated method on windows

nmittler:
Motivation:

strdup is deprecated on windows in favor of _strdup. The deprecated version of the method does not seem to be available when attempting to link with against libcmt.lib.

Modifications:

Changed jnilib.c to use _strdup instead.

Result:
Windows now resolves _strdup when linking against libcmt.lib.


nmittler:
@normanmaurer this is an easy one ... can you take a look?

normanmaurer:
@nmittler ship it while it's hot ;)

nmittler:
Cherry-picked as 9a4e6a9

Issue #129, opened 2016-03-23
Linux .so for boring isn't stripped


ejona86:
This is effectively a repeat of #45, but for boring and MB involved instead of KB. OS X's native library is 1.9 MB uncompressed, whereas Linux's is 7.1 MB. Running strip on the Linux .so reduces its size to 1.9 MB. I think that will change the compressed size from 2.5 MB to ~900 KB.

normanmaurer:
@nmittler any idea how to do this with ninja build ?

normanmaurer:
Actually I think I know howto fix this... let me take care

normanmaurer:
@davidben Could this be because `-ggdb` is used in boringssl here: https://boringssl.googlesource.com/boringssl/+/chromium-stable/CMakeLists.txt#30 ? (@nmittler told me to ping you ;))

davidben:
I could believe it. That's probably why there's so much to strip away. Having debug symbols available is generally useful and the standalone build, on our end, is used for development, so I'd want that to stay the default.

That said, it seems `CMAKE_BUILD_TYPE` distinguishes between "Release" and "RelWithDebInfo". I need to read up on how that knob is supposed to work, but I wouldn't be opposed to making only "RelWithDebInfo" and "Debug" generate debug symbols while plain "Release" does not.
https://cmake.org/cmake/help/v3.0/variable/CMAKE_BUILD_TYPE.html

Could you confirm that taking that bit out does the trick for you?

normanmaurer:
@davidben confirmed here...

Before change the CMakeLists.txt:

```
# cmake -DCMAKE_POSITION_INDEPENDENT_CODE=TRUE -DCMAKE_BUILD_TYPE=Release -DCMAKE_ASM_FLAGS=-Wa,--noexecstack -GNinja ..
# ninja
# du -h ssl/libssl.a
2,8M	ssl/libssl.a
```

After the change:

```
# cmake -DCMAKE_POSITION_INDEPENDENT_CODE=TRUE -DCMAKE_BUILD_TYPE=Release -DCMAKE_ASM_FLAGS=-Wa,--noexecstack -GNinja ..
# ninja
# du -h ssl/libssl.a                                                                                                    
472K	ssl/libssl.a
```

normanmaurer:
Also @nmittler @ejona86 :)

davidben:
Alright. Filed https://crbug.com/boringssl/56.

davidben:
[...]

PR #144, opened 2016-05-20
[#129] Ensure we strip static linked library on linux

normanmaurer:
Motivation:

We should ensure we strip the static linked library to save space.

Modification:

Call strip when needed.

Result:

Less space needed by the jar.

### DIFF
--- a/boringssl-static/pom.xml
+++ b/boringssl-static/pom.xml
@@ -137,13 +137,8 @@
 
                     <mkdir dir="${boringsslBuildDir}" />
 
-                    <!-- Determine if we're on windows -->
-                    <condition property="isWindows" else="false">
-                      <os family="windows" />
-                    </condition>
-
                     <if>
-                      <equals arg1="${isWindows}" arg2="true" />
+                      <equals arg1="${os.detected.name}" arg2="windows" />
                       <then>
                         <!-- On Windows, build with /MT for static linking -->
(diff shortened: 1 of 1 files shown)


normanmaurer:
@nmittler @ejona86 PTAL

boringssl-static/pom.xml
```diff
@@ -180,6 +180,19 @@
                 </goals>
                 <configuration>
                   <target>
+                    <!-- Add the ant tasks from ant-contrib -->
+                    <taskdef resource="net/sf/antcontrib/antcontrib.properties" />
+
+                    <!-- Strip on linux. See https://github.com/netty/netty-tcnative/issues/129 -->
+                    <if>
+                      <equals arg1="${os.detected.name}" arg2="linux" />
```
nmittler: This is better than what I did ... could you change line 146 to match this?

nmittler:
@normanmaurer LGTM

boringssl-static/pom.xml
normanmaurer: done

normanmaurer:
Cherry-picked into 1.1.33 as f809d61ad53151a29c0d79deb00cfa11c6480174

ejona86:
@normanmaurer, I'm late to the party, but LGTM

PR #153, opened 2016-06-09
Add session ticket key statistics to the context

cconroy:


normanmaurer:
Let me check

openssl-dynamic/src/main/c/ssl_private.h
```diff
@@ -302,9 +302,20 @@ struct tcn_ssl_ctxt_t {
...
+
+    // The client did not present a ticket and we issued a new one.
+    volatile long            ticket_keys_new;
+    // The client presented a ticket derived from the primary key
+    volatile long            ticket_keys_resume;
+    // The client presented a ticket derived from an older key, and we upgraded to the primary key.
+    volatile long            ticket_keys_renew;
+    // The client presented a ticket that did not match any key in the list.
+    volatile long            ticket_keys_fail;
 };
```
normanmaurer: I think just increment a volatile is not good enough. We would need a mutex or some atomic operation

Scottmitch: +1 increment volatile will likely not yield the intended results.

cconroy: Whoops! So used to doing this in java...

Do you know off hand if the context mutex is re-entrant or not? It's not immediately obvious from glancing at the apr docs...

normanmaurer: @cconroy even in java a volatile would be not good enough ;) Sorry, no I don't know about the apr one.

cconroy: Yep yep. Looks like not re-entrant but we need a write lock anyways. Fixing.

cconroy: Actually, it looks like apr has support for atomic increments....I will look into that instead.

cconroy: Sadly, the apr atomic stuff doesn't support `long` so a mutex it is i guess.

normanmaurer:
@cconroy do you plan to expose this in netty via the `OpenSslSessionContext` ?

cconroy:
@normanmaurer Yep we should definitely expose it after this lands+releases. In the meantime you can access via `SSLContext.sessionTicketKeyNew(serverContext.sslCtxPointer());` (that's what I did to test locally)

normanmaurer:
@cconroy got it... I wonder how you feel about the perf overhead it introduce caused by the mutex. WDYT ?

cconroy:
Indeed, ideally we wouldn't need a write lock here. Though, there's already similar code in openssl grabbing writelocks while adding a session and updating stats, etc.

I just spun a new version using the apr_uint32_t for atomic increments instead. I'd prefer atomic long increments, but uint32 should be fine...The intended purpose (producing metrics) should already have to handle wraparound to 0 for e.g. server restarts.

openssl-dynamic/src/main/java/org/apache/tomcat/jni/SSLContext.java
```diff
@@ -312,7 +312,21 @@ public static native boolean setCertificateBio(
     public static native long sessionTimeouts(long ctx);
 
     /**
-     * Set TLS session keys.
+     * TLS session ticket key resumption statistics.
+     */
+    public static native long sessionTicketKeyNew(long ctx);
+    public static native long sessionTicketKeyResume(long ctx);
+    public static native long sessionTicketKeyRenew(long ctx);
+    public static native long sessionTicketKeyFail(long ctx);
```
normanmaurer: should these now be int ?

cconroy: no: the underlying data is `uint32` so `int` does not have enough space. Also, the other stats are all `long`

normanmaurer: ah sorry true... LGTM :)

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -1330,6 +1331,34 @@ TCN_IMPLEMENT_CALL(jlong, SSLContext, sessionCacheFull)(TCN_STDARGS, jlong ctx)
...
+    jlong rv = apr_atomic_read32(&c->ticket_keys_renew);
+    return rv;
+}
+
+TCN_IMPLEMENT_CALL(jlong, SSLContext, sessionTicketKeyFail)(TCN_STDARGS, jlong ctx)
+{
+    tcn_ssl_ctxt_t *c = J2P(ctx, tcn_ssl_ctxt_t *);
+    jlong rv = apr_atomic_read32(&c->ticket_keys_fail);
+    return rv;
+}
```
normanmaurer: @cconroy can you also add the "unimplemented" variant for now for these methods ? Even if we will most likely get rid of all of these soon.

cconroy: done

openssl-dynamic/src/main/java/org/apache/tomcat/jni/SSLContext.java
Scottmitch: `long` and `int` are ill defined in terms of underlying size in the c specifications. If the underlying data is `uint32` you should use the same type.

normanmaurer:
@cconroy thanks again!

Cherry-picked as 16f99bf116803037d3c2b354c9e8e34838daebed

normanmaurer:
@cconroy let me take care to expose these in netty :)

cconroy:
Thanks @normanmaurer. Can you let me know what the ETA is for the fork 18 release?

normanmaurer:
@cconroy no idea yet... need it badly ?

cconroy:
I can easily cut a release internally if there aren't any immediate plans for a release. So, no rush!

PR #155, opened 2016-06-09
Upgrading to MSVC 2015

### LINKED ISSUE #134: Migrate BoringSSL build to MSVC 2015
@nmittler asked me to file a bug here when we were ready to switch BoringSSL's MSVC version. BoringSSL tracks Chromium's MSVC version which has since switched to MSVC 2015. It should currently build with either, but I'll be switching our bots to 2015 soonish and it'll likely (accidentally or intentionally) break with 2013 when we're no longer regularly testing against it.

nmittler:
Motivation:

BoringSSL has upgraded to MSVC 2015 and will not be supporting 2013 any longer.

Modifications:

Update the tooling in the MSVC project files.

Result:

Fixes #134


nmittler:
@normanmaurer PTAL

nmittler:
@normanmaurer I suspect the build failures will go away once we upgrade the box to MSVC 2015

normanmaurer:
Cherry-picked as 71ab9576c95d9c86766cce7a2092137cd7031617

PR #157, opened 2016-06-12
[#104] Remove unused code

normanmaurer:
Motivation:

We only use a sub-set of the original tcnative code that we forked. We should remove everything we not care about.

Modification:

Remove all stuff that we not need and want to maintain.

Result:

Less code to maintain.


normanmaurer:
@Scottmitch @nmittler second try... this one works. Once this is merged I will take another stab to tidy up `SSLContext` and `SSL`. Also I want to remove Pool and just handle the pool creation / destroy under the hood.

normanmaurer:
And as a follow up I would like to rename packages and rename the artifact so it not clashes anymore with the "original" tcnative.

normanmaurer:
@trustin also see above

trustin:
@normanmaurer w00t! Nice move.

nmittler:
@davidben want to take a look as well?

normanmaurer:
That said this is only removal of non SSL stuff so not sure how interested @davidben is

normanmaurer:
@Scottmitch ping :)

nmittler:
@normanmaurer ah I see ... let me take a look

openssl-dynamic/src/main/c/error.c
```diff
@@ -169,80 +163,4 @@ TCN_IMPLEMENT_CALL(jstring, Error, strerror)(TCN_STDARGS, jint err)
...
-        APR_IS(92, APR_STATUS_IS_ECONNABORTED);
-        APR_IS(93, APR_STATUS_IS_ECONNRESET);
-        APR_ISX(94, APR_STATUS_IS_EINPROGRESS, TCN_EINPROGRESS);
-        APR_ISX(95, APR_STATUS_IS_EINTR, TCN_EINTR);
-        APR_IS(96, APR_STATUS_IS_ENOTSOCK);
-        APR_IS(97, APR_STATUS_IS_EINVAL);
-    }
-    return JNI_FALSE;
-}
+}
```
nmittler: nit: carriage return?

nmittler:
@normanmaurer LGTM.  Is the testing method just to run through the netty tests? I guess that's sufficient ... it either works or it doesn't.

normanmaurer:
yep

Scottmitch:
my browser doesn't like this PR :cry:

Scottmitch:
@normanmaurer - LGTM!

>  And as a follow up I would like to rename packages and rename the artifact so it not clashes anymore with the "original" tcnative.

+1. At this point we should do a major version bump ... actually do we still want to call this project `netty-tcnative` at that point? What about something like `netty-jni-openssl`?

nmittler:
@Scottmitch agreed that renaming the artifact would avoid confusion.  Would that involve a new github project as well?

normanmaurer:
@nmittler maybe... 

@Scottmitch @trustin @nmittler let me create a new branch called "next" and merge all the "removal" stuff into it (also will cherry-pick any other changes).  Once Im done we can discuss where to go from there.

openssl-dynamic/src/main/c/error.c
normanmaurer: yep

normanmaurer:
Merged into next branch as 7303f307239edc73cc4e8b99730c38e8e4f50ed6

PR #158, opened 2016-06-13
Verify the checksums of apr,openssl, and libressl tarballs

cconroy:
Motivation:

These sources are downloaded over plaintext connections. We must verify
the known checksum to ensure the file has not been tampered with.

Modifications:

Utilize the antrun checksum task to verify the expected checksum
external dependencies.

Added verification instructions for libressl releases.

Result:

The build will fail in the event of a mismatched checksum.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -50,9 +50,23 @@
     <nativeLibOnlyDir>${project.build.directory}/native-lib-only</nativeLibOnlyDir>
     <nativeJarWorkdir>${project.build.directory}/native-jar-work</nativeJarWorkdir>
     <aprVersion>1.5.2</aprVersion>
+    <aprMd5>98492e965963f852ab29f9e61b2ad700</aprMd5>
     <boringsslBranch>chromium-stable</boringsslBranch>
     <libresslVersion>2.3.4</libresslVersion>
+    <!--
+        NB: libressl does not currently publish sha256 signatures and instead relies on signify
+        to verify releases. The project does not have a securely published GPG key.
+
+        Verifying a libressl release:
+        - Download the new version and the SHA256.sig from http://ftp.openbsd.org/pub/OpenBSD/LibreSSL/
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -50,9 +50,23 @@
...
     <libresslVersion>2.3.4</libresslVersion>
+    <!--
+        NB: libressl does not currently publish sha256 signatures and instead relies on signify
+        to verify releases. The project does not have a securely published GPG key.
+
+        Verifying a libressl release:
+        - Download the new version and the SHA256.sig from http://ftp.openbsd.org/pub/OpenBSD/LibreSSL/
+        - Grab the signify public key from https://www.openbsd.org/libressl/signing.html if you
+          don't already have it
+        - Verify the release: signify -V -x SHA256.sig  -p libressl.pub -m libressl-{libresslVersion}.tar.gz -e
```
cconroy: actually, i need to read up more on signify: this appears not actually validate the signature...sigh

pom.xml
```diff
@@ -50,9 +50,23 @@
     <nativeLibOnlyDir>${project.build.directory}/native-lib-only</nativeLibOnlyDir>
     <nativeJarWorkdir>${project.build.directory}/native-jar-work</nativeJarWorkdir>
     <aprVersion>1.5.2</aprVersion>
+    <aprMd5>98492e965963f852ab29f9e61b2ad700</aprMd5>
```
cconroy: apache publishes md5 checksums. It would be a bit more secure but more process to instead have a maintainer verify the md5 of the archive but compute a sha256 to use here. wdyt?

pom.xml
cconroy: fixed

pom.xml
normanmaurer: I think re-using whatever apache publishes is fine for now.

normanmaurer:
Cherry-picked as 2f16005d332e5bfa8c3824f4f114883c94150adc.

@cconroy thanks!

PR #160, opened 2016-06-16
Remove Pool class and implementation.

normanmaurer:
Motivation:

We only use the Pool to create a SSLContext later on. We should just create the pool under the hood and so not expose the Pool class at all.

Modifications:

- Remove Pool class and pool.c
- Create the apr pool when calling SSLContext.make(...) and destroy it when calling SSLContext.free(..)

Result:

Easier to use API for our use-case


normanmaurer:
@nmittler @Scottmitch @trustin FYI

Scottmitch:
lgtm

normanmaurer:
cherry-picked into next as 9358e74

PR #162, opened 2016-06-16
Allow to specify the length of the DH parameters to use.

normanmaurer:
Motivation:

Java8 allows to set a custom length for the DH paramaters. We should provide a way to do the same when using OpenSSL.

Modifications:

Add new method which allows to set the length.

Result:

Its now possible to set the length.


normanmaurer:
@Scottmitch @nmittler PTAL

Scottmitch:
lgtm

nmittler:
ship it!

normanmaurer:
Cherry-picked into 1.1.33 (03d0e88) and next (99a7e51e7963f6a64acbcd295c294ddbf03e3be7)

PR #163, opened 2016-06-16
Remove HAVE_OPENSSL checks as we only compile if openssl is present

normanmaurer:
Motivation:

We not need the HAVE_OPENSSL checks and so the empty stubs as we only compile when openssl is present anyway.

Modifications:

Remove HAVE_OPENSSL ifdefs

Result:

Less code to maintain (as we not need to create stubs).


normanmaurer:
@nmittler @Scottmitch PTAL

Scottmitch:
:+1: for less code!

normanmaurer:
Cherry-picked into next as 071bb6c36247bebe419f36833dc24d8bebadc99f

PR #164, opened 2016-06-22
Add support for setting callback via SSL_CTX_set_client_cert_cb

normanmaurer:
Motivation:

To be able to hook in X509KeyManager we need to allow setting a callback via SSL_CTX_set_client_cert_cb.

Modification:

Add CertificateRequestedCallback and needed SSL methods to set the cert / privatekey so we can hook in X509KeyManager from within netty.

Result:

Be able to hook in X509KeyManager

### DIFF
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -50,7 +50,6 @@ struct CRYPTO_dynlock_value {
     apr_thread_mutex_t *mutex;
 };
 
-
 /*
  * Handle the Temporary RSA Keys and DH Params
  */
@@ -835,11 +834,11 @@ TCN_IMPLEMENT_CALL(jint, SSL, fipsModeSet)(TCN_STDARGS, jint mode)
     if(1 != (r = (jint)FIPS_mode_set((int)mode))) {
       /* arrange to get a human-readable error message */
       unsigned long err = ERR_get_error();
-      char msg[256];
+      char msg[ERR_LEN];
 
       /* ERR_load_crypto_strings() already called in initialize() */
 
-      ERR_error_string_n(err, msg, 256);
+      ERR_error_string_n(err, msg, ERR_LEN);
 
       tcn_ThrowException(e, msg);
     }
@@ -1205,7 +1204,7 @@ TCN_IMPLEMENT_CALL(jboolean, SSL, loadDSATempKey)(TCN_STDARGS, jint idx,
 
 TCN_IMPLEMENT_CALL(jstring, SSL, getLastError)(TCN_STDARGS)
 {
-    char buf[256];
+    char buf[ERR_LEN];
     UNREFERENCED(o);
     ERR_error_string(ERR_get_error(), buf);
     return tcn_new_string(e, buf);
@@ -1686,7 +1685,9 @@ TCN_IMPLEMENT_CALL(jobjectArray, SSL, getPeerCertChain)(TCN_STDARGS,
         buf = NULL;
         length = i2d_X509(cert, &buf);
         if (length < 0) {
-            OPENSSL_free(buf);
+            if (buf != NULL) {
+                OPENSSL_free(buf);
+            }
             // In case of error just return an empty byte[][]
             return (*e)->NewObjectArray(e, 0, byteArrayClass, NULL);
         }
@@ -1743,7 +1744,7 @@ TCN_IMPLEMENT_CALL(jbyteArray, SSL, getPeerCertificate)(TCN_STDARGS,
 
 TCN_IMPLEMENT_CALL(jstring, SSL, getErrorString)(TCN_STDARGS, jlong number)
 {
-    char buf[256];
+    char buf[ERR_LEN];
     UNREFERENCED(o);
     ERR_error_string(number, buf);
     return tcn_new_string(e, buf);
@@ -1971,7 +1972,7 @@ TCN_IMPLEMENT_CALL(jboolean, SSL, setCipherSuites)(TCN_STDARGS, jlong ssl,
     }
 
     if (!SSL_set_cipher_list(ssl_, J2S(ciphers))) {
-        char err[256];
+        char err[ERR_LEN];
         ERR_error_string(ERR_get_error(), err);
         tcn_Throw(e, "Unable to configure permitted SSL ciphers (%s)", err);
         rv = JNI_FALSE;
@@ -2077,13 +2078,123 @@ TCN_IMPLEMENT_CALL(void, SSL, setTlsExtHostName)(TCN_STDARGS, jlong ssl, jstring
         UNREFERENCED(o);
 
         if (SSL_set_tlsext_host_name(ssl_, J2S(hostname)) != 1) {
-            char err[256];
+            char err[ERR_LEN];
             ERR_error_string(ERR_get_error(), err);
             tcn_Throw(e, "Unable to set TLS servername extension (%s)", err);
         }
     }
 
     TCN_FREE_CSTRING(hostname);
+}
+
+TCN_IMPLEMENT_CALL(jobjectArray, SSL, authenticationMethods)(TCN_STDARGS, jlong ssl) {
+    SSL *ssl_ = J2P(ssl, SSL *);
+    const STACK_OF(SSL_CIPHER) *ciphers = NULL;
+    int len;
+    int i;
+    jobjectArray array;
+
+    TCN_ASSERT(ssl_ != NULL);
+
+    UNREFERENCED(o);
+
+    ciphers = SSL_get_ciphers(ssl_);
+    len = sk_num((_STACK*) ciphers);
+
+    array = (*e)->NewObjectArray(e, len, stringClass, NULL);
+
+    for (i = 0; i < len; i++) {
+        (*e)->SetObjectArrayElement(e, array, i,
+        (*e)->NewStringUTF(e, SSL_cipher_authentication_method((SSL_CIPHER*) sk_value((_STACK*) ciphers, i))));
+    }
+    return array;
+}
+
+TCN_IMPLEMENT_CALL(void, SSL, setCertificateBio)(TCN_STDARGS, jlong ssl,
+                                                         jlong cert, jlong key,
+                                                         jstring password)
+{
+    SSL *ssl_ = J2P(ssl, SSL *);
+    BIO *cert_bio = J2P(cert, BIO *);
+    BIO *key_bio = J2P(key, BIO *);
+    EVP_PKEY* pkey;
+    X509* xcert;
+    tcn_pass_cb_t* cb_data;
+    TCN_ALLOC_CSTRING(password);
+    char err[ERR_LEN];
+
+    UNREFERENCED(o);
+    TCN_ASSERT(ssl != NULL);
+
+    cb_data = &tcn_password_callback;
+    cb_data->password[0] = '\0';
+    if (J2S(password) != NULL) {
+        strncat(cb_data->password, J2S(password), SSL_MAX_PASSWORD_LEN - 1);
+    }
+
+    if (key <= 0) {
+        key = cert;
+    }
+
+    if (cert <= 0 || key <= 0) {
+        tcn_Throw(e, "No Certificate file specified or invalid file format");
+        goto cleanup;
+    }
+
+    if ((pkey = load_pem_key_bio(cb_data, key_bio)) == NULL) {
+        ERR_error_string_n(ERR_get_error(), err, ERR_LEN);
+        ERR_clear_error();
+        tcn_Throw(e, "Unable to load certificate key (%s)",err);
+        goto cleanup;
+    }
+    if ((xcert = load_pem_cert_bio(cb_data, cert_bio)) == NULL) {
+        ERR_error_string_n(ERR_get_error(), err, ERR_LEN);
+        ERR_clear_error();
+        tcn_Throw(e, "Unable to load certificate (%s) ", err);
+        goto cleanup;
+    }
+
+    if (SSL_use_certificate(ssl_, xcert) <= 0) {
+        ERR_error_string_n(ERR_get_error(), err, ERR_LEN);
+        ERR_clear_error();
+        tcn_Throw(e, "Error setting certificate (%s)", err);
+        goto cleanup;
+    }
+    if (SSL_use_PrivateKey(ssl_, pkey) <= 0) {
+        ERR_error_string_n(ERR_get_error(), err, ERR_LEN);
+        ERR_clear_error();
+        tcn_Throw(e, "Error setting private key (%s)", err);
+        goto cleanup;
+    }
+    if (SSL_check_private_key(ssl_) <= 0) {
+        ERR_error_string_n(ERR_get_error(), err, ERR_LEN);
+        ERR_clear_error();
+
+        tcn_Throw(e, "Private key does not match the certificate public key (%s)",
+                  err);
+        goto cleanup;
+    }
+cleanup:
+    TCN_FREE_CSTRING(password);
+}
+
+TCN_IMPLEMENT_CALL(void, SSL, setCertificateChainBio)(TCN_STDARGS, jlong ssl,
+                                                                  jlong chain,
+                                                                  jboolean skipfirst)
+{
+    SSL *ssl_ = J2P(ssl, SSL *);
+    BIO *b = J2P(chain, BIO *);
+    char err[ERR_LEN];
+
+    UNREFERENCED(o);
+    TCN_ASSERT(ssl_ != NULL);
+    TCN_ASSERT(b != NULL);
+
+    if (SSL_use_certificate_chain_bio(ssl_, b, skipfirst) < 0)  {
+        ERR_error_string_n(ERR_get_error(), err, ERR_LEN);
+        ERR_clear_error();
+        tcn_Throw(e, "Error setting certificate chain (%s)", err);
+    }
 }
 
 /*** End Apple API Additions ***/
@@ -2517,5 +2628,10 @@ TCN_IMPLEMENT_CALL(void, SSL, setTlsExtHostName)(TCN_STDARGS, jlong ssl, jstring
   tcn_ThrowException(e, "Not implemented");
 }
 
+TCN_IMPLEMENT_CALL(jobjectArray, SSL, authenticationMethods)(TCN_STDARGS, jlong ssl) {
+  UNREFERENCED(o);
+  UNREFERENCED(ssl);
+  tcn_ThrowException(e, "Not implemented");
+}
 /*** End Apple API Additions ***/
 #endif
--- a/openssl-dynamic/src/main/c/sslcontext.c
+++ b/openssl-dynamic/src/main/c/sslcontext.c
@@ -31,8 +31,6 @@
 #ifdef HAVE_OPENSSL
 #include "ssl_private.h"
 
-static const char* UNKNOWN_AUTH_METHOD = "UNKNOWN";
-
 static jclass byteArrayClass;
 
 static apr_status_t ssl_context_cleanup(void *data)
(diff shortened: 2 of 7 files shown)


normanmaurer:
@Scottmitch @nmittler PTAL... I need this for a follow up pr to add support for KeyManagerFactor in netty when using openssl

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -2086,6 +2086,117 @@ TCN_IMPLEMENT_CALL(void, SSL, setTlsExtHostName)(TCN_STDARGS, jlong ssl, jstring
...
+    tcn_pass_cb_t* cb_data;
+    TCN_ALLOC_CSTRING(password);
+    char err[256];
+
+    UNREFERENCED(o);
+    TCN_ASSERT(ssl != NULL);
+
+    if (J2S(password)) {
+        cb_data = &tcn_password_callback;
+        strncpy(cb_data->password, J2S(password), SSL_MAX_PASSWORD_LEN);
```
Scottmitch: use [strncat](http://www.cplusplus.com/reference/cstring/strncat/) instead of [strncpy](http://www.cplusplus.com/reference/cstring/strncpy/). `strncat` will always null terminate and doesn't "pad with zeros" unnecessarily.

```c
cb_data->password[0] = '\0';
strncat(cb_data->password, J2S(password), SSL_MAX_PASSWORD_LEN - 1);
```

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -2086,6 +2086,117 @@ TCN_IMPLEMENT_CALL(void, SSL, setTlsExtHostName)(TCN_STDARGS, jlong ssl, jstring
...
+    SSL *ssl_ = J2P(ssl, SSL *);
+    const STACK_OF(SSL_CIPHER) *ciphers = NULL;
+    int len;
+    int i;
+    jobjectArray array;
+
+    if (ssl_ == NULL) {
+        tcn_ThrowException(e, "ssl is null");
+        return 0;
+    } else {
```
Scottmitch: nit: remove this `else` (you return above)

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -2086,6 +2086,117 @@ TCN_IMPLEMENT_CALL(void, SSL, setTlsExtHostName)(TCN_STDARGS, jlong ssl, jstring
...
+    }
+    if (key <= 0)
+        key = cert;
+    if (cert <= 0 || key <= 0) {
+        tcn_Throw(e, "No Certificate file specified or invalid file format");
+        goto cleanup;
+    }
+
+    if ((pkey = load_pem_key_bio(cb_data, key_bio)) == NULL) {
+        ERR_error_string_n(ERR_get_error(), err, 256);
```
Scottmitch: consider having a `const` or `#define` for the "commonly used error string length"

openssl-dynamic/src/main/c/ssl.c
normanmaurer: +1

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -2086,6 +2086,117 @@ TCN_IMPLEMENT_CALL(void, SSL, setTlsExtHostName)(TCN_STDARGS, jlong ssl, jstring
...
+    TCN_ASSERT(b != NULL);
+
+    if (SSL_use_certificate_chain_bio(ssl_, b, skipfirst) < 0)  {
+        ERR_error_string_n(ERR_get_error(), err, 256);
+        ERR_clear_error();
+        tcn_Throw(e, "Error setting certificate chain (%s)", err);
+    }
+}
+
+
```
Scottmitch: kill line

openssl-dynamic/src/main/c/ssl.c
normanmaurer: +1

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -1562,7 +1442,7 @@ static const char* authentication_method(const SSL* ssl) {
                 // No cipher available so return UNKNOWN.
                 return UNKNOWN_AUTH_METHOD;
             }
-            return cipher_authentication_method(sk_value((_STACK*) ciphers, 0));
+            return SSL_cipher_authentication_method(sk_value((_STACK*) ciphers, 0));
```
Scottmitch: i would advise against using the `SSL_*` for method name prefix. This namespace is implicitly owned by OpenSSL and  is misleading as to where the method comes from and using the same prefix as another library means the probability of method name conflict goes up.

normanmaurer: @Scottmitch its consistent with other methods defined n ssl_private.h

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -1659,6 +1538,153 @@ TCN_IMPLEMENT_CALL(void, SSLContext, setCertVerifyCallback)(TCN_STDARGS, jlong c
...
+        return NULL;
+    }
+
+    for (int i = 0; i < count; i++) {
+        principal = sk_X509_NAME_value(names, i);
+        buf = NULL;
+        length = i2d_X509_NAME(principal, &buf);
+        if (length < 0) {
+            // We need to delete the local references so we not leak memory as this method is called via callback.
+            OPENSSL_free(buf);
```
Scottmitch: through some code inspection it looks like it is possible for `free` to be used here. Calling `free(NULL)` is undefined and may lead to bad things ... we should check that `buf != NULL` before calling this.

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -1659,6 +1538,153 @@ TCN_IMPLEMENT_CALL(void, SSLContext, setCertVerifyCallback)(TCN_STDARGS, jlong c
...
+    tcn_get_java_env(&e);
+
+    /* Clear output of key and certificate in case of early exit due to error. */
+    *x509Out = NULL;
+    *pkeyOut = NULL;
+
+#if !defined(OPENSSL_IS_BORINGSSL) && OPENSSL_VERSION_NUMBER < 0x1000200fL
+    switch (ssl->version) {
+        case SSL2_VERSION:
+            ctype_bytes = (jbyte*) &ssl2_ctype;
```
Scottmitch: `&ssl2_ctype` -> `&SSL3_CT_RSA_SIGN` ?

normanmaurer: this not compiles ... tried it ;)

openssl-dynamic/src/main/c/sslcontext.c
normanmaurer: good catch...

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -1659,6 +1538,153 @@ TCN_IMPLEMENT_CALL(void, SSLContext, setCertVerifyCallback)(TCN_STDARGS, jlong c
...
+        case TLS1_VERSION:
+        case TLS1_1_VERSION:
+        case TLS1_2_VERSION:
+        case DTLS1_VERSION:
+            ctype_bytes = (jbyte*) ssl->s3->tmp.ctype;
+            ctype_num = ssl->s3->tmp.ctype_num;
+            break;
+    }
+#else
+    const uint8_t* ctype = NULL;
```
Scottmitch: openssl will cast this to a `char*` under the hood anyways ... can we just pass in `ctype_bytes` directly?

openssl-dynamic/src/main/c/sslutils.c
```diff
@@ -482,11 +568,82 @@ int SSL_CTX_use_certificate_chain_bio(SSL_CTX *ctx, BIO *bio,
...
+    EVP_PKEY *key = NULL;
+    int i;
+
+    if (cb_data == NULL)
+        cb_data = &tcn_password_callback;
+    for (i = 0; i < 3; i++) {
+        key = PEM_read_bio_PrivateKey((BIO*) bio, NULL,
+                    (pem_password_cb *)SSL_password_callback,
+                    (void *)cb_data);
+        if (key)
```
Scottmitch: `key != NULL`

openssl-dynamic/src/main/java/org/apache/tomcat/jni/SSL.java
```diff
@@ -761,4 +761,52 @@ public static native boolean setCipherSuites(long ssl, String ciphers)
...
+     * <br>
+     * If the key is not combined with the certificate, use key param
+     * to point at the key file.  Keep in mind that if
+     * you've both a RSA and a DSA private key you can configure
+     * both in parallel (to also allow the use of DSA ciphers, etc.)
+     * @param ssl Server or Client to use.
+     * @param certBio Certificate BIO.
+     * @param keyBio Private Key BIO to use if not in cert.
+     * @param password Certificate password. If null and certificate
+     *                 is encrypted, password prompt will be displayed.
```
Scottmitch: `password prompt will be displayed` -> reminder to clean this up :)

openssl-dynamic/src/main/c/sslcontext.c
Scottmitch: can we clean this up as a followup PR, or are you concerned about API breaking changes?

normanmaurer: yep let us cleanup as followup.

openssl-dynamic/src/main/c/sslcontext.c
normanmaurer: nope this fails on boringssl

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -2086,6 +2090,115 @@ TCN_IMPLEMENT_CALL(void, SSL, setTlsExtHostName)(TCN_STDARGS, jlong ssl, jstring
...
+    EVP_PKEY* pkey;
+    X509* xcert;
+    tcn_pass_cb_t* cb_data;
+    TCN_ALLOC_CSTRING(password);
+    char err[ERR_LEN];
+
+    UNREFERENCED(o);
+    TCN_ASSERT(ssl != NULL);
+
+    if (J2S(password)) {
```
nmittler: != NULL to be consistent with checks elsewhere?

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -1659,6 +1540,155 @@ TCN_IMPLEMENT_CALL(void, SSLContext, setCertVerifyCallback)(TCN_STDARGS, jlong c
...
+    }
+
+    for (i = 0; i < count; i++) {
+        principal = sk_X509_NAME_value(names, i);
+        buf = NULL;
+        length = i2d_X509_NAME(principal, &buf);
+        if (length < 0) {
+            if (buf != NULL) {
+                // We need to delete the local references so we not leak memory as this method is called via callback.
+                OPENSSL_free(buf);
```
nmittler: does OPENSSL_free not do the null check?

nmittler:
@normanmaurer LGTM

normanmaurer:
@Scottmitch @nmittler I noticed this only works when using openssl 1.0.2+ / libressl 2.4+ or boringssl. So I will need to add some more ifdef magic... stay tuned

openssl-dynamic/src/main/c/sslcontext.c
Scottmitch: @nmittler - I did some code inspection (referenced in https://github.com/netty/netty-tcnative/pull/164#discussion_r68122457) and IIUC this can be set to `free`. `free` does not have the same `NULL` protections offered by `delete`.

openssl-dynamic/src/main/c/ssl.c
Scottmitch: also if the password is null ... should we set `cb_data = NULL`?

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -51,6 +51,8 @@ struct CRYPTO_dynlock_value {
 };
 
 
+#define ERR_LEN 256
```
Scottmitch: lets make this more generally available (in a shared header file) and use it everywhere ... unless we have good reason not to. `s/ERR_error_string/ERR_error_string_n`

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -2086,6 +2090,115 @@ TCN_IMPLEMENT_CALL(void, SSL, setTlsExtHostName)(TCN_STDARGS, jlong ssl, jstring
...
+
+    UNREFERENCED(o);
+    TCN_ASSERT(ssl != NULL);
+
+    if (J2S(password)) {
+        cb_data = &tcn_password_callback;
+        cb_data->password[0] = '\0';
+        strncat(cb_data->password, J2S(password), SSL_MAX_PASSWORD_LEN - 1);
+    }
+    if (key <= 0)
```
Scottmitch: nit: add newline?

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -2517,5 +2630,11 @@ TCN_IMPLEMENT_CALL(void, SSL, setTlsExtHostName)(TCN_STDARGS, jlong ssl, jstring
   tcn_ThrowException(e, "Not implemented");
 }
 
+
```
Scottmitch: kill line

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -2086,6 +2090,115 @@ TCN_IMPLEMENT_CALL(void, SSL, setTlsExtHostName)(TCN_STDARGS, jlong ssl, jstring
...
 }
 
+TCN_IMPLEMENT_CALL(jobjectArray, SSL, authenticationMethods)(TCN_STDARGS, jlong ssl) {
+    SSL *ssl_ = J2P(ssl, SSL *);
+    const STACK_OF(SSL_CIPHER) *ciphers = NULL;
+    int len;
+    int i;
+    jobjectArray array;
+
+    if (ssl_ == NULL) {
```
Scottmitch: we use `TCN_ASSERT(ssl_ != NULL)` in other methods ... lets be consistent and choose one

openssl-dynamic/src/main/c/sslcontext.c
Scottmitch: @normanmaurer - then can you just move the definition of `char ssl2_ctype = SSL3_CT_RSA_SIGN;` into this block? its only needed here.

Scottmitch: assuming this code survives the `#ifdef` chopping block

openssl-dynamic/src/main/c/sslcontext.c
Scottmitch: cast until its happy :) ... we are casting after the fact anyways

openssl-dynamic/src/main/c/sslcontext.c
nmittler: Do you think it would make sense to make our own wrapper method that does the null check so we don't have to check null everywhere?

openssl-dynamic/src/main/c/ssl.c
normanmaurer: I didnt want to make to many changes which are not related to this feature... Let me to a followup

openssl-dynamic/src/main/c/sslcontext.c
Scottmitch: we could. any suggestions for names (`TCN_OPENSSL_free`)?

in general if you know the variable is not `null` you wouldn't have to do an additional check ... but here it appears like it could be null.

nmittler: if this won't have a very big impact, then maybe we just leave it as-is for now?

@normanmaurer WDYT?

normanmaurer: Yep... Will do some more cleanup rounds anyway

openssl-dynamic/src/main/c/ssl.c
normanmaurer: +1 ...

openssl-dynamic/src/main/c/sslcontext.c
normanmaurer: this not compiles on centos as its not c99 compliant...

openssl-dynamic/src/main/c/sslcontext.c
normanmaurer: cast all the things!

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -2086,6 +2087,117 @@ TCN_IMPLEMENT_CALL(void, SSL, setTlsExtHostName)(TCN_STDARGS, jlong ssl, jstring
...
+    if (SSL_check_private_key(ssl_) <= 0) {
+        ERR_error_string_n(ERR_get_error(), err, ERR_LEN);
+        ERR_clear_error();
+
+        tcn_Throw(e, "Private key does not match the certificate public key (%s)",
+                  err);
+        goto cleanup;
+    }
+cleanup:
+    TCN_FREE_CSTRING(password);
```
Scottmitch: is it safe to call this (`ReleaseStringUTFChars`) if password is `null`?

normanmaurer: not sure... better not risk it.

openssl-dynamic/src/main/c/sslcontext.c
Scottmitch: ok sgtm .. we had `uint8_t` defined below previously but now that is gone so fine as is

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -2086,6 +2087,117 @@ TCN_IMPLEMENT_CALL(void, SSL, setTlsExtHostName)(TCN_STDARGS, jlong ssl, jstring
...
+
+    UNREFERENCED(o);
+    TCN_ASSERT(ssl != NULL);
+
+    if (J2S(password) != NULL) {
+        cb_data = &tcn_password_callback;
+        cb_data->password[0] = '\0';
+        strncat(cb_data->password, J2S(password), SSL_MAX_PASSWORD_LEN - 1);
+    } else {
+        cb_data = NULL;;
```
Scottmitch: remove extra `;`

openssl-dynamic/src/main/c/sslutils.c
```diff
@@ -482,11 +568,89 @@ int SSL_CTX_use_certificate_chain_bio(SSL_CTX *ctx, BIO *bio,
...
+    return cert;
+}
+
+EVP_PKEY *load_pem_key_bio(tcn_pass_cb_t *cb_data, const BIO *bio)
+{
+    EVP_PKEY *key = NULL;
+    int i;
+
+    if (cb_data == NULL)
+        cb_data = &tcn_password_callback;
```
Scottmitch: `sslcontext::setCertificateBio` may call this with `cb_data == null` ... however the password structure is not touched ... should we do a `cb_data->password[0] = '\0';` under this condition (somewhere)?

Scottmitch: same issue for `load_pem_cert_bio` ... i wonder if we should be setting `c->cb_data = &tcn_password_callback;` and nulling put a `null` in the first element event if `password` is  `null` in `sslcontext::setCertificateBio`?

normanmaurer: @Scottmitch I think I not really get what you mean ... Can you explain ? Also please note that this code was there before, I just moved it.

Scottmitch:
@normanmaurer - few comments and lgtm

openssl-dynamic/src/main/c/ssl.c
normanmaurer: its fine... checked the macro and it handles it

openssl-dynamic/src/main/c/sslutils.c
normanmaurer: we talked offline and we will do some more cleanup later on in this area and others. so fine for now

normanmaurer:
Cherry-picked into 1.1.33 (6592bd675e796da6f97c6a9f6a73ab57555eb61f) and next (e73f36a166997cf7aa55701900fc46e438cf6ee8)

openssl-dynamic/src/main/c/ssl.c
Scottmitch: +1. I missed this if check in the macro on my first pass.

PR #169, opened 2016-07-08
Add script to support auto-generation of Bundle-NativeCode manifest

xsir:
Motivation:

The build output a OSGi bundle with missing Bundle-NativeCode header, will cause the Library.class initialize with null error. Add the Ant script to generate the header automatically.

Modifications:

1. Using the ant plugin to generate the Bundle-NativeCode header according to the compiler's OS;
2. Generate a static header for the "uber" package.

Result:

The Bundle-NativeCode manifest header will be added to the output jar.

Signed-off-by: XU JINCHUAN <xsir@msn.com>

Sorry for my mistake to submit a new pull request. And I have signed the ICLA.
Thank you!

### DIFF
--- a/boringssl-static/pom.xml
+++ b/boringssl-static/pom.xml
@@ -373,6 +373,9 @@
                   </supportedProjectTypes>
                   <instructions>
                     <Export-Package>org.apache.tomcat.jni.*</Export-Package>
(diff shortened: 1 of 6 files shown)


normanmaurer:
Will take a look

normanmaurer:
Cherry-picked into 1.1.33 (50f15899cd4815b6ea181dcb88d0301383e416f6) and next (146e1aa)

xsir:
Associated with #135 , may be is should be closed:)

PR #170, opened 2016-07-13
Fixes the OS alias miss spelling in Bundle-NativeCode header.

xsir:
Motivation:

The commit #169 contains a bug: the `osname` value in Bundle-NativeCode header does NOT follow the OSGi specification. For example, the Windows family should use the alias `win32` or the name such as `Windows7`, `Windows8`, etc. Otherwise, the netty-tcnative bundle will encounter the "unresolve native library" runtime problem.
This patch trying to fix the OS name and alias miss match problem according to the OSGi specification.

Modifications:

Using the OS alias as `osname`, instead of the detected OS name.

Result:

The Bundle-NativeCode header generated by ant script will work as design. For example, the bundle build  on Windows, the generated header would be like:
`Bundle-NativeCode: META-INF/native/netty-tcnative.dll;osname=win32;processor=x86_64`

Reference:
[The OSGi specification Reference](https://www.osgi.org/developer/specifications/reference/)

Signed-off-by: XU JINCHUAN <xsir@msn.com>

### DIFF
--- a/boringssl-static/pom.xml
+++ b/boringssl-static/pom.xml
@@ -373,9 +373,11 @@
                   </supportedProjectTypes>
                   <instructions>
                     <Export-Package>org.apache.tomcat.jni.*</Export-Package>
-                    <Bundle-NativeCode>META-INF/native/libnetty-tcnative-osx-x86_64.jnilib;osname=osx;processor=x86_64,
-                                       META-INF/native/libnetty-tcnative-linux-x86_64.so;osname=linux;processor=x86_64,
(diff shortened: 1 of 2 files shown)


Scottmitch:
thanks @xsir 

@normanmaurer - PTAL

normanmaurer:
Cherry-picked into 1.1.33 (ea27386540f857032873292183f1c35e0c0cdbc6) and next (ff900df1f4e99efb036a458e5e99cc3c438e8476)

PR #171, opened 2016-07-14
Fix unused variable warnings

carl-mastrangelo:
Motivation:
Some compilers complain if variables are not used

Modifications:
Remove an unused variable, and explicitly ignore another in case it isn't used.

Result:
Easier compilation time.

cc: @Scottmitch

### DIFF
--- a/openssl-dynamic/src/main/c/sslcontext.c
+++ b/openssl-dynamic/src/main/c/sslcontext.c
@@ -1625,7 +1625,6 @@ static jobjectArray principalBytes(JNIEnv* e, const STACK_OF(X509_NAME)* names) 
  */
 static int cert_requested(SSL* ssl, X509** x509Out, EVP_PKEY** pkeyOut) {
     tcn_ssl_ctxt_t *c = SSL_get_app_data2(ssl);
-    char ssl2_ctype = SSL3_CT_RSA_SIGN;
     int ctype_num;
     jbyte* ctype_bytes;
     jobjectArray issuers;
@@ -1640,6 +1639,7 @@ static int cert_requested(SSL* ssl, X509** x509Out, EVP_PKEY** pkeyOut) {
     *pkeyOut = NULL;
 
 #if !defined(OPENSSL_IS_BORINGSSL) && (OPENSSL_VERSION_NUMBER < 0x1000200fL || LIBRESSL_VERSION_NUMBER < 0x20400000L)
+    char ssl2_ctype = SSL3_CT_RSA_SIGN;
     switch (ssl->version) {
         case SSL2_VERSION:
             ctype_bytes = (jbyte*) &ssl2_ctype;
@@ -1655,7 +1655,6 @@ static int cert_requested(SSL* ssl, X509** x509Out, EVP_PKEY** pkeyOut) {
             break;
     }
 #else
-    (void) ssl2_ctype;
     ctype_num = SSL_get0_certificate_types(ssl, (const uint8_t **) &ctype_bytes);
 #endif
     if (ctype_num <= 0) {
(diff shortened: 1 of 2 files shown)


normanmaurer:
LGTM... @Scottmitch ok ?

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -1655,6 +1655,7 @@ static int cert_requested(SSL* ssl, X509** x509Out, EVP_PKEY** pkeyOut) {
             break;
     }
 #else
+    (void) ssl2_ctype;
```
Scottmitch: can we instead just declare this variable in the same `ifdef` condition in which it is used in?

carl-mastrangelo: I assumed that all declarations had to be at the beginning of the block (c89 I think?)  If that isn't a requirement I can move it down.

Scottmitch: I think we should be able to declare variables at the beginning of any new scope.

Option 1. Wrap existing declaration in `#if`

```c
tcn_ssl_ctxt_t *c = SSL_get_app_data2(ssl);
#if !defined(OPENSSL_IS_BORINGSSL) && (OPENSSL_VERSION_NUMBER < 0x1000200fL || LIBRESSL_VERSION_NUMBER < 0x20400000L)
char ssl2_ctype = SSL3_CT_RSA_SIGN;
#endif
int ctype_num;
...
```

Option 2. Declare a new scope in the existing `#if` (I haven't tried this so please confirm)

```c
#if !defined(OPENSSL_IS_BORINGSSL) && (OPENSSL_VERSION_NUMBER < 0x1000200fL || LIBRESSL_VERSION_NUMBER < 0x20400000L)
{
     char ssl2_ctype = SSL3_CT_RSA_SIGN;
     switch (ssl->version) {
      ...
}
#else
```

carl-mastrangelo: Option 2 won't work because we take the address of it, and it's allocated on the stack.

Option 1 works, but my version is overall less code.

Scottmitch: > Option 2 won't work because we take the address of it, and it's allocated on the stack.

Good point ... its used else where i see.

> Option 1 works, but my version is overall less code.

After compile or before compile :wink:. I'm ok with the current approach ... however if the issue we are trying to resolve is "get rid of a warning because we don't use the variable" ... instead of artificially using it to appease the warning ... why not just "fix" the warning and get rid of the variable if its not needed (ifdef it out)?

carl-mastrangelo: That's fair.  Done.

Scottmitch:
lgtm assuming the variable is OK to define after the `*x509Out = NULL; ...` statements. not sure about the build failure though...

```
[18:38:26][io.netty:netty-tcnative-boringssl-static] [INFO] Working directory: /var/lib/teamcity-agent/work/dd2ea5c9a46891e0/boringssl-static/target/boringssl-chromium-stable
[18:38:35][io.netty:netty-tcnative-boringssl-static] [ERROR] Provider message:
[18:38:35][io.netty:netty-tcnative-boringssl-static] [ERROR] The git-pull command failed.
[18:38:35][io.netty:netty-tcnative-boringssl-static] [ERROR] Command output:
[18:38:35][io.netty:netty-tcnative-boringssl-static] [ERROR] error: Couldn't resolve host 'boringssl.googlesource.com' while accessing https://boringssl.googlesource.com/boringssl/info/refs
[18:38:35][io.netty:netty-tcnative-boringssl-static] 
[18:38:35][io.netty:netty-tcnative-boringssl-static] fatal: HTTP request failed
```

normanmaurer:
@carl-mastrangelo thanks!

I squashed and cherry-picked into 1.1.33 (05e46fda03f1e8577c4e0a94a48b4239b4cb91ca) and next (766b47f)

Issue #172, opened 2016-07-14
Support Java 6


pjulien:
Never noticed before because I'm always on Java 8 but this project requires Java 7.  Now that I'm stuck on Java 6, please consider adding support for Java 6.

PR incoming.

normanmaurer:
Fixed by https://github.com/netty/netty-tcnative/pull/173

PR #173, opened 2016-07-14
[172] Java 6 Support

pjulien:
Motivation

Add support for Java 6 for #172.

Modifications

Other than the pom.xml, Java 7 and 8 constructs have been removed.

Result

Usable from Java 6 now.

### DIFF
--- a/openssl-dynamic/src/main/java/org/apache/tomcat/jni/socket/AprSocketContext.java
+++ b/openssl-dynamic/src/main/java/org/apache/tomcat/jni/socket/AprSocketContext.java
@@ -108,7 +108,7 @@ public class AprSocketContext {
     /**
      * Pollers.
      */
-    private final List<AprPoller> pollers = new ArrayList<>();
+    private final List<AprPoller> pollers = new ArrayList<AprPoller>();
 
     // Set on all accepted or connected sockets.
     // TODO: add the other properties
@@ -123,7 +123,7 @@ public class AprSocketContext {
     private boolean nonBlockingAccept = false;
 
     private final BlockingQueue<AprSocket> acceptedQueue =
-            new LinkedBlockingQueue<>();
+            new LinkedBlockingQueue<AprSocket>();
 
     /**
      * Root APR memory pool.
@@ -157,7 +157,7 @@ public class AprSocketContext {
     final RawDataHandler rawDataHandler = null;
 
     // TODO: do we need this here ?
-    private final Map<String, HostInfo> hosts = new HashMap<>();
+    private final Map<String, HostInfo> hosts = new HashMap<String, HostInfo>();
 
     private String certFile;
     private String keyFile;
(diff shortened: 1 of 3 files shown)


normanmaurer:
LGTM... @pjulien can you sign our icla:

http://netty.io/s/icla

pjulien:
@normanmaurer 

> LGTM... @pjulien can you sign our icla:

done

Scottmitch:
lgtm

normanmaurer:
@pjulien thanks a lot...

Cherry-picked into 1.1.33 (197327ffbdb2c61d575b8f390ab15a6636e709a6) and next (8fd6d83738a83d71913ee655822a4cc9f6a97b91)

openssl-dynamic/src/main/java/org/apache/tomcat/jni/socket/AprSocketContext.java
```diff
@@ -797,10 +797,16 @@ void prepare() throws IOException {
...
-                // Easiest ( maybe safest ) way to interrupt accept
-                // we could have it in non-blocking mode, etc
-                sock.connect(new InetSocketAddress("127.0.0.1", port));
+            java.net.Socket sock = new java.net.Socket();
+
+            try {
+                try {
+                    // Easiest ( maybe safest ) way to interrupt accept
+                    // we could have it in non-blocking mode, etc
+                    sock.connect(new InetSocketAddress("127.0.0.1", port));
```
netkins: MAJOR Make this IP "127.0.0.1" address configurable. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS1313)

openssl-dynamic/src/main/java/org/apache/tomcat/jni/socket/AprSocketContext.java
```diff
@@ -797,10 +797,16 @@ void prepare() throws IOException {
         }
 
         void unblock() {
-            try (java.net.Socket sock = new java.net.Socket()) {
-                // Easiest ( maybe safest ) way to interrupt accept
-                // we could have it in non-blocking mode, etc
-                sock.connect(new InetSocketAddress("127.0.0.1", port));
+            java.net.Socket sock = new java.net.Socket();
```
netkins: BLOCKER Close this "Socket". [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AS2095)

openssl-dynamic/src/main/java/org/apache/tomcat/jni/socket/AprSocketContext.java
```diff
@@ -797,10 +797,16 @@ void prepare() throws IOException {
...
+
+            try {
+                try {
```
Scottmitch: cleanup: we should only need a single `try` here. Lets move the finally to the outer `try` and remove the inner try.

Scottmitch: @pjulien - would you mind submitting a followup PR?

pjulien: @Scottmitch 

>  Lets move the finally to the outer try and remove the inner try.

The code is correct as-is.  Even if I were to do what you say here, you would still need another try/finally block because ``close`` throws ``IOException``.

Considering every exception is ignored here, I would argue that this style makes slightly more sense.  That being said, I don't know if there is a util class that supports closing and swallowing the exception.  If that's available, this would be much nicer.

Scottmitch: thanks for clarifying. I'm not aware of any such utility ... fine as is

PR #175, opened 2016-08-04
Remove casts by using correct macros

normanmaurer:
Motivation:

We can use the correct macros and so remove manual casts.

Modifications:

Use correct macros.

Result:

Cleaner code.


Scottmitch:
lgtm (assuming these interfaces are available in older ssl versions ... which looks like they are)

normanmaurer:
Cherry-picked into 1.1.33 (b670d74) and next (94d4c41)

PR #178, opened 2016-08-11
Fix memory leaks

Scottmitch:
Motivation:
Through a code inspection some potential memory leaks were discovered.

Modifications:
- Fix memory leaks

Result:
Less memory leaks


Scottmitch:
@normanmaurer - PTAL

normanmaurer:
@Scottmitch ship it.

Scottmitch:
1.1.33 (a418fb8)

PR #181, opened 2016-08-22
Refactor CertificateRequestedCallback to fix possible segfault.

normanmaurer:
Motivation:

In the previous CertificateRequestedCallback interface the contract was that users will directly call SSL_ methods to set the key, cert and chain. This could lead to memory leaks or segfaults depending on if the key, cert and chain is freed or not.

Modification:

Change the CertificateRequestedCallback interface to return a KeyMaterial object that holds the chain, key and password. We will call the correct SSL_ methods after calling the callback and so help to prevent leaks and segfaults.

Results:

No more leaks and segfaults.

### DIFF
--- a/openssl-dynamic/src/main/c/jnilib.c
+++ b/openssl-dynamic/src/main/c/jnilib.c
@@ -46,6 +46,10 @@ static jclass    jFinfo_class;
 static jclass    jAinfo_class;
 static jmethodID jString_init;
 static jmethodID jString_getBytes;
+static jclass    byteArrayClass;
+static jclass    keyMaterialClass;
+static jfieldID  keyMaterialCertificateChainFieldId;
+static jfieldID  keyMaterialPrivateKeyFieldId;
 
 int tcn_parent_pid = 0;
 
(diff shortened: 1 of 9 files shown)


normanmaurer:
@Scottmitch @nmittler PTAL

openssl-dynamic/src/main/c/jnilib.c
```diff
@@ -98,7 +102,15 @@ JNIEXPORT jint JNICALL JNI_OnLoad_netty_tcnative(JavaVM *vm, void *reserved)
     tcn_parent_pid = getppid();
 #endif
 
-    return  TCN_JNI_VERSION;  
+    TCN_LOAD_CLASS(env, byteArrayClass, "[B", JNI_ERR);
```
nmittler: do you need to convert to global references?

normanmaurer: Thats done in the macro

nmittler:
@normanmaurer LGTM.  Since we're only adding methods here, I suspect that older versions of Netty will still work with this, correct?

normanmaurer:
Nope it will break as the CertificateRequestedCallback changes

openssl-dynamic/src/main/java/org/apache/tomcat/jni/SSL.java
```diff
@@ -813,4 +813,39 @@ public static native boolean setCipherSuites(long ssl, String ciphers)
...
+     * If the {@code EVP_PKEY} pointer is used with the client certificate callback
+     * {@link CertificateRequestedCallback} the ownership goes over to OpenSsl / Tcnative and so calling
+     * {@link #freePrivateKey(long)} should <strong>NOT</strong> be done in this case. Otherwise you may
+     * need to call {@link #freePrivateKey(long)} to decrement the reference count and free memory.
+     */
+    public static native long parsePrivateKey(long privateKeyBio, String password) throws Exception;
+
+    /**
+     * Free private key ({@code EVP_PKEY} pointer).
+     *
```
Scottmitch: nit: kill line

Scottmitch:
1 nit and lgtm!

normanmaurer:
@nmittler any more comments? Also see my answer to your question ...

nmittler:
@normanmaurer nope ... ship it!

normanmaurer:
Cherry-picked into 1.1.33 (af8947887442a7c662c99e5b644ee6e9bbda5e8e)

PR #183, opened 2016-08-31
Use openssl 1.1.0 for openssl-static

normanmaurer:
Motivation:

openssl 1.1.0 was released and so we should use it when compile static against openssl.

Modifications:

Bump up openssl version.

Result:

Use latest openssl version for static compile


normanmaurer:
@nmittler @Scottmitch PTAL

nmittler:
LGTM

PR #190, opened 2016-09-21
Ensure we correctly handle parsing of the cert chain when multiple ce…

normanmaurer:
…rtificates are used.

Motivation:

We incorrectly used the return value of sk_X509_push which could lead to a double-free bug when more then 1 certificate was included in the certificate chain.

Modifications:

Correctly test for sk_X509_push(...) > 0.

Result:

No more double-free can happen.


normanmaurer:
@Scottmitch FYI

Scottmitch:
ship it

normanmaurer:
Cherry-picked as ecd8d6eae2c68259d638c503c882f984c221e4a3

PR #193, opened 2016-09-29
Compile against apr installed via brew on macos

normanmaurer:
Motivation:

Latest version of macos do not include apr anymore, we need to compile against the version that is installed via brew.

Modifications:

Define new profile that is used on osx and compile against brew installed apr.

Result:

Building tcnative on macos works again.


normanmaurer:
Cherry-picked as 1115c1c208b546a17ec677121dc9a282e805c4aa

PR #201, opened 2016-11-09
OCSP stapling support for netty-tcnative

rkapsi:
OCSP stapling support for netty-tcnative using OpenSSL's tlsext_status API. This is a low-level API and not meant to be used directly. There will be a saparate PR for Netty.

Motivation

OCSP stapling (formally known as TLS Certificate Status Request extension) is alternative approach for checking the revocation status of X.509 Certificates. Servers can preemptively fetch the OCSP response from the CA's responder, cache it for some period of time, and pass it along during (a.k.a. staple) the TLS handshake. The client no longer has to reach out on its own to the CA to check the validity of a cetitficate. Some of the key benefits are:

1) Speed. The client doesn't have to crosscheck the certificate.
2) Efficiency. The Internet is no longer DDoS'ing the CA's OCSP responder servers.
3) Safety. Less operational dependence on the CA. Certificate owners can sustain short CA outages.
4) Privacy. The CA can lo longer track the users of a certificate.

https://en.wikipedia.org/wiki/OCSP_stapling
https://letsencrypt.org/2016/10/24/squarespace-ocsp-impl.html

Modifications

https://www.openssl.org/docs/man1.0.2/ssl/SSL_set_tlsext_status_type.html

Result

Low-level API to enable OCSP stapling. There will be an another PR for Netty to make use of it.

NOTE: This PR is untested in the sense that I've been unable to build netty-tcnative due to an unknown (to me) reason.

```
[exec] file_io/unix/dir.c: In function 'apr_dir_read':
[exec] file_io/unix/dir.c:162:5: warning: 'readdir_r' is deprecated [-Wdeprecated-declarations]
[exec]      ret = readdir_r(thedir->dirstruct, thedir->entry, &retent);
[exec]      ^~~
[exec] In file included from <path>/netty-tcnative/openssl-static/target/apr-1.5.2/include/arch/unix/apr_arch_file_io.h:50:0,
[exec]                  from file_io/unix/dir.c:17:
[exec] /usr/include/dirent.h:183:12: note: declared here
[exec]  extern int readdir_r (DIR *__restrict __dirp,
[exec]
```

```
[INFO] Reactor Summary:
[INFO] 
[INFO] Netty/TomcatNative [Parent] ........................ SUCCESS [  6.616 s]
[INFO] Netty/TomcatNative [OpenSSL - Dynamic] ............. SUCCESS [ 53.135 s]
[INFO] Netty/TomcatNative [OpenSSL - Static] .............. FAILURE [ 42.609 s]
[INFO] Netty/TomcatNative [BoringSSL - Static] ............ SKIPPED
[INFO] Netty/TomcatNative [LibreSSL - Static] ............. SKIPPED
```


openssl-dynamic/src/main/java/io/netty/tcnative/jni/ocsp/StatusCode.java
```diff
@@ -0,0 +1,23 @@
...
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.tcnative.jni.ocsp;
+
+/**
+ * @see StatusCallback#callback(long)
+ */
+interface StatusCode {
+    public int value();
```
netkins: MINOR Avoid modifiers which are implied by the context [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AUnusedModifier)

openssl-dynamic/src/main/java/io/netty/tcnative/jni/ocsp/StatusCallback.java
```diff
@@ -0,0 +1,32 @@
...
+ * in the sense that {@link ServerStatusCode#SSL_TLSEXT_ERR_NOACK} will likely end the
+ * handshake while it's not mentioned at all how to opt-out from the OCSP status request.
+ * 
+ * https://www.openssl.org/docs/man1.0.2/ssl/SSL_set_tlsext_status_type.html
+ * 
+ * @see ClientStatusCode
+ * @see ServerStatusCode
+ */
+public interface StatusCallback {
+    public int callback(long ssl);
```
netkins: MINOR Avoid modifiers which are implied by the context [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=pmd%3AUnusedModifier)

rkapsi:
@normanmaurer PTAL and I'm just seeing that all the new stuff is happening on master. Shall I target that instead?

normanmaurer:
@rkapsi 1.1.33 is fine.. I am currently apply all to both.. I will review soon.

rkapsi:
@normanmaurer poke

normanmaurer:
@rkapsi sorry so busy... will try to check this week..

Scottmitch:
@rkapsi - I'm also sorry (been busy) ... regarding the build failure. What OS are you using? Have you tried updating your APR lib dependency.  If you are on macOS Sierra you should be able to grab it from homebrew. Check http://netty.io/wiki/forked-tomcat-native.html#building-on-mac-os-x

Scottmitch:
Is this feature supported by BoringSSL?

```
[INFO] libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I./src -DHAVE_OPENSSL -I<mydir>/netty-tcnative/boringssl-static/target/boringssl-chromium-stable/include -O3 -Werror -I<mydir>/git/netty-tcnative/boringssl-static/target/apr/include/apr-1 -D_LARGEFILE64_SOURCE -I/Library/Java/JavaVirtualMachines/jdk1.8.0_112.jdk/Contents/Home/include -I/Library/Java/JavaVirtualMachines/jdk1.8.0_112.jdk/Contents/Home/include/darwin -c src/io_netty_tcnative_jni_ocsp_Native.c  -fno-common -DPIC -o src/.libs/io_netty_tcnative_jni_ocsp_Native.o
[INFO] src/io_netty_tcnative_jni_ocsp_Native.c:52:5: error: implicit declaration of function 'SSL_CTX_set_tlsext_status_arg' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
[INFO]     SSL_CTX_set_tlsext_status_arg(c->ctx, NULL);
[INFO]     ^
[INFO] src/io_netty_tcnative_jni_ocsp_Native.c:52:5: note: did you mean 'SSL_CTX_set_tlsext_use_srtp'?
[INFO] <mydir>/git/netty-tcnative/boringssl-static/target/boringssl-chromium-stable/include/openssl/ssl.h:3310:20: note: 'SSL_CTX_set_tlsext_use_srtp' declared here
[INFO] OPENSSL_EXPORT int SSL_CTX_set_tlsext_use_srtp(SSL_CTX *ctx,
[INFO]                    ^
[INFO] src/io_netty_tcnative_jni_ocsp_Native.c:53:5: error: implicit declaration of function 'SSL_CTX_set_tlsext_status_cb' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
[INFO]     SSL_CTX_set_tlsext_status_cb(c->ctx, NULL);
[INFO]     ^
[INFO] src/io_netty_tcnative_jni_ocsp_Native.c:53:5: note: did you mean 'SSL_CTX_set_tlsext_status_arg'?
[INFO] src/io_netty_tcnative_jni_ocsp_Native.c:52:5: note: 'SSL_CTX_set_tlsext_status_arg' declared here
[INFO]     SSL_CTX_set_tlsext_status_arg(c->ctx, NULL);
[INFO]     ^
[INFO] src/io_netty_tcnative_jni_ocsp_Native.c:242:16: error: implicit declaration of function 'SSL_set_tlsext_status_type' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
[INFO]   long error = SSL_set_tlsext_status_type((SSL*)ssl, type);
[INFO]                ^
[INFO] src/io_netty_tcnative_jni_ocsp_Native.c:280:16: error: implicit declaration of function 'SSL_set_tlsext_status_ocsp_resp' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
[INFO]   long error = SSL_set_tlsext_status_ocsp_resp((SSL*)ssl, value, length);
[INFO]                ^
[INFO] src/io_netty_tcnative_jni_ocsp_Native.c:280:16: note: did you mean 'SSL_set_tlsext_status_type'?
[INFO] src/io_netty_tcnative_jni_ocsp_Native.c:242:16: note: 'SSL_set_tlsext_status_type' declared here
[INFO]   long error = SSL_set_tlsext_status_type((SSL*)ssl, type);
[INFO]                ^
[INFO] src/io_netty_tcnative_jni_ocsp_Native.c:300:17: error: implicit declaration of function 'SSL_get_tlsext_status_ocsp_resp' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
[INFO]   long length = SSL_get_tlsext_status_ocsp_resp((SSL*)ssl, &value);
[INFO]                 ^
[INFO] src/io_netty_tcnative_jni_ocsp_Native.c:300:17: note: did you mean 'SSL_set_tlsext_status_ocsp_resp'?
[INFO] src/io_netty_tcnative_jni_ocsp_Native.c:280:16: note: 'SSL_set_tlsext_status_ocsp_resp' declared here
[INFO]   long error = SSL_set_tlsext_status_ocsp_resp((SSL*)ssl, value, length);
[INFO]                ^
[INFO] 5 errors generated.
[INFO] make: *** [src/io_netty_tcnative_jni_ocsp_Native.lo] Error 1
[INFO] rc: 2
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO]
[INFO] Netty/TomcatNative [Parent] ........................ SUCCESS [  3.064 s]
[INFO] Netty/TomcatNative [OpenSSL - Dynamic] ............. SUCCESS [ 26.741 s]
[INFO] Netty/TomcatNative [OpenSSL - Static] .............. SUCCESS [02:59 min]
[INFO] Netty/TomcatNative [BoringSSL - Static] ............ FAILURE [01:46 min]
[INFO] Netty/TomcatNative [LibreSSL - Static] ............. SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
```

rkapsi:
@Scottmitch I'm on Linux (Arch Linux to be exact).

I believe BoringSSL doesn't support OCSP - at least https://www.imperialviolet.org/2015/10/17/boringssl.html seems to imply that.

On the other end of the spectrum there's a movement that is pushing for something that's called "OCSP Must-Staple" - https://blog.mozilla.org/security/2015/11/23/improving-revocation-ocsp-must-staple-and-short-lived-certificates/

If BoringSSL does not (and will not) support it and if some CAs turn on that flag then there'll be certain types of certs that can't be used with BoringSSL.

Scottmitch:
summoning @agl @davidben ... Can you comment if OCSP or anything similar is on the road map for BoringSSL? We can discuss offline if necessary.

@rkapsi - If BoringSSL doesn't support this we will have to work around that so we can still link/compile with BoringSSL.

davidben:
We have APIs for OCSP stapling, but they're not using OpenSSL's OCSP stack. They're very straightforward ones that take and return opaque byte strings:
https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_CTX_set_ocsp_response
https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_get0_ocsp_response
https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_enable_ocsp_stapling

We also define `OPENSSL_NO_OCSP` to report that we don't have OpenSSL's OCSP API, so if you condition code on that, it'll turn off OCSP code in both BoringSSL and OpenSSLs which configured the OCSP stuff off.

openssl-dynamic/src/main/c/io_netty_tcnative_jni_ocsp_Native.c
```diff
@@ -0,0 +1,314 @@
...
+/**
+ * Installs the OCSP stapling callback.
+ */
+JNIEXPORT jobject JNICALL Java_io_netty_tcnative_jni_ocsp_Native_newCallback
+    (JNIEnv *env, jclass clazz, jlong ctx, jboolean client, jobject callback) {
+
+  long error = -1L;
+  
+  if (ctx == 0L) {
+    fprintf(stderr, "ctx ptr is 0L");
```
Scottmitch: use `tcn_ThrowException` instead of `fprintf`?

Scottmitch: general comment when ever possible to propagate the exception (via java exception or return code) when ever possible.

openssl-dynamic/src/main/c/io_netty_tcnative_jni_ocsp_Native.c
```diff
@@ -0,0 +1,314 @@
...
+  //
+  // The client and server use slightly different return values.
+  // In particular in the event of an error. We need therefore
+  // two slightly different callback functions.
+  //
+  
+  if (client == JNI_TRUE) {
+    error = SSL_CTX_set_tlsext_status_cb(c->ctx, io_netty_tcnative_jni_ocsp_client_callback);
+    if (error <= 0L) {
+      fprintf(stderr, "newCallback/SSL_CTX_set_tlsext_status_cb (client): %li\n", error);
```
Scottmitch: throw exception with `tcn_ThrowException` after cleanup instead?

openssl-dynamic/src/main/c/io_netty_tcnative_jni_ocsp_Native.c
```diff
@@ -0,0 +1,314 @@
...
+  if (value == NULL) {
+    fprintf(stderr, "OPENSSL_malloc() failed: %i\n", length);
+    return JNI_FALSE;
+  }
+
+  (*env)->GetByteArrayRegion(env, response, 0, length, value);
+  
+  long error = SSL_set_tlsext_status_ocsp_resp((SSL*)ssl, value, length);
+  if (error <= 0L) {
+    fprintf(stderr, "setResponse/SSL_set_tlsext_status_ocsp_resp: %li\n", error);
```
Scottmitch: remove this printout and throw exception instead ... also if this is fails we should free the memory we allocated. I looked at the source code and it looks like if the method is present it will always succeed ... but if we link against a library which doesn't support it ... then it will not be freed.

openssl-dynamic/src/main/c/io_netty_tcnative_jni_ocsp_Native.c
```diff
@@ -0,0 +1,314 @@
...
+JNIEXPORT jbyteArray JNICALL Java_io_netty_tcnative_jni_ocsp_Native_getResponse
+  (JNIEnv *env, jclass clazz, jlong ssl) {
+
+  if (ssl == 0L) {
+    fprintf(stderr, "SSL ptr is 0L");
+    return NULL;
+  }
+
+  jbyte *value;
+  long length = SSL_get_tlsext_status_ocsp_resp((SSL*)ssl, &value);
```
Scottmitch: the length is actually accepted by `SSL_set_tlsext_status_ocsp_resp` is an `int` and should match here right?

openssl-dynamic/src/main/c/io_netty_tcnative_jni_ocsp_Native.c
```diff
@@ -0,0 +1,314 @@
...
+  if (length == -1L) {
+    fprintf(stderr, "getResponse/SSL_get_tlsext_status_ocsp_resp: %li\n", length);
```
Scottmitch: throw an exception instead, and combine these two if statements if we are not using any other additional error feedback mechanisms.

openssl-dynamic/src/main/c/io_netty_tcnative_jni_ocsp_Native.c
Scottmitch: maybe just cast to `jint` here to avoid casting multiple times below?

Scottmitch: also the type for `SSL_get_tlsext_status_ocsp_resp` is `unsigned char **` can you be explicit about the type conversion.

openssl-dynamic/src/main/c/io_netty_tcnative_jni_ocsp_Native.c
```diff
@@ -0,0 +1,314 @@
...
+JNIEXPORT jboolean JNICALL Java_io_netty_tcnative_jni_ocsp_Native_setResponse
+    (JNIEnv *env, jclass clazz, jlong ssl, jbyteArray response) {
+  
+  if (ssl == 0L) {
+    fprintf(stderr, "SSL ptr is 0L");
+    return JNI_FALSE;
+  }
+
+  jsize length = (*env)->GetArrayLength(env, response);
+  if ((*env)->ExceptionCheck(env)) {
```
Scottmitch: can you explain why this is necessary? If this is necessary I'm guessing we will need to do this more globally.

openssl-dynamic/src/main/c/io_netty_tcnative_jni_ocsp_Native.c
```diff
@@ -0,0 +1,314 @@
...
+  
+  //
+  // The client and server use slightly different return values.
+  // In particular in the event of an error. We need therefore
+  // two slightly different callback functions.
+  //
+  
+  if (client == JNI_TRUE) {
+    error = SSL_CTX_set_tlsext_status_cb(c->ctx, io_netty_tcnative_jni_ocsp_client_callback);
+    if (error <= 0L) {
```
Scottmitch: should we explicitly check for `== 1` here instead?

openssl-dynamic/src/main/c/io_netty_tcnative_jni_ocsp_Native.c
```diff
@@ -0,0 +1,314 @@
...
+    return NULL;
+  }
+  
+  //
+  // The client and server use slightly different return values.
+  // In particular in the event of an error. We need therefore
+  // two slightly different callback functions.
+  //
+  
+  if (client == JNI_TRUE) {
```
Scottmitch: we should be able to just use this as a boolean right? `if (client)`

openssl-dynamic/src/main/c/io_netty_tcnative_jni_ocsp_Native.c
```diff
@@ -0,0 +1,314 @@
...
+  
+  if (ctx == 0L) {
+    fprintf(stderr, "ctx ptr is 0L");
+    return NULL;
+  }
+
+  //
+  // ATTENTION: We need to create a global reference of the callback.
+  //
+  jweak callbackRef = (*env)->NewWeakGlobalRef(env, callback);
```
Scottmitch: should we control the lifetime more precisely? What if the java lifetime is shorter than the C lifetime (for some reason)? For example have this be cleaned up in `ssl_context_cleanup`.

openssl-dynamic/src/main/c/io_netty_tcnative_jni_ocsp_Native.c
```diff
@@ -0,0 +1,314 @@
...
+
+/**
+ * See:
+ *   io_netty_tcnative_jni_ocsp_client_callback
+ *   io_netty_tcnative_jni_ocsp_server_callback
+ */
+static int io_netty_tcnative_jni_ocsp_callback(int errorCode, SSL *ssl, void *arg) {
+
+  // This is conceptually not possible.
+  if (CALLBACK_CLAZZ == NULL) {
```
Scottmitch: is this check necessary? the `CALLBACK_METHOD` check should be sufficient right and `CALLBACK_CLAZZ` is not directly used in this method.

openssl-dynamic/src/main/c/io_netty_tcnative_jni_ocsp_Native.c
```diff
@@ -0,0 +1,314 @@
...
+/**
+ * The server uses SSL_TLSEXT_ERR_ALERT_FATAL for errors (it's a non-negative number).
+ */
+static int io_netty_tcnative_jni_ocsp_server_callback(SSL *ssl, void *arg) {
+  return io_netty_tcnative_jni_ocsp_callback(SSL_TLSEXT_ERR_ALERT_FATAL, ssl, arg);
+}
+
+jint io_netty_tcnative_jni_ocsp_Native_OnLoad(JNIEnv *env) {
+  jclass clazz = (*env)->FindClass(env, "io/netty/tcnative/jni/ocsp/StatusCallback");
+  if ((*env)->ExceptionCheck(env)) {
```
Scottmitch: it should be good enough to just check `if (clazz == null) { return; }` right .. the exception should be pending and be thrown when we go back to java (we do this in the epoll module).

openssl-dynamic/src/main/c/io_netty_tcnative_jni_ocsp_Native.c
```diff
@@ -0,0 +1,314 @@
...
+  freeCallback(env, c, callback);
+}
+
+/**
+ * Enables or disables OCSP stapling for the given SSLEngine.
+ */
+JNIEXPORT jboolean JNICALL Java_io_netty_tcnative_jni_ocsp_Native_setType
+    (JNIEnv *env, jclass clazz, jlong ssl, jint type) {
+
+  if (ssl == 0L) {
```
Scottmitch: `0L` -> `NULL`

openssl-dynamic/src/main/c/io_netty_tcnative_jni_ocsp_Native.c
```diff
@@ -0,0 +1,314 @@
...
+    fprintf(stderr, "SSL ptr is 0L");
+    return JNI_FALSE;
+  }
+
+  long error = SSL_set_tlsext_status_type((SSL*)ssl, type);
+  if (error <= 0L) {
+    fprintf(stderr, "setType/SSL_set_tlsext_status_type: %li\n", error);
```
Scottmitch: I think we can just remove this ... as the return status should indicate a failure occurred, and the only other error path in this class should be covered by a `tcn_ThrowException(e, "ssl is null");`.

openssl-dynamic/src/main/c/jnilib.c
```diff
@@ -110,6 +114,16 @@ JNIEXPORT jint JNICALL JNI_OnLoad_netty_tcnative(JavaVM *vm, void *reserved)
     TCN_GET_FIELD(env, keyMaterialClass, keyMaterialPrivateKeyFieldId,
                    "privateKey", "J", JNI_ERR);
 
+//
+// BoringSSL doesn't have OCSP support as far as I know.
+//   https://www.imperialviolet.org/2015/10/17/boringssl.html
+//
+#ifndef OPENSSL_IS_BORINGSSL
+    if (io_netty_tcnative_jni_ocsp_Native_OnLoad(env) != JNI_OK) {
```
Scottmitch: can we just group the code into `ssl.c` and `sslcontext.c` instead of introducing new modules? That would be more consistent and helps orient you as to what the scope the methods apply to.

openssl-dynamic/src/main/java/io/netty/tcnative/jni/ocsp/ClientStatusCode.java
```diff
@@ -0,0 +1,40 @@
...
+package io.netty.tcnative.jni.ocsp;
+
+/**
+ * These are the return values for {@link StatusCallback#callback(long)} if it's being called in client mode.
+ * 
+ * https://www.openssl.org/docs/man1.0.2/ssl/SSL_set_tlsext_status_type.html
+ * 
+ * @see StatusCallback
+ */
+public enum ClientStatusCode implements StatusCode {
```
Scottmitch: this doesn't appear to be used ... is the intention for this to be used to wrap the value returned by `StatusCallback#callback` by the caller? Should we just have an interface for the caller and interface for the server if this is the case?

openssl-dynamic/src/main/java/io/netty/tcnative/jni/ocsp/Native.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+package io.netty.tcnative.jni.ocsp;
+
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLEngine;
+
+/**
+ * https://www.openssl.org/docs/man1.0.2/ssl/SSL_set_tlsext_status_type.html
+ */
+public class Native {
```
Scottmitch: can we move these methods into `SSL.java` and `SSLContext.java`?

openssl-dynamic/src/main/java/io/netty/tcnative/jni/ocsp/StatusCallback.java
Scottmitch: `ClientStatusCode` and `ServerStatusCode` are not actually used. Is the intention that the caller will wrap the value returned by `StatusCallback#callback` in the "correct" status? If that is the case can we consider an alternative approach:
- Provide constants that can be directly used by java?
```java
public static final int CLIENT_STATUS_ACK = ..;
public static final int CLIENT_STATUS_NACK = ..;
public static final int SERVER_STATUS_OK = ..;
```
- Have 2 different interfaces. One which returns `ClientStatusCode` and one which returns `ServerStatusCode`

openssl-dynamic/src/main/java/io/netty/tcnative/jni/ocsp/package-info.java
```diff
@@ -0,0 +1,66 @@
...
+ *          0c:1e:86:96:14:04:10:28:c8:35:75:66:e2:1e:45:a6:53:e7:
+ *          dd:79:ed:4c:be:97:7c:57:a9:24:ec:55:7f:4b:94:75:3c:7a:
+ *          d3:af:9f:36:bf:3d:c7:c6:03:d9:a2:b2:c7:4e:a0:a9:a2:d7:
+ *          dd:34:fa:ea:4e:6d:68:7c:c0:4d:43:d1:ac:de:b7:71:35:73:
+ *          78:1a:e7:3d:a2:f3:c0:2c:cb:d0:1d:a9:ae:bb:e4:c7:c0:f2:
+ *          89:98:de:fa
+ * ======================================
+ * </code>
+ */
+package io.netty.tcnative.jni.ocsp;
```
Scottmitch: nit: newline

openssl-dynamic/src/main/java/io/netty/tcnative/jni/ocsp/StatusType.java
```diff
@@ -0,0 +1,37 @@
...
+    private final int value;
+
+    private StatusType(int value) {
+        this.value = value;
+    }
+
+    public int value() {
+        return value;
+    }
+}
```
Scottmitch: nit: newline

openssl-dynamic/src/main/java/io/netty/tcnative/jni/ocsp/Native.java
```diff
@@ -0,0 +1,62 @@
...
+     */
+    public static native boolean setResponse(long ssl, byte[] response);
+
+    /**
+     * Returns the OCSP staple for the given {@link SSLEngine} or {@code null} if the server
+     * didn't provide a staple.
+     *
+     * NOTE: This is only meant to be called for client {@link SSLEngine}s.
+     */
+    public static native byte[] getResponse(long ssl);
```
Scottmitch: Does it make sense to have a `ByteBuffer` variant of this (instead or in addition to)? So we can take a `ByteBuffer` as a parameter and write into it? Same question for the `setResponse` method.

openssl-dynamic/src/main/c/include/io_netty_tcnative_jni_ocsp_Native.h
```diff
@@ -0,0 +1,53 @@
...
+#define _Included_io_netty_tcnative_jni_ocsp_Native
+#ifdef __cplusplus
+extern "C" {
+#endif
+/*
+ * Class:     io_netty_tcnative_jni_ocsp_Native
+ * Method:    newCallback
+ * Signature: (JZLio/netty/tcnative/jni/ocsp/StatusCallback;)Lio/netty/tcnative/jni/ocsp/StatusCallback;
+ */
+JNIEXPORT jobject JNICALL Java_io_netty_tcnative_jni_ocsp_Native_newCallback
```
Scottmitch: Is this header file necessary? Can we use `TCN_IMPLEMENT_CALL` like in `ssl.c`?

Scottmitch:
thanks @davidben!

@rkapsi - We can add BoringSSL support as a followup effort if you aren't interested, but this PR must compile/link with BoringSSL and behave gracefully (at a minimum throw exceptions or somehow indicate and error if the functionality is not supported).

Scottmitch:
@rkapsi - Also it looks like most of these methods were introduced in OpenSSL 1.0.2. We support older versions of OpenSSL so we must be able to work gracefully with these older versions too. See [the ALPN code in ssl.c](https://github.com/netty/netty-tcnative/blob/1.1.33/openssl-dynamic/src/main/c/ssl.c#L1686) for hints.

rkapsi:
Closing and opening a new PR in a second...

Scottmitch:
@rkapsi - why did you have to open a new PR?

rkapsi:
@Scottmitch I shot myself in the foot with rebase. It seemed easier to open a new PR.

PR #204, opened 2016-12-05
Use latest centos/redhat version to verify against when release netty…

normanmaurer:
…-tcnative

Motivation:

A new version of centos was released we should verify against it when release.

Modifications:

Bump up version.

Result:

Release on latest centos version.


nmittler:
LGTM :)

normanmaurer:
Cherry-picked into 1.1.33 (a7be04e) and master (3afa95d)

PR #205, opened 2016-12-06
Add max bio size param

siyengar:
Add max bio size param

Motivation:

We observed that when the number of connections is very large
on a service, the machines started running out of memory.
On memory analysis the major consumer of memory was found
to be the bio allocation.

Modifications:

The default max bio size is 17K. This allows the parameter to be
tunable.

Result:

We've found this extremely useful for scaling large services. On
reducing the buffer size memory usage can be reduced signficantly. The
CPU usage goes up a bit when this is reduced so everyone find their
own sweet spot.

### DIFF
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -1477,9 +1477,12 @@ TCN_IMPLEMENT_CALL(void, SSL, freeSSL)(TCN_STDARGS,
     SSL_free(ssl_);
 }
 
-// Make a BIO pair (network and internal) for the provided SSL * and return the network BIO
-TCN_IMPLEMENT_CALL(jlong, SSL, makeNetworkBIO)(TCN_STDARGS,
-                                               jlong ssl /* SSL * */) {
+// Make a BIO pair (network and internal) for the provided SSL * which has a
+// maximum size of max_bio_size, and return the network BIO.
+// The value max_bio_size = 0 can be supplied to use the default BIO size.
+TCN_IMPLEMENT_CALL(jlong, SSL, makeNetworkBIO0)(TCN_STDARGS,
+                                               jlong ssl /* SSL * */,
+                                               jint max_bio_size) {
     SSL *ssl_ = J2P(ssl, SSL *);
     BIO *internal_bio;
     BIO *network_bio;
@@ -1489,7 +1492,16 @@ TCN_IMPLEMENT_CALL(jlong, SSL, makeNetworkBIO)(TCN_STDARGS,
         return 0;
     }
 
-    if (BIO_new_bio_pair(&internal_bio, 0, &network_bio, 0) != 1) {
+    if (max_bio_size < 0) {
+        tcn_ThrowException(e, "max_bio_size < 0");
+        return 0;
+    }
+
+    if (BIO_new_bio_pair(
+            &internal_bio,
+            (size_t) max_bio_size,
+            &network_bio,
+            (size_t) max_bio_size) != 1) {
         tcn_ThrowException(e, "BIO_new_bio_pair failed");
         return 0;
     }
--- a/openssl-dynamic/src/main/java/org/apache/tomcat/jni/SSL.java
+++ b/openssl-dynamic/src/main/java/org/apache/tomcat/jni/SSL.java
@@ -523,6 +523,15 @@ public final class SSL {
     public static native void freeSSL(long ssl);
 
     /**
+     * Creates a BIO with the default max BIO size.
+     *
+     * @see #makeNetworkBIO(long, int)
+     */
+    public static long makeNetworkBIO(long ssl) {
+        return makeNetworkBIO0(ssl, 0);
+    }
+
+    /**
      * Wire up internal and network BIOs for the given SSL instance.
      *
      * <b>Warning: you must explicitly free this resource by calling freeBIO</b>
@@ -531,9 +540,14 @@ public final class SSL {
      * the provided SSL instance, you must call freeBIO on the returned network BIO.
      *
      * @param ssl the SSL instance (SSL *)
+     * @param maxBioSize The maximum size of the BIO. Pass 0 to use the default max size.
      * @return pointer to the Network BIO (BIO *)
      */
-    public static native long makeNetworkBIO(long ssl);
+    public static long makeNetworkBIO(long ssl, int maxBioSize) {
+        return makeNetworkBIO0(ssl, maxBioSize);
+    }
+
+    private static native long makeNetworkBIO0(long ssl, int maxBioSize);
 
     /**
      * BIO_free


normanmaurer:
@siyengar thanks for this PR! Could you please also sign our cla and adjust the commit message to match our template?

http://netty.io
http://netty.io/wiki/writing-a-commit-message.html

Please let me know once done

normanmaurer:
@siyengar also do you use netty-tcnative as a standalone thing or with netty ?

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1477,28 +1498,21 @@ TCN_IMPLEMENT_CALL(void, SSL, freeSSL)(TCN_STDARGS,
     SSL_free(ssl_);
 }
 
-// Make a BIO pair (network and internal) for the provided SSL * and return the network BIO
+// Make a BIO pair (network and internal) for the provided SSL * with the default BIO size
+// and return the network BIO
 TCN_IMPLEMENT_CALL(jlong, SSL, makeNetworkBIO)(TCN_STDARGS,
```
normanmaurer: just remove this method and always call `makeNetworkBIO0`

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1477,28 +1498,21 @@ TCN_IMPLEMENT_CALL(void, SSL, freeSSL)(TCN_STDARGS,
...
     UNREFERENCED(o);
+    return createNetworkBIO(e, ssl, 0);
+}
 
-    SSL_set_bio(ssl_, internal_bio, internal_bio);
-
-    return P2J(network_bio);
+// Make a BIO pair (network and internal) for the provided SSL * with a max size and max size
+// of max_bio_size and return the network BIO
+TCN_IMPLEMENT_CALL(jlong, SSL, makeNetworkBIOWithSize)(TCN_STDARGS,
```
normanmaurer: rename to `makeNetworkBIO0`

openssl-dynamic/src/main/java/org/apache/tomcat/jni/SSL.java
```diff
@@ -531,8 +531,16 @@ public static native long newBIO(long pool, BIOCallback callback)
      * the provided SSL instance, you must call freeBIO on the returned network BIO.
      *
      * @param ssl the SSL instance (SSL *)
+     * @param maxBioSize The maximum size of the BIO.
      * @return pointer to the Network BIO (BIO *)
      */
+    public static native long makeNetworkBIOWithSize(long ssl, int maxBioSize);
```
normanmaurer: remove this method.

openssl-dynamic/src/main/java/org/apache/tomcat/jni/SSL.java
```diff
@@ -531,8 +531,16 @@ public static native long newBIO(long pool, BIOCallback callback)
...
+
+    /**
+     * Creates a BIO with the default max BIO size.
+     *
+     * @see #makeNetworkBIOWithSize(long, int)
+     */
     public static native long makeNetworkBIO(long ssl);
```
normanmaurer: @siyengar can you just rename this method to `makeNetworkBIO0` and have it take `maxBioSize `. Then add static public methods to preserve old API:

Something like this:

```java
public static long makeNetworkBIO(long ssl) {
    return makeNetworkBIO0(ssl, maxBioSize);
}

public static long makeNetworkBIO(long ssl, int maxBioSize) {
    return makeNetworkBIO0(ssl, maxBioSize);
}

private static native long makeNetworkBIO0(long ssl, int maxBioSize);
```

This way we can keep the changes / c code minimal. WDYT ?

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1229,6 +1229,27 @@ static void ssl_info_callback(const SSL *ssl, int where, int ret) {
     }
 }
 
+// max_bio_size = 0 is the default BIO size.
+static jlong createNetworkBIO(JNIEnv *e, jlong ssl, jint max_bio_size) {
```
normanmaurer: see below (SSL.java) for comments :)

openssl-dynamic/src/main/c/ssl.c
siyengar: will do

openssl-dynamic/src/main/c/ssl.c
siyengar: Will do

siyengar:
@normanmaurer we do use netty-tcnative with netty, however we're still on an older version of netty so in some places we use netty-tcnative directly. The project is https://github.com/facebook/nifty which is thrift over netty.

will update the CLA and message

normanmaurer:
@siyengar I see... can you share what buffer size you usually use and if you are aware of a minimum that is required ?

siyengar:
@normanmaurer I'm still experimenting with the buffer size. After I found this problem, I first tried a buffer of 1K for read and write, and that reduced memory usage significantly, by > 3gigs since we have a large number connections on this server in certain cases, however it increased CPU utilization by about 10% presumably because more ssl record encryption calls + syscalls calls need to be made for 1 application write. So we're still tuning this buffer w.r.t how much we can tolerate. Openssl and boringssl have no minimum size restriction on this buffer from reading their source.

normanmaurer:
@siyengar ok cool... let me know once you find the "sweet spot". About how many connections we are talking here on the box ?

Also say hi to @bmatheny and @andrewcox ;)

siyengar:
Ah ya, there are > 60K connections on average, which can spike sometimes, with a headroom of about 4Gigs. Openssl and boringssl do ephemeral allocations per write as well of 16K for read and write buffers (write_fragment_size), however after this change it doesn't seem to bite us much.

Will definitely say hi :), we're all enjoying using "netty" with wangle very much as well.

siyengar:
made the requested changes. Thanks for the reviews @normanmaurer

bmatheny:
Careful @normanmaurer, I stalk the netty PR's ;) Come visit next time you're here!

normanmaurer:
@bmatheny lol... I will for sure... March it is ;)

siyengar:
looks like CI failed during cmake of boringssl stable which is weird, since it passes for me locally just fine, but then again I'm building on a mac

normanmaurer:
@siyengar I installed a new ci server today and it seems like I missed to install `ninja`. Just installed it, let us see if it pass now :)

siyengar:
Thanks @normanmaurer, would love a release with this change if everything passes. I can then update open-source nifty to use this API rather than relying on internal changes.

normanmaurer:
@siyengar thats the plan at the end ...

openssl-dynamic/src/main/java/org/apache/tomcat/jni/SSL.java
```diff
@@ -531,9 +550,10 @@ public static native long newBIO(long pool, BIOCallback callback)
      * the provided SSL instance, you must call freeBIO on the returned network BIO.
      *
      * @param ssl the SSL instance (SSL *)
+     * @param maxBioSize The maximum size of the BIO. Pass 0 to use the default max size.
      * @return pointer to the Network BIO (BIO *)
      */
-    public static native long makeNetworkBIO(long ssl);
+    public static native long makeNetworkBIO0(long ssl, int maxBioSize);
```
siyengar: will do

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1477,9 +1477,12 @@ TCN_IMPLEMENT_CALL(void, SSL, freeSSL)(TCN_STDARGS,
     SSL_free(ssl_);
 }
 
-// Make a BIO pair (network and internal) for the provided SSL * and return the network BIO
-TCN_IMPLEMENT_CALL(jlong, SSL, makeNetworkBIO)(TCN_STDARGS,
-                                               jlong ssl /* SSL * */) {
+// Make a BIO pair (network and internal) for the provided SSL * with a max size and max size
```
siyengar: will do

siyengar:
addressed comments.

openssl-dynamic/src/main/java/org/apache/tomcat/jni/SSL.java
```diff
@@ -522,6 +522,16 @@ public static native long newBIO(long pool, BIOCallback callback)
      */
     public static native void freeSSL(long ssl);
 
+
```
siyengar: will do

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1489,7 +1492,7 @@ TCN_IMPLEMENT_CALL(jlong, SSL, makeNetworkBIO)(TCN_STDARGS,
         return 0;
     }
 
-    if (BIO_new_bio_pair(&internal_bio, 0, &network_bio, 0) != 1) {
+    if (BIO_new_bio_pair(&internal_bio, max_bio_size, &network_bio, max_bio_size) != 1) {
```
Scottmitch: There is an implicit cast of signed -> unsigned here which may cause warnings, which may be treated as errors depending on environment, on some compilers. Can you be explicit about the type conversion?

siyengar: great catch. will do an explicit conversion here.

siyengar:
I didn't add another check that size_t size > jint size and making that assumption of the arch here. Does netty-tcnative run on any archs where that is not true?

normanmaurer:
That should be fine

normanmaurer:
@Scottmitch satisfied as well?

Scottmitch:
1.1.33 (e691680) master (3a3a1e5)

siyengar:
@normanmaurer do you have an estimate on when Fork24 will be released?

normanmaurer:
@siyengar tomorrow 👍

siyengar:
we finally ended up using a 2k buffer which seems to work fine.

We're now toying with the idea of deallocating and allocating the buffer after each write.

Scottmitch:
@siyengar - you may be interested in https://github.com/netty/netty-tcnative/pull/216. I expect this can be used to reduce memory even further.

PR #207, opened 2016-12-08
Expose BIO_should_retry(...) method.

normanmaurer:
Motivation:

When a BIO operation fails we need to call BIO_should_retry(...) to find out if we should retry the same operation later or not.

Modifications:

Expose BIO_should_retry(...).

Result:

Its now possible to check if BIO operation should be retried later.


normanmaurer:
@Scottmitch reviewed offline... 

Cherry-picked into 1.1.33 (3f80b8905f2cc3de09a7b031351a9596ab8a6619) and master (d26de4b)

PR #208, opened 2016-12-09
Remove some unused static functions.

davidben:
By the way, I don't understand Maven will enough to mess with the build files, but if you build with `-Wunused-variable`, it should catch these and probably others.


normanmaurer:
@davidben thanks will have a look!

normanmaurer:
@davidben thanks!

Merged into master (c7cc328a0ec5cf3599646f89db106f251ae2a750)

PR #210, opened 2016-12-13
Allow to use different sizes for both BIO buffers.

normanmaurer:
Motivation:

e691680c4caaf6873ce457b11c67bed2b11ebc74 introduced a new method that allowed to set the BIO buffer size but this size is used for both buffers while sometimes different sizes can make sense.

Modifications:

Add new method that take two different buffer sizes.

Result:

More flexible way of creating network BIOs.

### DIFF
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -1543,7 +1543,7 @@ TCN_IMPLEMENT_CALL(void, SSL, freeSSL)(TCN_STDARGS,
 // The value max_bio_size = 0 can be supplied to use the default BIO size.
 TCN_IMPLEMENT_CALL(jlong, SSL, makeNetworkBIO0)(TCN_STDARGS,
                                                jlong ssl /* SSL * */,
-                                               jint max_bio_size) {
+                                               jint maxInternalBIOSize, jint maxNetworkBIOSize) {
     SSL *ssl_ = J2P(ssl, SSL *);
     BIO *internal_bio;
     BIO *network_bio;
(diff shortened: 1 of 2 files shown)


normanmaurer:
@nmittler @Scottmitch

normanmaurer:
@nmittler nope as the old java methods are still there. Only the native method changed and new java methods were added.

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -2600,9 +2605,12 @@ TCN_IMPLEMENT_CALL(void, SSL, freeSSL)(TCN_STDARGS, jlong ssl) {
   tcn_ThrowException(e, "Not implemented");
 }
 
-TCN_IMPLEMENT_CALL(jlong, SSL, makeNetworkBIO)(TCN_STDARGS, jlong ssl) {
+TCN_IMPLEMENT_CALL(jlong, SSL, makeNetworkBIO)(TCN_STDARGS, jlong ssl, jint max_bio_size1, jint max_bio_size2) {
```
Scottmitch: `max_bio_size1` -> `maxInternalBIOSize`
`max_bio_size2` -> `maxNetworkBIOSize`

openssl-dynamic/src/main/java/org/apache/tomcat/jni/SSL.java
```diff
@@ -551,10 +551,27 @@ public static long makeNetworkBIO(long ssl) {
...
+    /**
+     * Wire up internal and network BIOs for the given SSL instance.
+     *
+     * <b>Warning: you must explicitly free this resource by calling freeBIO</b>
+     *
+     * While the SSL's internal/application data BIO will be freed when freeSSL is called on
+     * the provided SSL instance, you must call freeBIO on the returned network BIO.
+     *
+     * @param ssl the SSL instance (SSL *)
+     * @param maxInternalBIOSize The maximum size of the BIO. Pass 0 to use the default max size.
```
Scottmitch: clarify that this BIO is on the application side. also do the same for the network BIO. Add a link to https://www.openssl.org/docs/man1.0.1/crypto/BIO_s_bio.html (see the example section)

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1553,16 +1553,21 @@ TCN_IMPLEMENT_CALL(jlong, SSL, makeNetworkBIO0)(TCN_STDARGS,
         return 0;
     }
 
-    if (max_bio_size < 0) {
-        tcn_ThrowException(e, "max_bio_size < 0");
+    if (maxInternalBIOSize < 0) {
```
Scottmitch: nit: most of the time this will not be violated ... I wonder if we can just reduce the amount of conditionals and combine the precondition check into a single conditional:

```c
if (maxInternalBIOSize < 0 || maxNetworkBIOSize < 0) {
  tcn_ThrowException(e, "maxInternalBIOSize and maxNetworkBIOSize both must be >= 0");
  return 0;
}
```

normanmaurer: @Scottmitch let us just keep it as it is as this one will be called only very infrequently anyway.

openssl-dynamic/src/main/java/org/apache/tomcat/jni/SSL.java
normanmaurer: +1

normanmaurer:
@Scottmitch PTAL again

normanmaurer:
Cherry-picked into 1.1.33 (c542258800261e9cf5055129dbeeda7edb88102d) and master (1fedfa6)

PR #211, opened 2016-12-19
Allow to set CA list for SSLContext via BIO

normanmaurer:
Motivation:

To correct announce CA list we need to allow setting it via a BIO so we can hook it up with the X509TrustManager.
This is needed to fix https://github.com/netty/netty/issues/6141 .

Modifications:

Add new SSLContext.setCACertifcateBio(...) method

Result:

Be able to set the to be announced CA list.

### DIFF
--- a/openssl-dynamic/src/main/c/sslcontext.c
+++ b/openssl-dynamic/src/main/c/sslcontext.c
@@ -593,6 +593,17 @@ cleanup:
     return rv;
 }
 
+TCN_IMPLEMENT_CALL(jboolean, SSLContext, setCACertificateBio)(TCN_STDARGS, jlong ctx, jlong certs)
+{
+    tcn_ssl_ctxt_t *c = J2P(ctx, tcn_ssl_ctxt_t *);
+    BIO *b = J2P(certs, BIO *);
+
+    UNREFERENCED(o);
+    TCN_ASSERT(c != NULL);
+
+    return b != NULL && c->mode != SSL_MODE_CLIENT && SSL_CTX_use_client_CA_bio(c->ctx, b) > 0 ? JNI_TRUE : JNI_FALSE;
+}
+
 TCN_IMPLEMENT_CALL(void, SSLContext, setTmpDH)(TCN_STDARGS, jlong ctx,
                                                                   jstring file)
 {
@@ -1899,6 +1910,16 @@ TCN_IMPLEMENT_CALL(jboolean, SSLContext, setCACertificate)(TCN_STDARGS,
     return JNI_FALSE;
 }
 
+TCN_IMPLEMENT_CALL(jboolean, SSLContext, setCACertificateBio)(TCN_STDARGS,
+                                                           jlong ctx,
+                                                           jlong certs)
+{
+    UNREFERENCED_STDARGS;
+    UNREFERENCED(ctx);
+    UNREFERENCED(certs);
+    return JNI_FALSE;
+}
+
 TCN_IMPLEMENT_CALL(void, SSLContext, setShutdownType)(TCN_STDARGS, jlong ctx,
                                                       jint type)
 {
@@ -1949,6 +1970,7 @@ TCN_IMPLEMENT_CALL(jboolean, SSLContext, setCertificateBio)(TCN_STDARGS, jlong c
     UNREFERENCED(idx);
     return JNI_FALSE;
 }
+
 TCN_IMPLEMENT_CALL(void, SSLContext, setNpnProtos)(TCN_STDARGS, jlong ctx, jobjectArray next_protos,
         jint selectorFailureBehavior)
 {
@@ -1956,7 +1978,6 @@ TCN_IMPLEMENT_CALL(void, SSLContext, setNpnProtos)(TCN_STDARGS, jlong ctx, jobje
     UNREFERENCED(ctx);
     UNREFERENCED(next_protos);
 }
-
 
 TCN_IMPLEMENT_CALL(void, SSLContext, setAlpnProtos)(TCN_STDARGS, jlong ctx, jobjectArray alpn_protos,
         jint selectorFailureBehavior)
--- a/openssl-dynamic/src/main/c/sslutils.c
+++ b/openssl-dynamic/src/main/c/sslutils.c
@@ -538,8 +538,7 @@ DH *SSL_callback_tmp_DH_4096(SSL *ssl, int export, int keylen)
  * format, possibly followed by a sequence of CA certificates that
  * should be sent to the peer in the SSL Certificate message.
  */
-int SSL_CTX_use_certificate_chain(SSL_CTX *ctx, const char *file,
-                                  int skipfirst)
+int SSL_CTX_use_certificate_chain(SSL_CTX *ctx, const char *file, bool skipfirst)
 {
     BIO *bio;
     int n;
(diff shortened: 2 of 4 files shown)


openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -593,6 +593,25 @@ TCN_IMPLEMENT_CALL(jboolean, SSLContext, setCACertificate)(TCN_STDARGS,
     return rv;
 }
 
+TCN_IMPLEMENT_CALL(jboolean, SSLContext, setCACertificateBio)(TCN_STDARGS,
```
Scottmitch: It doesn't look like the return value is used ... should we just make this method void and throw if something goes wrong?

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -593,6 +593,25 @@ TCN_IMPLEMENT_CALL(jboolean, SSLContext, setCACertificate)(TCN_STDARGS,
...
+                                                           jlong certs)
+{
+    tcn_ssl_ctxt_t *c = J2P(ctx, tcn_ssl_ctxt_t *);
+    BIO *b = J2P(certs, BIO *);
+
+    UNREFERENCED(o);
+    TCN_ASSERT(ctx != 0);
+
+    if (b == NULL) {
+        return JNI_FALSE;
```
Scottmitch: if we want to keep this method as boolean then we could change to this:

```c
return b != NULL && c->mode != SSL_MODE_CLIENT && SSL_CTX_use_client_CA_bio(c->ctx, b) > 0 ? JNI_TRUE : JNI_FALSE;
```

openssl-dynamic/src/main/c/sslutils.c
```diff
@@ -570,18 +570,29 @@ int SSL_CTX_use_certificate_chain_bio(SSL_CTX *ctx, BIO *bio,
...
-    /* free a perhaps already configured extra chain */
-    SSL_CTX_clear_extra_chain_certs(ctx);
-
-    /* create new extra chain by loading the certs */
     n = 0;
-    while ((x509 = PEM_read_bio_X509(bio, NULL, NULL, NULL)) != NULL) {
-        if (SSL_CTX_add_extra_chain_cert(ctx, x509) != 1) {
-            X509_free(x509);
-            return -1;
+    if (ca == 1) {
```
Scottmitch: make this data type `jboolean` and/or treat it as a boolean `if (ca) { .. }`?

openssl-dynamic/src/main/c/sslutils.c
```diff
@@ -593,6 +604,18 @@ int SSL_CTX_use_certificate_chain_bio(SSL_CTX *ctx, BIO *bio,
     return n;
 }
 
+int SSL_CTX_use_certificate_chain_bio(SSL_CTX *ctx, BIO *bio,
+                                  int skipfirst)
```
Scottmitch: move this argument to the previous line?

openssl-dynamic/src/main/java/org/apache/tomcat/jni/SSLContext.java
```diff
@@ -441,6 +441,21 @@ public static native boolean setCACertificate(long ctx, String file,
...
+     * Certificates of Certification Authorities (CA) whose clients you deal with.
+     * These are used for Client Authentication. Such a BIO is simply the
+     * concatenation of the various PEM-encoded Certificate files, in order of
+     * preference. This can be used alternatively and/or additionally to
+     * path.
+     * <br>
+     * @param ctx Server context to use.
+     * @param certBio Directory of PEM-encoded CA Certificates for Client Auth.
+     */
+    public static native boolean setCACertificateBio(long ctx, long certBio);
```
Scottmitch: make this `void` or add a `@returns` describing the return value.

openssl-dynamic/src/main/c/sslcontext.c
normanmaurer: I think we should return jboolean and use it :)

openssl-dynamic/src/main/c/sslcontext.c

openssl-dynamic/src/main/c/sslutils.c
normanmaurer: +1

openssl-dynamic/src/main/c/sslutils.c
normanmaurer: +1

openssl-dynamic/src/main/c/sslcontext.c
Scottmitch: jboolean sgtm

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -593,6 +593,17 @@ TCN_IMPLEMENT_CALL(jboolean, SSLContext, setCACertificate)(TCN_STDARGS,
     return rv;
 }
 
+TCN_IMPLEMENT_CALL(jboolean, SSLContext, setCACertificateBio)(TCN_STDARGS, jlong ctx, jlong certs)
+{
+    tcn_ssl_ctxt_t *c = J2P(ctx, tcn_ssl_ctxt_t *);
+    BIO *b = J2P(certs, BIO *);
+
+    UNREFERENCED(o);
+    TCN_ASSERT(ctx != 0);
```
Scottmitch: `ctx != 0` -> `c != NULL`

openssl-dynamic/src/main/c/sslutils.c
```diff
@@ -555,8 +554,7 @@ int SSL_CTX_use_certificate_chain(SSL_CTX *ctx, const char *file,
     return n;
 }
 
-int SSL_CTX_use_certificate_chain_bio(SSL_CTX *ctx, BIO *bio,
-                                  int skipfirst)
+static int SSL_CTX_setup_certs(SSL_CTX *ctx, BIO *bio, int skipfirst, int ca)
```
Scottmitch: what c standard do we target, c99? If so consider using `#include <stdbool.h>` and `bool`

normanmaurer: good question. I just used what we used in other methods. So keep it for now as it is ?

Scottmitch: I prefer to be explicit about types when possible in typed languages. If its an `int` then compare it to an explicit value, if its a `bool` then use it like a boolean. For example below we use the `int` like a boolean and there is nothing that describes this assumption ... making it a `bool` removes the ambiguity.

normanmaurer: @Scottmitch I think I would like to keep it as it is and cleanup all the methods in a followup PR as this should be done for all the methods in all classes where we use an int as long. Ok ?

Scottmitch: @normanmaurer - Do you have other examples where we do this? As long as we clean it up that is fine with me.

normanmaurer: actually its the same method or methods that call it... `skipfirst` is also just a boolean.

normanmaurer:
Cherry-picked into 1.1.33 (9d9a37c1f9b38104801be463244d4be587d20b7e) and master (cbd38f31f250bd5329b3dfd535c89e4ae7b1c431)

PR #212, opened 2016-12-20
Remove unused TCN* APIs.

davidben:
Here's a bit more dead code that can go away.


normanmaurer:
@davidben thanks a lot... merged into master branch :)

openssl-dynamic/src/main/c/tcn.h
```diff
@@ -162,34 +142,6 @@ char           *tcn_pstrdup(JNIEnv *, jstring, apr_pool_t *);
         }                                           \
     TCN_END_MACRO
 
-#define TCN_MAX_METHODS 8
-
-#define TCN_MIN(a, b) ((a) < (b) ? (a) : (b))
```
Scottmitch: I'm actually using this in a pending PR ... I can just add it back unless there is a better alternative to use.

Scottmitch:
thanks @davidben ... one question/comment

davidben:
I don't think there's a standard library version in C. (C++ has `std::min`.) If you use it or want to use it, I have no particular need to have any of this added or removed. I was just idly poking around and noticed another swath of dead code.

Scottmitch:
> I don't think there's a standard library version in C. (C++ has std::min.)

This is what I thought (I've seen this macro in almost every `c` project).

normanmaurer:
@Scottmotch just add it back in your PR if you need it

Scottmitch:
yip not a problem I will https://github.com/netty/netty-tcnative/pull/212#discussion_r93333424

PR #216, opened 2017-01-10
ByteBuffer BIO

Scottmitch:
Motivation:
Currently netty-tcnative uses BIO_new_bio_pair so we can control all FD lifetime and event notification in Java but delegates to OpenSSL for encryption/decryption. The current implementation sets up a pair of BIO buffers to read/write encrypted/plaintext data. This approach requires copying of data from Java ByteBuffers to native memory BIO buffers, and also requires both BIO buffers to be sufficiently large to hold application data. If direct ByteBuffers are used we can avoid copying to/from the intermediate BIO buffer and just read/write directly from the direct ByteBuffer memory. We still need an internal buffer because OpenSSL may generate write data as a result of read calls (e.g. handshake, alerts, renegotiation, etc..), but this buffer doesn't have to be be large enough to hold application data.

Modifications:
- Implement BIO which reads/writes from/to direct ByteBuffer memory.
- The BIO must also support an intermediate storage buffer for when OpenSSL generates write data but we are reading.

Result:
Allows for less memory copying and lower memory for internal buffers when used from Java.

### DIFF
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -29,6 +29,8 @@
  * limitations under the License.
  */
 
+#include <stdbool.h>
+#include <openssl/bio.h>
 #include "tcn.h"
 #include "apr_file_io.h"
 #include "apr_thread_mutex.h"
@@ -52,6 +54,17 @@ struct CRYPTO_dynlock_value {
     const char* file;
     int line;
     apr_thread_mutex_t *mutex;
+};
+
+struct TCN_bio_bytebuffer {
+    // Pointer arithmetic is done on this variable. The type must correspond to a "byte" size.
+    char* buffer;
+    char* nonApplicationBuffer;
+    jint  nonApplicationBufferSize;
+    jint  nonApplicationBufferOffset;
+    jint  nonApplicationBufferLength;
+    jint  bufferLength;
+    bool  bufferIsSSLWriteSink;
 };
 
 /*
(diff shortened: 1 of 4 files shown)


nmittler:
@Scottmitch this is great! Could you post some benchmarks?

nmittler:
FYI @kruton @davidben 

We'll want to port this to conscrypt

normanmaurer:
We have no benchmark yet but will provide one soon

nmittler:
👍

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -221,6 +234,243 @@ static const jint supported_ssl_opts = 0
 #endif
      | 0;
 
+static jint tcn_flush_sslbuffer_to_bytebuffer(struct TCN_bio_bytebuffer* bioUserData) {
+    jint writeAmount = TCN_MIN(bioUserData->bufferLength, bioUserData->sslWriteBufferLength) * sizeof(char);
+    jint writeChunk = bioUserData->sslWriteBufferSize - bioUserData->sslWriteBufferOffset;
+//fprintf(stderr, "tcn_flush_sslbuffer_to_bytebuffer1 bioUserData->sslWriteBufferLength %d bioUserData->sslWriteBufferOffset %d writeChunk %d writeAmount %d\n", bioUserData->sslWriteBufferLength, bioUserData->sslWriteBufferOffset, writeChunk, writeAmount);
```
normanmaurer: remove

openssl-dynamic/src/main/java/io/netty/tcnative/jni/SSL.java
```diff
@@ -249,28 +256,71 @@ private SSL() { }
...
+     * After you are done buffering data you should call {@link #clearBIOBuffer(long)}.
+     * @param bio {@code BIO*}.
+     * @param bufferAddress The memory address (typically from a direct {@link java.nio.ByteBuffer}) which will be used
+     *                    to either write encrypted data to, or read encrypted data from by OpenSSL's internal BIO pair.
+     * @param maxUsableBytes The maximum usable length in bytes starting at {@code bufferAddress}.
+     * @param isSSLWriteSink {@code true} if this buffer is expected to buffer data as a result of calls to {@code SSL_write}.
+     *                       {@code false} if this buffer is expected to buffer data as a result of calls to {@code SSL_read}.
      */
-    public static native int readFromBIO(long bio, long rbuf, int rlen);
+    public static native void setBIOBuffer(long bio, long bufferAddress, int maxUsableBytes, boolean isSSLWriteSink);
```
normanmaurer: @Scottmitch just a nit but should we be more consistent with naming. Like sometimes we start with `bio` sometimes we have `BIO` as suffix etc.

openssl-dynamic/src/main/c/ssl_private.h
```diff
@@ -213,6 +217,11 @@ struct tcn_ssl_ctxt_t {
     apr_uint32_t            ticket_keys_renew;
     // The client presented a ticket that did not match any key in the list.
     apr_uint32_t            ticket_keys_fail;
+
+    jmethodID               verifier_method;
+    jmethodID               cert_requested_callback_method;
```
normanmaurer: @Scottmitch did you change the ordering for any reasons like memory savings due layout ?

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -709,61 +979,69 @@ TCN_IMPLEMENT_CALL(jint, SSL, pendingReadableBytesInSSL)(TCN_STDARGS, jlong ssl
...
+TCN_IMPLEMENT_CALL(jint, SSL, flushBIOBuffer)(TCN_STDARGS, jlong bioAddress) {
+    BIO* bio = J2P(bioAddress, BIO*);
+    struct TCN_bio_bytebuffer* bioUserData;
 
-    return BIO_should_retry(b) ? JNI_TRUE : JNI_FALSE;
+    return (bio == NULL ||
+        (bioUserData = (struct TCN_bio_bytebuffer*) BIO_get_data(bio)) == NULL ||
+        bioUserData->sslWriteBufferLength == 0 ||
+        bioUserData->buffer == NULL ||
+        !bioUserData->bufferIsSSLWriteSink)
```
normanmaurer: @Scottmitch actually I think it would be easier to read if you would do the assignment etc in a separate step. I know this will cause more branches but will be easier to read at the end.

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -678,21 +935,34 @@ TCN_IMPLEMENT_CALL(jint, SSL, getError)(TCN_STDARGS,
...
         tcn_ThrowException(e, "bio is null");
         return 0;
     }
 
-    UNREFERENCED_STDARGS;
-
     return BIO_ctrl_pending(b);
 }
 
+#ifdef OPENSSL_IS_BORINGSSL
```
normanmaurer: really ? They not have the method ?

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -221,6 +234,243 @@ static const jint supported_ssl_opts = 0
...
+    }
+
+    if (bioUserData->bufferIsSSLWriteSink || bioUserData->buffer == NULL || bioUserData->bufferLength == 0) {
+        // During handshake this may happen, and it means we are not setup to read yet.
+        BIO_set_retry_read(bio);
+        return -1;
+    }
+
+    readAmount = TCN_MIN(bioUserData->bufferLength, (jint) outl) * sizeof(char);
+//fprintf(stderr, "tcn_read_from_bytebuffer readAmount %d\n", readAmount);
```
normanmaurer: remove ?

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -221,6 +234,243 @@ static const jint supported_ssl_opts = 0
...
+    bioUserData->buffer += writeChunk; // Pointer arithmetic based on char* type
+
+    return writeAmount + writeChunk;
+}
+
+static jint tcn_read_from_bytebuffer(BIO* bio, char *out, int outl) {
+    jint readAmount;
+    struct TCN_bio_bytebuffer* bioUserData = (struct TCN_bio_bytebuffer*) BIO_get_data(bio);
+    TCN_ASSERT(bioUserData != NULL);
+//fprintf(stderr, "tcn_read_from_bytebuffer bioUserData->bufferIsSSLWriteSink %d outl %d [%.*s]\n", bioUserData->bufferIsSSLWriteSink, outl, outl, out);
```
normanmaurer: remove ?

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -221,6 +234,243 @@ static const jint supported_ssl_opts = 0
...
+    }
+
+    // First check if we need to drain data queued in the internal SSL buffer.
+    if (bioUserData->sslWriteBufferLength != 0) {
+        writeAmount = tcn_flush_sslbuffer_to_bytebuffer(bioUserData);
+    }
+
+    // Next write "in" into what ever space the ByteBuffer has available.
+    writeChunk = TCN_MIN(bioUserData->bufferLength, (jint) inl) * sizeof(char);
+//fprintf(stderr, "tcn_write_to_bytebuffer2 writeChunk %d\n", writeChunk);
```
normanmaurer: remove ?

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -221,6 +234,243 @@ static const jint supported_ssl_opts = 0
...
+//fprintf(stderr, "tcn_write_to_bytebuffer sslWriteBufferFreeSpace %d\n", sslWriteBufferFreeSpace);
+        if (sslWriteBufferFreeSpace == 0) {
+            BIO_set_retry_write(bio); /* buffer is full */
+            return -1;
+        }
+
+        writeAmount = TCN_MIN(sslWriteBufferFreeSpace, (jint) inl) * sizeof(char);
+        startIndex = bioUserData->sslWriteBufferOffset + bioUserData->sslWriteBufferLength;
+        writeChunk = bioUserData->sslWriteBufferSize - startIndex;
+//fprintf(stderr, "tcn_write_to_bytebuffer bioUserData->sslWriteBufferLength %d bioUserData->sslWriteBufferOffset %d startIndex %d writeChunk %d writeAmount %d\n", bioUserData->sslWriteBufferLength, bioUserData->sslWriteBufferOffset, startIndex, writeChunk, writeAmount);
```
normanmaurer: remove ?

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -221,6 +234,243 @@ static const jint supported_ssl_opts = 0
...
+
+    if (in == NULL || inl <= 0) {
+        return 0;
+    }
+
+    // If the buffer is currently being used for reading then we have to use the internal SSL buffer to queue the data.
+    if (!bioUserData->bufferIsSSLWriteSink) {
+        jint sslWriteBufferFreeSpace = bioUserData->sslWriteBufferSize - bioUserData->sslWriteBufferLength;
+        jint startIndex;
+//fprintf(stderr, "tcn_write_to_bytebuffer sslWriteBufferFreeSpace %d\n", sslWriteBufferFreeSpace);
```
normanmaurer: remove ?

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -221,6 +234,243 @@ static const jint supported_ssl_opts = 0
...
+//fprintf(stderr, "tcn_flush_sslbuffer_to_bytebuffer2 bioUserData->sslWriteBufferLength %d bioUserData->sslWriteBufferOffset %d\n", bioUserData->sslWriteBufferLength, bioUserData->sslWriteBufferOffset);
+    return writeAmount;
+}
+
+static jint tcn_write_to_bytebuffer(BIO* bio, const char* in, int inl) {
+    jint writeAmount = 0;
+    jint writeChunk;
+    struct TCN_bio_bytebuffer* bioUserData = (struct TCN_bio_bytebuffer*) BIO_get_data(bio);
+    TCN_ASSERT(bioUserData != NULL);
+//fprintf(stderr, "tcn_write_to_bytebuffer bioUserData->bufferIsSSLWriteSink %d inl %d [%.*s]\n", bioUserData->bufferIsSSLWriteSink, inl, inl, in);
```
normanmaurer: remove ?

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -221,6 +234,243 @@ static const jint supported_ssl_opts = 0
...
+        memcpy(bioUserData->buffer, &bioUserData->sslWriteBuffer[bioUserData->sslWriteBufferOffset], (size_t) writeAmount);
+        bioUserData->sslWriteBufferOffset += writeAmount;
+        if (bioUserData->sslWriteBufferOffset == bioUserData->sslWriteBufferSize) {
+            bioUserData->sslWriteBufferOffset = 0;
+        }
+    }
+    bioUserData->sslWriteBufferLength -= writeAmount;
+    bioUserData->bufferLength -= writeAmount;
+    bioUserData->buffer += writeAmount; // Pointer arithmetic based on char* type
+//fprintf(stderr, "tcn_flush_sslbuffer_to_bytebuffer2 bioUserData->sslWriteBufferLength %d bioUserData->sslWriteBufferOffset %d\n", bioUserData->sslWriteBufferLength, bioUserData->sslWriteBufferOffset);
```
normanmaurer: remove ?

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -54,6 +56,17 @@ struct CRYPTO_dynlock_value {
     apr_thread_mutex_t *mutex;
 };
 
+struct TCN_bio_bytebuffer {
```
normanmaurer: @Scottmitch I wonder if you maybe want to move the while own bio type thing into another file ?

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -221,6 +234,243 @@ static const jint supported_ssl_opts = 0
...
+    bio_java_bytebuffer_ctrl,
+    bio_java_bytebuffer_create,
+    bio_java_bytebuffer_destroy,
+    NULL
+};
+#else
+static BIO_METHOD* bio_java_bytebuffer_methods = NULL;
+
+static void init_bio_methods(void) {
+    bio_java_bytebuffer_methods = BIO_meth_new(BIO_TYPE_FILE, "Java ByteBuffer");
```
normanmaurer: Why `BIO_TYPE_FILE ` and not `BIO_TYPE_MEM` ?

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -221,6 +234,243 @@ static const jint supported_ssl_opts = 0
...
+    BIO_meth_free(bio_java_bytebuffer_methods);
+}
+#endif
+
+static BIO_METHOD* BIO_java_bytebuffer() {
+#if OPENSSL_VERSION_NUMBER < 0x10100000L || defined(LIBRESSL_VERSION_NUMBER)
+    return &bio_java_bytebuffer_methods;
+#else
+    return bio_java_bytebuffer_methods;
+#endif
```
normanmaurer: Why do we need the different impls here ?

openssl-dynamic/src/main/java/io/netty/tcnative/jni/SSL.java
```diff
@@ -249,28 +256,71 @@ private SSL() { }
...
+    public static native int writeToBIO(long bioAddress, long wbufAddress, int wlen);
 
     /**
-     * BIO_read
-     * @param bio
-     * @param rbuf
-     * @param rlen
-     * @return
+     * Initialize the BIO for the SSL instance. This is a custom BIO which is designed to play nicely with a direct
+     * {@link java.nio.ByteBuffer}. Because it is a special BIO it requires special usage such that
```
normanmaurer: no need for fqcn ?

openssl-dynamic/src/main/java/io/netty/tcnative/jni/SSL.java
```diff
@@ -249,28 +256,71 @@ private SSL() { }
...
+     *
+     * @see #initBufferBIO(long, int)
+     */
+    public static long initBufferBIO(long ssl) {
+        return initBufferBIO(ssl, 0);
+    }
+
+    /**
+     * Initialize the BIO for the SSL instance. This is a custom BIO which is designed to play nicely with a direct
+     * {@link java.nio.ByteBuffer}. Because it is a special BIO it requires special usage such that
```
normanmaurer: same as above

openssl-dynamic/src/main/java/io/netty/tcnative/jni/SSL.java
```diff
@@ -249,28 +256,71 @@ private SSL() { }
...
     /**
-     * BIO_read
-     * @param bio
-     * @param rbuf
-     * @param rlen
-     * @return
+     * Initialize the BIO for the SSL instance. This is a custom BIO which is designed to play nicely with a direct
+     * {@link java.nio.ByteBuffer}. Because it is a special BIO it requires special usage such that
+     * {@link #setBIOBuffer(long, long, int, boolean)} and {@link #clearBIOBuffer(long)} are called in order to provide
+     * to supply data to SSL, and also to ensure the internal SSL buffering mechanism is expecting write at the appropriate times.
```
normanmaurer: Or even remove this and just point to `intBufferBIO(long, int)`

openssl-dynamic/src/main/c/ssl_private.h
Scottmitch: @normanmaurer - Yes this was the intention. To pack memory tighter and avoid padding ... and not rely upon compiler optimizations/pragma definitions to do so. At one point in development I actually modified this structure but then realized I didn't need to ... I can factor this change out into another PR if you prefer.

normanmaurer: yes extra PR would be nice.

openssl-dynamic/src/main/java/io/netty/tcnative/jni/SSL.java
normanmaurer: Also can you add to the javadocs that we not transfer the ownership and so the memory must be freed by the caller  later on ?

openssl-dynamic/src/main/c/ssl.c
Scottmitch: The APIs changed between SSL versions. `BIO_meth_new` and `BIO_meth_free` are newer.

openssl-dynamic/src/main/c/ssl.c
Scottmitch: not while I was developing this PR no.

openssl-dynamic/src/main/java/io/netty/tcnative/jni/SSL.java
Scottmitch: sure

openssl-dynamic/src/main/java/io/netty/tcnative/jni/SSL.java
Scottmitch: method has been removed

openssl-dynamic/src/main/java/io/netty/tcnative/jni/SSL.java
Scottmitch: I updated names ... ptal

Scottmitch: @normanmaurer - How would the ownership be transferred? ByteBuffer is GCed.

openssl-dynamic/src/main/c/ssl.c
Scottmitch: these are useful for debugging / visibility for now ... I will remove them before pulling in though.

openssl-dynamic/src/main/c/ssl_private.h
Scottmitch: https://github.com/netty/netty-tcnative/pull/225

openssl-dynamic/src/main/c/ssl.c
davidben: `BIO_wpending`? I'm... not sure why OpenSSL has two spellings for the same thing. (Or why we have both spellings for `BIO_pending` but only one for `BIO_wpending`. I guess nothing used `BIO_ctrl_wpending` so we didn't bother?)

Scottmitch: IIRC I was trying to be a "good citizen" when writing BIO and use existing APIs when possible but its possible I'm actually abusing these interfaces.

Here is how they are described in the header files:
https://github.com/openssl/openssl/blob/master/include/openssl/bio.h#L90
```c
# define BIO_CTRL_WPENDING       13/* opt - number of bytes still to write */
```

https://github.com/openssl/openssl/blob/master/include/openssl/bio.h#L87
```c
# define BIO_CTRL_PENDING        10/* opt - is their more data buffered */
```

I'm using `BIO_CTRL_PENDING` to mean "how many bytes are pending in the current ByteBuffer" and I'm using `BIO_CTRL_WPENDING` to mean "how many bytes are pending in the `char[]` which is part of the BIO" (which must be written to the peer). I could just create another method(s) which don't hook into the `BIO_*ctrl` mechanism if that is safer/more clear ... @davidben WDYT?

openssl-dynamic/src/main/c/ssl.c
Scottmitch: I think I will leave it here for now unless you are strong on this. It is only used in the context of `ssl` and that is how the existing code is generally partitioned.

openssl-dynamic/src/main/c/ssl.c
Scottmitch: Unless you are strong here I prefer to just leave it as is. The code is relatively self contained and I don't think it makes a significant difference either way.

openssl-dynamic/src/main/c/ssl.c
davidben: Personally, I think `BIO_ctrl` is insane, so not hooking into it seems pretty sound. But it's also possible they're appropriate here? (Didn't look at this PR.) I dunno. OpenSSL doesn't really use values, so it doesn't especially matter either way.

But my comment was just pointing out that `BIO_ctrl_wpending` and `BIO_wpending` are the exact same thing.

Scottmitch: > not hooking into it seems pretty sound.

The goal of re-using these APIs is that other applications which may use this BIO implementation won't have to have access to my "special APIs". The risk with hooking into `BIO_ctrl` is I implement these values in an unexpected way which may result in "bad behavior" from someone use may use another BIO and attempts to plug this BIO in. I think its unlikely this code will be used in another context because it already requires some special APIs so I will just make new APIs for these methods too.

> But my comment was just pointing out that BIO_ctrl_wpending and BIO_wpending are the exact same thing.

I see ... here is some additional info from OpenSSL's code base:
https://github.com/openssl/openssl/blob/master/crypto/bio/bio_lib.c#L567
```c
/*
 * It is unfortunate to duplicate in functions what the BIO_(w)pending macros
 * do; but those macros have inappropriate return type, and for interfacing
 * from other programming languages, C macros aren't much of a help anyway.
 */
```

openssl-dynamic/src/main/c/ssl.c
Scottmitch: cpy/pasta ... changed to `BIO_TYPE_MEM` ... although I'm not sure this matters. In `1.1.0` OpenSSL added [BIO_get_new_index](https://www.openssl.org/docs/man1.1.0/crypto/BIO_meth_new.html) to generate a new index.

Scottmitch:
@normanmaurer - PTAL

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -221,6 +234,290 @@ static const jint supported_ssl_opts = 0
...
+    bioUserData->bufferLength = 0;
+    bioUserData->bufferIsSSLWriteSink = false;
+    bioUserData->nonApplicationBuffer = NULL;
+    bioUserData->nonApplicationBufferSize = 0;
+    bioUserData->nonApplicationBufferOffset = 0;
+    bioUserData->nonApplicationBufferLength = 0;
+
+    BIO_set_data(bio, bioUserData);
+
+    // https://www.openssl.org/docs/man1.1.0/crypto/BIO_set_data.html
```
Scottmitch: this actually link was meant to help provide motivation for the call to `BIO_set_shutdown` and `BIO_set_init` which is necessary to inform OpenSSL of the BIO's initialization state. I can add comments to this effect but I think the comment belongs here.

Scottmitch:
@normanmaurer - Any more comments?

normanmaurer:
@Scottmitch nope all good! Nice work, 🍻 on me

normanmaurer:
@trustin maybe you are also interested in this

openssl-dynamic/src/main/c/ssl.c
normanmaurer: ok fine by me

Scottmitch:
I just cleaned up the ssl_private.h style fixes ... I will pull in tomorrow after running some benchmarks.

Scottmitch:
master (826513b)

Scottmitch:
Initial benchmark results show no concerns. I will address any concerns that pop up in the future.

PR #219, opened 2017-01-19
add some function to ssl.c for client side session resumption.

dark264sh:
Motivation:
It need getSession, setSession and freeSession for client side session resumption.

Modifications:
Add getSession(SSL_get1_session), setSession(SSL_set_session), freeSession(SSL_SESSION_free) method.

Result:
Be able to handle session for client side session resumption.

### DIFF
--- a/openssl-dynamic/src/main/java/org/apache/tomcat/jni/SSL.java
+++ b/openssl-dynamic/src/main/java/org/apache/tomcat/jni/SSL.java
@@ -881,4 +881,32 @@ public final class SSL {
      * Free x509 chain ({@code STACK_OF(X509)} pointer).
      */
     public static native void freeX509Chain(long x509Chain);
+
+    /**
+     * Call SSL_get1_session.
+     * https://www.openssl.org/docs/man1.1.0/ssl/SSL_get1_session.html
+     *
+     * @param ssl the SSL instance (SSL *)
+     * @return pointer to SSL_SESSION instance (SSL_SESSION *)
+     */
+    public static native long getSession(long ssl);
+
+    /**
+     * Call SSL_set_session.
+     * https://www.openssl.org/docs/man1.0.1/ssl/SSL_set_session.html
+     *
+     * @param ssl the SSL instance (SSL *)
+     * @param sess the SSL_SESSION instance (SSL_SESSION *)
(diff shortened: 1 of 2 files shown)


Scottmitch:
Thanks for the PR @dark264sh! Can you sign the [ICLA](https://docs.google.com/spreadsheet/viewform?formkey=dHBjc1YzdWhsZERUQnhlSklsbG1KT1E6MQ) please?

dark264sh:
I've signed the ICLA.

normanmaurer:
LGTM... @Scottmitch WDYT ?

openssl-dynamic/src/main/java/org/apache/tomcat/jni/SSL.java
```diff
@@ -881,4 +881,32 @@ public static native void setCertificateBio(
      * Free x509 chain ({@code STACK_OF(X509)} pointer).
      */
     public static native void freeX509Chain(long x509Chain);
+
+    /**
+     * Call SSL_get1_session.
+     * https://www.openssl.org/docs/man1.1.0/ssl/SSL_get1_session.html
```
Scottmitch: make this a html hyperlink? `<a href="">..</a>`

Scottmitch: same comment for links below

dark264sh: OK.

dark264sh:
Now we are testing. We has a plan to submit a PR for Netty.

normanmaurer:
@dark264sh nice... would be cool to see how this is used with netty together before merge this :)

Scottmitch:
@normanmaurer +1.

We currently hide all of the session stuff behind other APIs today (e.g. `getTime`). I'd be curious if we could do the same thing here. Otherwise I would like to see the lifetime implications these new APIs on the session objects.

dark264sh:
I understand your concern. I have not thought about the lifetime of ssl session. After thinking and studying about ssl session, I will send pull request again. Thank you for your review.

Scottmitch:
sounds good @dark264sh ... feel free to open a new PR when you have a better understanding for lifetime and usage from java.

PR #220, opened 2017-01-19
SSL_CTX_add_client_CA does not take ownership of the X509 cert

Scottmitch:
Motivation:
OpenSSL's SSL_CTX_add_client_CA does not transfer ownership of the x509 certificate. Instead it just makes a copy of the subject name via X509_get_subject_name. We should always call X509_free after the call to SSL_CTX_add_client_CA.

Modifications:
- Call X509_free after calling SSL_CTX_add_client_CA

Result:
No more memory leak.

### DIFF
--- a/openssl-dynamic/src/main/c/sslutils.c
+++ b/openssl-dynamic/src/main/c/sslutils.c
@@ -575,6 +575,10 @@ static int SSL_CTX_setup_certs(SSL_CTX *ctx, BIO *bio, bool skipfirst, bool ca)
                 X509_free(x509);
                 return -1;
             }
+            // SSL_CTX_add_client_CA does not take ownership of the x509. It just calls X509_get_subject_name
+            // and make a duplicate of this value. So we should always free the x509 after this call.
+            // See https://github.com/netty/netty/issues/6249.
+            X509_free(x509);
             n++;
         }
     } else {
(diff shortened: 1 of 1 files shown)


openssl-dynamic/src/main/c/sslutils.c
```diff
@@ -575,6 +575,9 @@ static int SSL_CTX_setup_certs(SSL_CTX *ctx, BIO *bio, bool skipfirst, bool ca)
                 X509_free(x509);
                 return -1;
             }
+            // SSL_CTX_add_client_CA does not take ownership of the x509. It just calls X509_get_subject_name
+            // and make a duplicate of this value. So we should always free the x509 after this call.
+            X509_free(x509);
```
normanmaurer: @Scottmitch can you also add a link to the netty issue it fixes or at least link it in the PR. Also please add a comment that while this one needs this handling `SSL_CTX_add_extra_chain_cert` needs not...  You can also use "strong words" to express your feelings ;)

Scottmitch: +1

I'll add a link to https://github.com/netty/netty/issues/6249 and comment above the call to `SSL_CTX_add_extra_chain_cert`.

Scottmitch:
1.1.33 (e3d920a) master (3a25a7d)

PR #221, opened 2017-01-26
Explicit specify the c99 should be used when compiling boringssl-static

normanmaurer:
Motivation:

On some linux distributions its needed to explicit specify that c99 should be used when compile boringssl-static. If this is not done the build will fail.

Modifications:

Ensure we use c99 when compile boringssl-static.

Result:

No more compile errors on some linux distributions


normanmaurer:
Cherry-picked into 1.1.33 (6c7d220946b5a3723a64021cd7e2562ea385cb6d) and master (22a97a9)

PR #222, opened 2017-01-26
Update openssl to 1.0.2k

normanmaurer:
Motivation:

New version of openssl was released

Modifications:

Update version.

Result:

Use up to date version.


normanmaurer:
Cherry-Picked into 1.1.33 (8119d20) and master (afe9fce)

Scottmitch:
does boringssl also have any updates we need to consume?

normanmaurer:
We do this automatically if they have as we just pull the branch...

PR #225, opened 2017-02-03
tcn_ssl_ctxt_t alignment to avoid padding

Scottmitch:
Motivation:
tcn_ssl_ctxt_t is not properly aligned which may result in increased memory or additional padding inserted. Relying on the compiler optimization can be fragile and lead to inconsistencies between compilers.

Modifications:
- Align internals of tcn_ssl_ctxt_t to be grouped in 64 byte chunks.

Result:
More proper memory alignment of the tcn_ssl_ctxt_t structure.

### DIFF
--- a/openssl-dynamic/src/main/c/ssl_private.h
+++ b/openssl-dynamic/src/main/c/ssl_private.h
@@ -166,42 +166,41 @@ typedef struct {
 } tcn_ssl_ticket_key_t;
 
 struct tcn_ssl_ctxt_t {
-    apr_pool_t              *pool;
-    SSL_CTX                 *ctx;
+    apr_pool_t*             pool;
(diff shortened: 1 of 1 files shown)


normanmaurer:
@Scottmitch LGTM... that said I wonder if you want to add some comments about the size of the different fields and so make it easier to understand and so do the right thing when a new one is added

Scottmitch:
I think this is just a general consideration when modifying c structures. IMO it would be overkill to comment on every c structure about how alignment was done. When the fields are added/removed we should consider alignment/padding.

Scottmitch:
master (6756ffd)

normanmaurer:
@Scottmitch ok fair enough... I also set the milestone, please do next time ;)

PR #227, opened 2017-02-07
Update to libressl 2.4.5

normanmaurer:
Motivation:

New version of libressl was released.

Modifications:

Update to libressl 2.4.5

Result:

Use latest libressl version


normanmaurer:
Cherry-picked into master (f8c9134)

Issue #52, opened 2015-07-15
sslcontext.c load_pem_key method looping


Scottmitch:
The load_pem_key method in sslcontext.c has a loop to call `PEM_read_bio_PrivateKey` 3 times (or until success).  Why is this loop necessary, and why 3 iterations?  We should at least clarify this situation with comments.

Scottmitch:
@normanmaurer - FYI. Lets update this to link to mention new code in the future.

normanmaurer:
Let me fix this

normanmaurer:
Just noticed I forgot to close this one.

PR #228, opened 2017-02-13
Remove unnessary loop during read the private key.

normanmaurer:
Motivation:

Looping when reading the private key is not needed anymore as we not support reading the password from the commandline and so re-reading the password will make no difference.

Modifications:

Remove loop and only try to read the private key one time.

Result:

Fixes [#52].


Scottmitch:
good to push upstream

normanmaurer:
Cherry-picked into master (da44f23)

PR #229, opened 2017-02-13
Fix 'SSL3_SEND_SERVER_KEY_EXCHANGE:missing tmp dh key' when using lib…

normanmaurer:
…ressl

Motivation:

We need to ensure we correctly call SSL_dh_get_tmp_param(...) when using libressl.

Modifications:

Call SSL_dh_get_tmp_param(...) when libressl is used.

Result:

No more failures du missing tmp dh key.


normanmaurer:
Cherry-picked into master (f070b91)

PR #230, opened 2017-02-13
Rename package to internal to make it clear its for netty use only.

normanmaurer:
Motivation:

We should put everything in an internal package to make it clear its for netty usage only and so we are free to break it at any time.

Modifications:

Move everything to an internal package.

Result:

Ensure peple are aware its an internal project.

### DIFF
--- a/openssl-dynamic/src/main/c/jnilib.c
+++ b/openssl-dynamic/src/main/c/jnilib.c
@@ -68,7 +68,7 @@ JNIEXPORT jint JNICALL JNI_OnLoad_netty_tcnative(JavaVM *vm, void *reserved)
     apr_version(&apv);
     apvn = apv.major * 1000 + apv.minor * 100 + apv.patch;
     if (apvn < 1201) {
-        tcn_Throw(env, "Unupported APR version (%s)",
+        tcn_Throw(env, "Unsupported APR version (%s)",
                   apr_version_string());
         return JNI_ERR;
     }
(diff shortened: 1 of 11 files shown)


Scottmitch:
@nmittler - The master branch of netty-tcnative includes a major version bump. The intention is that 2.0 netty-tcnative won't be restricted to semantic versioning (which we weren't following strictly with 1.1.33) and is really an implementation detail of Netty.

nmittler:
@Scottmitch Conscrypt actually goes some way to dealing with this by keeping the engine class together with the native code in a single versioned module. If you assume Netty to be the only consumer netty-tcnative, then moving it into the Netty repo makes a lot of sense.  However, if you expect it to be an independent library, perhaps the Conscrypt approach is better?

Scottmitch:
> moving it into the Netty repo makes a lot of sense

I think this is closer to our mental model and how we generally treat the project. I would actually be in favor of this so versioning mismatch is no longer an issue. We would have to support shading in a similar fashion as we do with `EPOLL` (we should anyways but just haven't had a need to yet).

normanmaurer:
@Scottmitch yes at the end it would be nice to make it part of netty but this will need some more work and thoughts as we need to also support windows etc. 

Let us do this as next step.

normanmaurer:
@Scottmitch I will merge this one as soon as your PRs are merged as I not want to produce merging hell for you :)

Scottmitch:
@normanmaurer - I agree it needs more work. I'm fine with pushing out another release with the current structure so we can get a Netty release out with bug fixes.

normanmaurer:
Cherry-picked into master (92f7497668a2e7ba93d5d3854e211e585f43324c)

PR #231, opened 2017-02-13
Update defines and outdated options

Scottmitch:
Motivation:
- tcnative re-defines constants instead of deriving them from the underlying OpenSSL library. This can lead to setting incorrect options when different implementations are used. For example BoringSSL doesn't support all the same SSL options as OpenSSL and defines some constants to 0. This may result in tcnative setting unknown SSL options.
- There are options which we don't set (e.g. SSL_OP_LEGACY_SERVER_CONNECT) which we should explicitly set to disable insecure options in case the underlying OpenSSL library enabled these compatibility options by default.

Modifications:
- SSL.java should derive constant values from native method calls
- Remove options which aren't used, and are not advised to be used due to security
- Disable SSL_OP_LEGACY_SERVER_CONNECT and SSL_OP_ALLOW_UNSAFE_LEGACY_RENEGOTIATION explicitly. These options may lead to insecure behavior.

Result:
Constants are derived from the OpenSSL library implementation and options are updated.

### DIFF
--- a/openssl-dynamic/src/main/c/sslcontext.c
+++ b/openssl-dynamic/src/main/c/sslcontext.c
@@ -232,7 +232,6 @@ TCN_IMPLEMENT_CALL(jlong, SSLContext, make)(TCN_STDARGS, jint protocol, jint mod
     c->mode     = mode;
     c->ctx      = ctx;
     c->pool     = p;
-    SSL_CTX_set_options(c->ctx, SSL_OP_ALL);
     if (!(protocol & SSL_PROTOCOL_SSLV2))
         SSL_CTX_set_options(c->ctx, SSL_OP_NO_SSLv2);
     if (!(protocol & SSL_PROTOCOL_SSLV3))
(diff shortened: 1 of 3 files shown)


openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -232,7 +232,6 @@ TCN_IMPLEMENT_CALL(jlong, SSLContext, make)(TCN_STDARGS, jint protocol, jint mod
     c->mode     = mode;
     c->ctx      = ctx;
     c->pool     = p;
-    SSL_CTX_set_options(c->ctx, SSL_OP_ALL);
```
normanmaurer: @Scottmitch why remove this ?

Scottmitch: OP_ALL is implementation specific (not well defined as to what it includes and what it doesn't) and may include things that we don't want. For example in OpenSSL 1.0.2 it enables `SSL_OP_LEGACY_SERVER_CONNECT` which allows for insecure renegotiation ... I have been advised to avoid this https://github.com/openssl/openssl/issues/2551#issuecomment-277543068.

I don't think this is a good default and if someone needs it I would prefer to add support for the specific options and allow them to manually enable these, but not do it by default.

normanmaurer: got it...

normanmaurer:
Ship it!

Scottmitch:
master (e56115b)

PR #232, opened 2017-02-13
setVerify issues and updates

Scottmitch:
Motivation:
SSL#setVerify and SSLContext#setVerify both attempt to initialize a default trust store which can be derived from the file system and configured with environment variables. This may result in unintentional certificates being trusted, and we should instead only trust the certificates that are programmatically set. The verification callback also supports an unused mode SSL_CVERIFY_OPTIONAL_NO_CA which is not used. Calling SSL#setVerify can also manpulate the state for the entire SSLContext which is not desired.

Motivation:
- setVerify should not initialize a trust store based upon file or environment
- Remove SSL_CVERIFY_OPTIONAL_NO_CA
- We should not specify a callback for SSL_set_verify and SSL_CTX_set_verify because they will not be called because we override the default verification mechanism via SSL_CTX_set_cert_verify_callback
- Decouple the state for SSL#setVerify and SSLContext#setVerify

Result:
setVerify doesn't have unexpected side effects.

### DIFF
--- a/openssl-dynamic/src/main/c/sslutils.c
+++ b/openssl-dynamic/src/main/c/sslutils.c
@@ -135,30 +135,25 @@ const char* SSL_cipher_authentication_method(const SSL_CIPHER* cipher){
  */
 static int SSL_app_data2_idx = -1;
 static int SSL_app_data3_idx = -1;
-void SSL_init_app_data2_3_idx(void)
+static int SSL_app_data4_idx = -1;
+void SSL_init_app_data_idx()
 {
     int i;
 
-    if (SSL_app_data2_idx > -1) {
-        return;
+    if (SSL_app_data2_idx == -1) {
+        /* we _do_ need to call this two times */
+        for (i = 0; i <= 1; i++) {
+            SSL_app_data2_idx = SSL_get_ex_new_index(0, "tcn_ssl_ctxt_t*", NULL, NULL, NULL);
+        }
     }
 
-    /* we _do_ need to call this two times */
-    for (i = 0; i <= 1; i++) {
-        SSL_app_data2_idx =
-            SSL_get_ex_new_index(0,
-                                 "Second Application Data for SSL",
-                                 NULL, NULL, NULL);
+    if (SSL_app_data3_idx == -1) {
+        SSL_app_data3_idx = SSL_get_ex_new_index(0, "int* handshakeCount", NULL, NULL, NULL);
     }
 
-    if (SSL_app_data3_idx > -1) {
-        return;
+    if (SSL_app_data4_idx == -1) {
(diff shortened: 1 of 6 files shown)


openssl-dynamic/src/main/c/sslutils.c
```diff
@@ -565,39 +570,30 @@ int tcn_X509_up_ref(X509* cert) {
...
 
-    if (errdepth > depth) {
-        /* TODO: Some logging
-         * Certificate Verification: Certificate Chain too long
-         */
-        ok = 0;
+    switch (tcn_mode) {
+      case SSL_CVERIFY_IGNORED:
+        switch (c->verify_mode) {
+          case SSL_CVERIFY_NONE: return SSL_VERIFY_NONE;
```
normanmaurer: nit: not strong on this but I think its more readable to have the return in the next line

openssl-dynamic/src/main/c/sslutils.c
```diff
@@ -565,39 +570,30 @@ int tcn_X509_up_ref(X509* cert) {
...
-        return 1;
-
-    if (SSL_VERIFY_ERROR_IS_OPTIONAL(errnum) &&
-        (verify == SSL_CVERIFY_OPTIONAL_NO_CA)) {
-        ok = 1;
+int tcn_set_verify_config(tcn_ssl_verify_config_t* c, jint tcn_mode, jint depth) {
+    if (depth > 0) {
+        c->verify_depth = depth;
+    } else {
+        c->verify_depth = 0;
```
normanmaurer: c->verify_depth = depth > 0 ? depth : 0

openssl-dynamic/src/main/c/sslutils.c
```diff
@@ -135,30 +135,25 @@ const char* SSL_cipher_authentication_method(const SSL_CIPHER* cipher){
...
     }
 
-    /* we _do_ need to call this two times */
-    for (i = 0; i <= 1; i++) {
-        SSL_app_data2_idx =
-            SSL_get_ex_new_index(0,
-                                 "Second Application Data for SSL",
-                                 NULL, NULL, NULL);
+    if (SSL_app_data3_idx == -1) {
+        SSL_app_data3_idx = SSL_get_ex_new_index(0, "int* hanshakeCount", NULL, NULL, NULL);
```
normanmaurer: not introduced by you but fix typo in `handshakeCount `

Scottmitch: I actually did introduce this in this PR :) thx for pointing this out.

openssl-dynamic/src/main/c/sslutils.c
Scottmitch: sgtm

Scottmitch:
@normanmaurer - comments addressed

Scottmitch:
master (d34ee40)

PR #233, opened 2017-02-13
Support wrapping/unwrapping of multiple packets

Scottmitch:
Motivation:
The JDK SSLEngine documentation says that a call to wrap/unwrap "will attempt to consume one complete SSL/TLS network packet" [1]. This limitation can result in thrashing in the pipeline to decode and encode data that may be spread amongst multiple SSL/TLS network packets.

Modifications:
- Add methods for SSL_set_mode and SSL_get_mode and SSL_MODE_ENABLE_PARTIAL_WRITE.
- Use OpenSSL to derive the limits for plaintext and ciphertext length.

Result:
SSL_set_mode and SSL_get_mode are supported and we get limits directly from OpenSSL.

### DIFF
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -92,6 +92,27 @@ struct TCN_bio_bytebuffer {
     R |= SSL_TMP_KEY_INIT_DH(2048);             \
     R |= SSL_TMP_KEY_INIT_DH(4096)
 
+#if !defined(OPENSSL_IS_BORINGSSL)
+// This is the maximum overhead when encrypting plaintext as defined by
+// <a href="https://www.ietf.org/rfc/rfc5246.txt">rfc5264</a>,
+// <a href="https://www.ietf.org/rfc/rfc5289.txt">rfc5289</a> and openssl implementation itself.
+//
+// Please note that we use a padding of 16 here as openssl uses PKC#5 which uses 16 bytes while the spec itself
+// allow up to 255 bytes. 16 bytes is the max for PKC#5 (which handles it the same way as PKC#7) as we use a block
+// size of 16. See <a href="https://tools.ietf.org/html/rfc5652#section-6.3">rfc5652#section-6.3</a>.
+//
+// 16 (IV) + 48 (MAC) + 1 (Padding_length field) + 15 (Padding) + 1 (ContentType) + 2 (ProtocolVersion) + 2 (Length)
+//
+// TODO(scott): We may need to review this calculation once TLS 1.3 becomes available.
+//              Which may add a KeyUpdate in front of the current record.
+#define TCN_MAX_ENCRYPTED_PACKET_LENGTH (16 + 48 + 1 + 15 + 1 + 2 + 2)
+
+// This also includes the header overhead for TLS 1.2 and below.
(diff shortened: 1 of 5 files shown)


openssl-dynamic/src/main/c/ssl.c
```diff
@@ -92,6 +92,25 @@ struct TCN_bio_bytebuffer {
...
+// 16 (IV) + 48 (MAC) + 1 (Padding_length field) + 15 (Padding) + 1 (ContentType) + 2 (ProtocolVersion) + 2 (Length)
+//
+// TODO(scott): We may need to review this calculation once TLS 1.3 becomes available.
+//              Which may add a KeyUpdate in front of the current record.
+#define TCN_MAX_ENCRYPTED_PACKET_LENGTH (15 + 48 + 1 + 16 + 1 + 2 + 2)
+
+// This also includes the header overhead for TLS 1.2 and below.
+// See SSL#getMaxWrapOverhead for the overhead based upon the SSL*
+// TODO(scott): this may be an over estimate because we don't account for short headers.
+#define TCN_MAX_SEAL_OVERHEAD_LENGTH (TCN_MAX_ENCRYPTED_PACKET_LENGTH + SSL3_RT_HEADER_LENGTH)
```
normanmaurer: @davidben PTAL

nmittler: Do these only need to be defined if !boringssl?

Scottmitch: Note that `TCN_MAX_ENCRYPTED_PACKET_LENGTH` is taken from Netty [1]. I discussed briefly with @davidben but I would appreciate a look at the code if time permits :) The closest thing I could find in OpenSSL is [SSL3_RT_MAX_ENCRYPTED_OVERHEAD](https://github.com/openssl/openssl/blob/master/include/openssl/ssl3.h#L173) but this includes a bunch of slop IIUC (I'll defer to @davidben for a more thorough description).

[1] https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslEngine.java#L168

> Do these only need to be defined if !boringssl?

@nmittler - Yes sir. Let me `#ifdef` this.

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1595,6 +1621,50 @@ TCN_IMPLEMENT_CALL(jint, SSL, getOptions)(TCN_STDARGS, jlong ssl)
...
+    if (ssl == NULL) {
+        tcn_ThrowException(e, "ssl is null");
+        return 0;
+    }
+
+#ifdef OPENSSL_IS_BORINGSSL
+    return (jint) SSL_max_seal_overhead(ssl);
+#else
+    // TODO(scott): When OpenSSL supports something like SSL_max_seal_overhead ... use it!
+    // TODO(scott): If we support SSL_MODE_CBC_RECORD_SPLITTING this must be calculated dynamically!
```
normanmaurer: @Scottmitch please open issues to track these

normanmaurer:
@Scottmitch ship it!

Scottmitch:
I prefer to wait until after 4.1.9 to pull this in. I would still like to do benchmarking and get some more milage on it before it goes into a release.

Scottmitch:
master (04d68022)

PR #234, opened 2017-02-14
LibreSSL workaround for unsupported X509_V_ERR_UNSPECIFIED

Scottmitch:
Motivation:
LibreSSL 2.4.x does not support X509_V_ERR_UNSPECIFIED. Until we update to LibreSSL 2.5.x we need to introduce a work around for lack of support for X509_V_ERR_UNSPECIFIED.

Modifications:
- Use X509_V_ERR_CERT_REJECTED for LibreSSL instead of X509_V_ERR_UNSPECIFIED.

Result:
We can compile with LibreSSL 2.4.x.


normanmaurer:
Ship it!

Scottmitch:
master (0a010b6)

PR #236, opened 2017-02-16
CertificateVerifier should get values from OpenSSL

Scottmitch:
Motivation:
CertificateVerifier currently hard codes in values for the x509 error numbers. These numbers are not guaranteed to be the same or even supported across the different versions of OpenSSL and the different flavors of OpenSSL (Boring, Libre).

Modifications:
- Ensure we derive the x509 errors from OpenSSL, and account for any that are not supported at compile time. Note that there still may be runtime incompatibilities if dynamically linking with a library version that doesn't support these defines.

Result:
More correct x509 error codes and support.


openssl-dynamic/src/main/c/ssl.c
```diff
@@ -891,6 +891,318 @@ TCN_IMPLEMENT_CALL(jint, SSL, sslErrorWantAccept)(TCN_STDARGS) {
     return SSL_ERROR_WANT_ACCEPT;
 }
 
+TCN_IMPLEMENT_CALL(jint, SSL, x509vOK)(TCN_STDARGS) {
```
Scottmitch: we can move these "static defines" into another file as a followup if you prefer @normanmaurer. I just put them here for now because that is where everything else is defined.

openssl-dynamic/src/main/c/ssl_private.h
```diff
@@ -152,6 +152,15 @@ extern const char* TCN_UNKNOWN_AUTH_METHOD;
 
 extern void *SSL_temp_keys[SSL_TMP_KEY_MAX];
 
+// HACK!
+// LibreSSL 2.4.x doesn't support the X509_V_ERR_UNSPECIFIED so we introduce a work around to make sure a supported alert is used.
+// This should be reverted when we support LibreSSL 2.5.x (which does support X509_V_ERR_UNSPECIFIED).
+#ifndef X509_V_ERR_UNSPECIFIED
+#define TCN_X509_V_ERR_UNSPECIFIED 99999
```
Scottmitch: yes thats the idea ... just so we can distinguish it from other error codes that may be legitimately defined in OpenSSL but we just don't yet support.

normanmaurer: Ok... let us just hope they not have the crazy idea to use it ;) ship it!

Scottmitch: @normanmaurer - yip there is a risk ... but with the current defines I don't think we can do any better (there than maybe choosing a more "lucky" number)

Scottmitch:
master (4ec3e8f)

openssl-dynamic/src/main/c/ssl_private.h
normanmaurer: yep ... like I said just shiiiiiiip it :)

PR #237, opened 2017-02-17
Support explicitly enabling hostname verificaiton

Scottmitch:
Motivation:
OpenSSL doesn't automatically verify hostnames and requires extract method calls to enable this feature [1]. We should allow this to be configured.

Modifications:
- Enabled calling of X509_VERIFY_PARAM_set_hostflags and X509_VERIFY_PARAM_set1_host

Result:
Support for hostname verification.

[1] https://wiki.openssl.org/index.php/Hostname_validation

### DIFF
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -891,6 +891,38 @@ TCN_IMPLEMENT_CALL(jint, SSL, sslErrorWantAccept)(TCN_STDARGS) {
     return SSL_ERROR_WANT_ACCEPT;
 }
 
+TCN_IMPLEMENT_CALL(jint, SSL, x509CheckFlagAlwaysCheckSubject)(TCN_STDARGS) {
+#ifdef X509_CHECK_FLAG_ALWAYS_CHECK_SUBJECT
+    return X509_CHECK_FLAG_ALWAYS_CHECK_SUBJECT;
+#else
+    return 0;
+#endif
+}
+
+TCN_IMPLEMENT_CALL(jint, SSL, x509CheckFlagDisableWildCards)(TCN_STDARGS) {
+#ifdef X509_CHECK_FLAG_NO_WILD_CARDS
+    return X509_CHECK_FLAG_NO_WILD_CARDS;
+#else
+    return 0;
+#endif
+}
+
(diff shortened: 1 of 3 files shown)


normanmaurer:
@Scottmitch this looks great I just wonder if we not need to take the algorithm into account. See https://github.com/netty/netty/pull/6403#discussion_r101694705 .

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1979,6 +2011,8 @@ TCN_IMPLEMENT_CALL(void, SSL, setVerify)(TCN_STDARGS, jlong ssl, jint level, jin
            tcn_ThrowException(e, "failed to allocate tcn_ssl_verify_config_t");
            return;
        }
+       // Copy the verify depth form the context in case depth is <0.
+       verify_config->verify_depth = c->verify_config.verify_depth;
```
normanmaurer: @Scottmitch should this be a different PR ? Seems like its not really related.

openssl-dynamic/src/main/c/sslutils.c
```diff
@@ -571,7 +571,9 @@ int tcn_X509_up_ref(X509* cert) {
 }
 
 int tcn_set_verify_config(tcn_ssl_verify_config_t* c, jint tcn_mode, jint depth) {
-    c->verify_depth = depth > 0 ? depth : 0;
+    if (depth >= 0) {
+        c->verify_depth = depth;
+    }
```
normanmaurer: also seems not related to this PR

openssl-dynamic/src/main/c/ssl.c
Scottmitch: This PR creates the need to call `setVerify` and ignore the `depth`. Before this PR the depth would always be set so it didn't matter.

normanmaurer: ah got it... thanks 👍

openssl-dynamic/src/main/c/sslutils.c
Scottmitch: This PR creates the need to call `setVerify` and ignore the `depth` ... a change in behavior that wasn't necessary/supported before.

Scottmitch:
master (df8f3b1)

PR #238, opened 2017-02-20
Correctly handle dh key generation when using libressl

normanmaurer:
Motivation:

Commit f070b9167ac6e2dced746193838000dd6e99a122 changed an ifdef when using libressl to enable generation of temporary dh keys. Unfortunally I missed to also update another ifdef and so SSL.initialize() fails when using libressl now.

Modifications:

Update other ifdef as well.

Result:

netty-tcnative is usable again with libressl.


normanmaurer:
@Scottmitch @nmittler FYI... shame on me :(

normanmaurer:
Cherry-picked into master (7083304)

PR #239, opened 2017-02-27
Move statically referenced JNI methods into independent java class

Scottmitch:
Motivation:
Referencing a JNI method in a static member variable can lead to cyclic class loading issues. This can prevent the library from loading if the JNI code attempts to find a class which contains a static member that is initialized from the result of a JNI method call.

Modifications:
- Move JNI methods that are used to initialize a static member variable into an independent class

Result:
No cyclic class loading error.


Scottmitch:
this will be more likely to occur when we move tcnative into the Netty project and support shading.

openssl-dynamic/src/main/java/io/netty/internal/tcnative/NativeStaticallyReferencedJniMethods.java
```diff
@@ -0,0 +1,130 @@
...
+/**
+ * This class is necessary to break the following cyclic dependency:
+ * <ol>
+ * <li>JNI_OnLoad</li>
+ * <li>JNI Calls FindClass because RegisterNatives (used to register JNI methods) requires a class</li>
+ * <li>FindClass loads the class, but static members variables of that class attempt to call a JNI method which has not
+ * yet been registered.</li>
+ * <li>java.lang.UnsatisfiedLinkError is thrown because native method has not yet been registered.</li>
+ * </ol>
+ * Static members which call JNI methods must not be declared in this class!
```
normanmaurer: nit: enclose with `<strong> </strong>`

openssl-dynamic/src/main/java/io/netty/internal/tcnative/NativeStaticallyReferencedJniMethods.java
```diff
@@ -0,0 +1,130 @@
+package io.netty.internal.tcnative;
```
normanmaurer: missing license header

vs2010.vcxproj.static.template
```diff
@@ -189,6 +189,7 @@
     <ClCompile Include=".\src\bb.c" />
     <ClCompile Include=".\src\error.c" />
     <ClCompile Include=".\src\jnilib.c" />
+    <ClCompile Include=".\src\nativestaticallyreferencedjnimethods.c" />
```
normanmaurer: long filenames FTW ;)

openssl-dynamic/src/main/c/nativestaticallyreferencedjnimethods.c
```diff
@@ -0,0 +1,482 @@
...
+#endif
+}
+
+TCN_IMPLEMENT_CALL(jint, NativeStaticallyReferencedJniMethods, x509vErrDaneNoMatch)(TCN_STDARGS) {
+#ifdef X509_V_ERR_DANE_NO_MATCH
+ return X509_V_ERR_DANE_NO_MATCH;
+#else
+ return TCN_X509_V_ERR_UNSPECIFIED;
+#endif
+}
```
normanmaurer: Add EOL

openssl-dynamic/src/main/c/nativestaticallyreferencedjnimethods.c
```diff
@@ -0,0 +1,482 @@
...
+
+TCN_IMPLEMENT_CALL(jint, NativeStaticallyReferencedJniMethods, x509vErrDaneNoMatch)(TCN_STDARGS) {
+#ifdef X509_V_ERR_DANE_NO_MATCH
+ return X509_V_ERR_DANE_NO_MATCH;
+#else
+ return TCN_X509_V_ERR_UNSPECIFIED;
+#endif
```
normanmaurer: I think the formatting is not correct. We use 4 spaces (not 1).

openssl-dynamic/src/main/java/io/netty/internal/tcnative/NativeStaticallyReferencedJniMethods.java
```diff
@@ -0,0 +1,130 @@
...
+
+/**
+ * This class is necessary to break the following cyclic dependency:
+ * <ol>
+ * <li>JNI_OnLoad</li>
+ * <li>JNI Calls FindClass because RegisterNatives (used to register JNI methods) requires a class</li>
+ * <li>FindClass loads the class, but static members variables of that class attempt to call a JNI method which has not
+ * yet been registered.</li>
+ * <li>java.lang.UnsatisfiedLinkError is thrown because native method has not yet been registered.</li>
```
normanmaurer: {@link java.lang.UnsatisfiedLinkError}

openssl-dynamic/src/main/c/nativestaticallyreferencedjnimethods.c
Scottmitch: fixed

openssl-dynamic/src/main/c/nativestaticallyreferencedjnimethods.c
Scottmitch: done

openssl-dynamic/src/main/java/io/netty/internal/tcnative/NativeStaticallyReferencedJniMethods.java
Scottmitch: done

openssl-dynamic/src/main/java/io/netty/internal/tcnative/NativeStaticallyReferencedJniMethods.java
Scottmitch: done

vs2010.vcxproj.static.template
Scottmitch: this is consistent with transport-native-epoll ... and its an internal class so no-one will know (shhh) :)

Scottmitch:
@normanmaurer - Comments addressed. PTAL.

openssl-dynamic/src/main/c/nativestaticallyreferencedjnimethods.c
```diff
@@ -0,0 +1,482 @@
+/*
```
nmittler: nit: name of this file ... should it be "_" separated?

nmittler: In fact, would it be better to call this something like `native_constants.c`? `nativestaticallyreferencedjnimethods` is a bit of a mouthful :).

openssl-dynamic/src/main/java/io/netty/internal/tcnative/NativeStaticallyReferencedJniMethods.java
```diff
@@ -0,0 +1,130 @@
...
+ * </ol>
+ * Static members which call JNI methods must not be declared in this class!
+ */
+final class NativeStaticallyReferencedJniMethods {
```
nmittler: Same comment regarding the name ... what about `NativeConstants`?

Scottmitch: I don't have a strong preference either way but just to explain the rational for the existing name:

- To clarify that not all constants have to go here (only the statically referenced ones), and that this class exists specifically to avoid class loader issues. So folks should not add things to this class which may introduce a class loader cycle described in the class comments. For example someone may think "I will also add a `static final int MY_VAR = someJNIMethod();` in this class to make the value easier to access from java." which would be wrong.
- To be consistent with transport-native-epoll. The plan is to move tcnative into the Netty project so consistently will help folks understand similar concepts across module boundaries.

Also this class is internal so folks will not be exposed to it.

openssl-dynamic/src/main/java/io/netty/internal/tcnative/NativeStaticallyReferencedJniMethods.java
```diff
@@ -0,0 +1,145 @@
...
+ * <ol>
+ * <li>JNI_OnLoad</li>
+ * <li>JNI Calls FindClass because RegisterNatives (used to register JNI methods) requires a class</li>
+ * <li>FindClass loads the class, but static members variables of that class attempt to call a JNI method which has not
+ * yet been registered.</li>
+ * <li>{@link java.lang.UnsatisfiedLinkError} is thrown because native method has not yet been registered.</li>
+ * </ol>
+ * <strong>Static members which call JNI methods must not be declared in this class!</strong>
+ */
+final class NativeStaticallyReferencedJniMethods {
```
Scottmitch: this comment was collapsed but no action has yet been taken:

from @nmittler https://github.com/netty/netty-tcnative/pull/239#discussion_r103495559

> Same comment regarding the name ... what about `NativeConstants`?

from me:

> I don't have a strong preference either way but just to explain the rational for the existing name:
> - To clarify that not all constants have to go here (only the statically referenced ones), and that this class exists specifically to avoid class loader issues. So folks should not add things to this class which may introduce a class loader cycle described in the class comments. For example someone may think "I will also add a `static final int MY_VAR = someJNIMethod();` in this class to make the value easier to access from java." which would be wrong.
> - To be consistent with transport-native-epoll. The plan is to move tcnative into the Netty project so consistently will help folks understand similar concepts across module boundaries.
> Also this class is internal so folks will not be exposed to it.

Scottmitch: @normanmaurer @nmittler - WDYT?

normanmaurer: I dont care to much as well as its internal anyway. So whatever works for you works for me 😆

Scottmitch:
master (2773e2a)

I will just merge this for now then. We can always change/clarify in a followup PR.

normanmaurer:
@Scottmitch should I cut final?

PR #240, opened 2017-03-01
Allow to use -Dstrip.skip=true to disable stripping of boringssl stat…

normanmaurer:
…ic lib

Motivation:

Sometimes it makes sense to build the library and ensure its not stripped.

Modifications:

Allow to use -Dstrip.skip=true to skip stripping.

Result:

Easier to build artifact that is not stripped.


normanmaurer:
Cherry-picked into master (0ddbeef0f70dde0cd01b0219c1adaa300796ff10)

PR #242, opened 2017-03-03
Static OpenSSL Configuration Options

Scottmitch:
Motivation:
Some configuration options maybe enabled by default which are dangerous or not supported by Netty. We should be more explicit with configuration options of the static OpenSSL build.

Modifications:
- disable sslv2, sslv3, compression, and heartbeats

Result:
Statically built OpenSSL leaner and excludes stuff we don't need.


normanmaurer:
Cherry-picked into master (3e10806c98033a3b754e28fb69ca2dd4a0cfe1f1)

PR #245, opened 2017-03-07
JNI should load NativeStaticallyReferencedJniMethods before other cla…

Scottmitch:
…sses

Motivation:
The point of NativeStaticallyReferencedJniMethods is to make JNI values usable in a static context in Java and not worry about cyclic class loading issues. However if we don't explicitly load the NativeStaticallyReferencedJniMethods class from JNI there is a chance another class which statically references a value in NativeStaticallyReferencedJniMethods is loaded first, and creates a circular class loading issue.

Modifications:
- JNI OnLoad should ensure NativeStaticallyReferencedJniMethods is loaded before other tcnative classes

Result:
Methods from NativeStaticallyReferencedJniMethods can be used in a static context from Java without cyclic class loading issues.


Scottmitch:
master (c1946a8)

PR #247, opened 2017-03-12
Add method to be able to hook in SNIMatchers that exists since java8

normanmaurer:
Motivation:

Java8 supports the concept of SNIMatcher, we should allow to hook these into tcnative.

Modifications:

Add new method and interface to allow using SNIMatcher in netty when using java8+

Result:

More inline with features provided by java8+

### DIFF
--- a/openssl-dynamic/src/main/c/sslcontext.c
+++ b/openssl-dynamic/src/main/c/sslcontext.c
@@ -62,6 +62,13 @@ static apr_status_t ssl_context_cleanup(void *data)
         }
         c->cert_requested_callback_method = NULL;
 
+        if (c->sni_hostname_matcher != NULL) {
+            tcn_get_java_env(&e);
+            (*e)->DeleteGlobalRef(e, c->sni_hostname_matcher);
+            c->sni_hostname_matcher = NULL;
+        }
+        c->sni_hostname_matcher_method = NULL;
+
         if (c->next_proto_data != NULL) {
             free(c->next_proto_data);
             c->next_proto_data = NULL;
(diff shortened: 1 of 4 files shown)


openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -1480,6 +1487,49 @@ TCN_IMPLEMENT_CALL(void, SSLContext, setCertRequestedCallback)(TCN_STDARGS, jlon
...
+
+TCN_IMPLEMENT_CALL(void, SSLContext, setSniHostnameMatcher)(TCN_STDARGS, jlong ctx, jobject matcher)
+{
+    tcn_ssl_ctxt_t *c = J2P(ctx, tcn_ssl_ctxt_t *);
+
+    UNREFERENCED(o);
+    TCN_ASSERT(ctx != 0);
+
+    if (matcher == NULL) {
+        SSL_CTX_set_tlsext_servername_callback(c->ctx, NULL);
```
nmittler: Not a big deal, but maybe we should just always do this first? ... and then check `if (matcher != NULL)` for the rest? That way if `method == null` below, these will already have been properly cleared.

WDYT?

normanmaurer: @nmittler so you say we should always just set to NULL first and then maybe set to something not null if needed later in the method ?

nmittler: @normanmaurer yeah that was the thought ... WDYT?

normanmaurer: not sure this is really needed. I think I would prefer to keep it like it is...

@Scottmitch WDYT ?

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -1480,6 +1487,49 @@ TCN_IMPLEMENT_CALL(void, SSLContext, setCertRequestedCallback)(TCN_STDARGS, jlon
...
+        jclass matcher_class = (*e)->GetObjectClass(e, matcher);
+        jmethodID method = (*e)->GetMethodID(e, matcher_class, "match", "(JLjava/lang/String;)Z");
+        if (method == NULL) {
+            return;
+        }
+        // Delete the reference to the previous specified matcher if needed.
+        if (c->sni_hostname_matcher != NULL) {
+            (*e)->DeleteLocalRef(e, c->sni_hostname_matcher);
+        }
+        c->sni_hostname_matcher = (*e)->NewGlobalRef(e, matcher);
```
rkapsi: Don't know if it's applicable but maybe `NewWeakGlobalRef()` instead of `NewGlobalRef`?

```c
c->sni_hostname_matcher = (*e)->NewWeakGlobalRef(e, matcher);
```

```c
jobject matcher = (*e)->NewLocalRef(e, (jweak)c->sni_hostname_matcher);
if (!(*e)->IsSameObject(e, matcher, NULL)) {
  // Do something with the SNIMatcher
} else {
  // It appears the SNIMatcher got GC'd in Java land
}
(*e)->DeleteLocalRef(e, matcher);
```

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -1480,6 +1487,49 @@ TCN_IMPLEMENT_CALL(void, SSLContext, setCertRequestedCallback)(TCN_STDARGS, jlon
...
+        SSL_CTX_set_tlsext_servername_arg(c->ctx, NULL);
+    } else {
+        jclass matcher_class = (*e)->GetObjectClass(e, matcher);
+        jmethodID method = (*e)->GetMethodID(e, matcher_class, "match", "(JLjava/lang/String;)Z");
+        if (method == NULL) {
+            return;
+        }
+        // Delete the reference to the previous specified matcher if needed.
+        if (c->sni_hostname_matcher != NULL) {
+            (*e)->DeleteLocalRef(e, c->sni_hostname_matcher);
```
rkapsi: I think this is a bug, shouldn't it be `DeleteGlobalRef`?

normanmaurer: You are right... doh!

openssl-dynamic/src/main/c/sslcontext.c
normanmaurer: I think a a GlobalRef is ok

normanmaurer:
@Scottmitch PTAL

normanmaurer:
@trustin PTAL as well

openssl-dynamic/src/main/c/sslcontext.c
Scottmitch: I'm fine with only setting the value when necessary.

However shouldn't we always check if we need to free memory?

```java
if (c->sni_hostname_matcher != NULL) {
  (*e)->DeleteGlobalRef(e, c->sni_hostname_matcher);
}
```

normanmaurer: +1 let me fix this

normanmaurer:
Cherry-picked into master (563307abc5955194d93d5f9c71caa1c5a3346388)

Scottmitch:
See PR https://github.com/netty/netty-tcnative/pull/262 for followup

PR #248, opened 2017-03-12
Use OPENSSL_malloc and OPENSSL_free

normanmaurer:
Motivation:

We used malloc and free directly in a few places while in others OPENSSL_malloc / OPENSSL_free are used.

Modifications:

Replace malloc calls with OPENSSL_malloc
Replace free calls with OPENSSL_free

Result:

More consistent way of allocating and freeing memory.

### DIFF
--- a/openssl-dynamic/src/main/c/sslcontext.c
+++ b/openssl-dynamic/src/main/c/sslcontext.c
@@ -63,13 +63,13 @@ static apr_status_t ssl_context_cleanup(void *data)
         c->cert_requested_callback_method = NULL;
 
         if (c->next_proto_data != NULL) {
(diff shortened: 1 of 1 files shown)


normanmaurer:
@Scottmitch want to have a look as well ?

normanmaurer:
Cherry-picked into master (abded87)

normanmaurer:
I just wanted to make it consistent and as you stated it not hurts

Scottmitch:
It is a bit misleading on first glance. Typically `OPENSSL_malloc` and `OPENSSL_free` are used when it is expected that OpenSSL will eventually free the memory, but this is not the case here.

PR #250, opened 2017-03-15
Fix memory leak when a previous set verifier / cert callback is replaced

normanmaurer:
Motivation:

When a previous set verifier / cert callback is replaced we need to ensure the previous created global reference is deleted as well. Due a bug we not did this.

Modifications:

Replace DeleteLocalRef with DeleteGlobalRef

Result:

No more native leak when replacing verifier / cert callback


normanmaurer:
Doh! Found this while review another comment of @rkapsi

normanmaurer:
Cherry-picked into master (4f63c90cbca4850641c90937493cab93f1557a7e)

PR #251, opened 2017-03-16
Simplify loading of native library

normanmaurer:
Motivation:

We had some "complex" native code left that checks version / freatues when trying to load the native library which can be simplified.

Modifications:

- Simplify loading code

Result:

Simplier code to maintain.


normanmaurer:
Cherry-picked into master (79e2dea)

Issue #152, opened 2016-06-08
Release musl compatible binaries


pires:
For the ones using Alpine Linux, it would be great to be have `musl` compatible binaries.

ejona86:
We didn't confirm that glibc vs musl was the problem, but it seems likely and [only debug logging](https://github.com/netty/netty/blob/4.1/common/src/main/java/io/netty/util/internal/NativeLibraryLoader.java#L164) will provide more information.

Scottmitch:
related to https://github.com/netty/netty-tcnative/issues/111

ejona86:
I confirmed the problem was due to missing glibc. It is a musl vs glibc problem.

```
java.lang.UnsatisfiedLinkError: /tmp/libnetty-tcnative-linux-x86_641722850980897813599.so: Error loading shared library libcrypt.so.1: No such file or directory (needed by /tmp/libnetty-tcnative-linux-x86_641722850980897813599.so)
```

And then after a hack:
```
java.lang.UnsatisfiedLinkError: /tmp/libnetty-tcnative-linux-x86_64410176636862150824.so: Error relocating /tmp/libnetty-tcnative-linux-x86_64410176636862150824.so: pthread_mutex_consistent_np: symbol not found
```

Compiling on Alpine would probably be necessary.

pires:
@dimitarg thanks to your work I've [automated the build of binaries for Alpine](https://github.com/pires/netty-tcnative-alpine) with Docker.
To whomever needs this, feel free to use and abuse.

@ejona86 thank you very much for your support. Without you I'd take ages to get this going, if ever.

ejona86:
And for posterity and those not familiar with java logging, this is the config I used to get debug log output (named `log.properties`):
```
handlers = java.util.logging.ConsoleHandler
java.util.logging.ConsoleHandler.level = ALL
io.netty.util.internal.NativeLibraryLoader.level = ALL
```

And then specified `-Djava.util.logging.config.file=log.properties` when running java.

nmittler:
@ejona86 would you mind adding that to the tcnative wiki?

normanmaurer:
@ejona86 ^^

anuraaga:
Sorry to spam - came across this issue while looking up info related to alpine linux, and wondering if it could be closed as a duplicate of #111? They seem to be the same issue, and #111 has progressed further it seems.

Issue #111, opened 2016-02-14
Unable to use pre-built binaries on alpine linux


ankurcha:
I am trying to use the netty-tcnative libraries for my grpc application in a docker container. When i use the prebuilt binary I get the following error:

``` bash
Inconsistency detected by ld.so: dl-lookup.c: 128: check_match: Assertion `version->filename == ((void *)0) || ! _dl_name_match_p (version->filename, map)' failed!
```

normanmaurer:
we release versions build on centos and debian. I guess you will need to build it by your own for alpine.

dimitarg:
Assuming an alpine docker image which has an oracle jdk 8 installed, and netty-tcnative available under /opt/data/netty-tcnative (perhaps through a shared volume in the host, if you don't want to bother installing git on the alpine image), here is a setup which let me build netty-tcnative successfully:

```
MAVEN_VERSION=3.3.3 \
 && cd /usr/share \
 && wget http://archive.apache.org/dist/maven/maven-3/$MAVEN_VERSION/binaries/apache-maven-$MAVEN_VERSION-bin.tar.gz -O - | tar xzf - \
 && mv /usr/share/apache-maven-$MAVEN_VERSION /usr/share/maven \
 && ln -s /usr/share/maven/bin/mvn /usr/bin/mvn


export MAVEN_HOME=/usr/share/maven

cd /opt/data/netty-tcnative/

apk update

apk add autoconf

apk add automake

apk add libtool 

apk add alpine-sdk

apk add apr-util

apk add apr-util-dev

apk add build-base gcc abuild binutils binutils-doc gcc-doc

apk add cmake cmake-doc

apk add ninja

apk add go
```

, and then 

```mvn clean install```


@ankurcha and anyone else interested, I will let you know if the end result actually works. My usecase though is netty http2 server with openssl and epoll, I don't know if that would show that grpc works. We use netty directly though, no grpc involved in the test.

**Comment by dimitarg** [comment]
The built native jar as desribed above, under openssl-dynamic/target/netty-tcnative-<version>-linux-x86_64.jar , works for me, if I use the following, in a runtime image based FROM jeanblanchard/java:jdk-8:

```
run 'apk update'
run 'apk add apr-util apr-util-dev libuuid'
```

It's very important to change the base image only if you know what you are doing otherwise you might end up not finding libuuid during tcnative load.

Hope this helps. Hopefully I get the time to provide this in a matter more automated than github comments someday.

Cheers, Dimitar

[...]

PR #252, opened 2017-03-19
[WIP] Add a gradle configuration that builds tcnative linux versions within…

anuraaga:
… docker while also executing the host build.

Motivation: Currently tcnative does not run on Alpine linux due to libc incompatibility. By having more flexible building, taking advantage of docker, it should be possible to provide official binaries without much extra work during release.

See also #111, #152 

I went ahead and hacked this setup together to allow more flexibility in tcnative builds. Without more knowledge of the release process, I can't say this does exactly what's needed but will be happy to continue to update until it can be used for releasing. FWIW, ~/.m2/repository gets populated with the right jars so it should be mostly there :)

This PR does not change the tcnative build to gradle - that would be a much more invasive process. It uses gradle to take advantage of its flexibility in task execution - looking at examples of maven-docker-plugin made it look like it'd result in an unreadable pom.xml.

With this setup, it would be possible to build all of OSX (or windows which also has docker) and linux versions on the same machine. Care has been taken to allow the versions to be built in parallel using gradle's incubating parallel projects feature. On my macbook pro, all the builds end in about 20min.

```
# This builds three versions of tcnative in parallel
./gradlew buildAll
```

Another feature of this gradle script is it automatically downloads and runs maven. So for building linux binaries, the requirements are literally just Java and Docker.

Main questions I currently have are
1) The top-level pom restricts releasing to RHEL 6.8. How do the other versions get released?
2) RHEL 6.8 will produce openssl-dynamic with the fedora classifier since it's fedora-like. What distribution is used to build the non-fedora-like version? It'd be easy to add it here too :)


anuraaga:
@normanmaurer Hope this looks useful, PTAL whenever you get a chance :)
/cc @trustin @pires

anuraaga:
@normanmaurer Hello - just checking in, does this generally seem reasonable? The hope is to make releasing easier for the maintainers by reducing the needed VMs :)

normanmaurer:
@anuraaga sorry I had no time yet and just returned back from the US. I will have a look this week.

anuraaga:
@normanmaurer Hey - long time no check-in. Is there any future for this or official alpine linux support in tcnative? If not, I should just go ahead and close it I think.

PR #253, opened 2017-03-21
Remove the apr pool from the context

normanmaurer:
Motiviation:

We created and stored an apr pool in the context that we use per SSLContext. This pool is not used anymore and so not needed.

Modification:

Remove the pool creation and remove it from the struct

Result:

Less memory used for the struct and less resource creation.


normanmaurer:
Cherry-picked into master (33caa6b)

PR #254, opened 2017-03-22
Fix exported packages when using OSGi with boringssl

normanmaurer:
Motivation:

We missed to adjust the exported pckages in boringgsl-static.

Modifications:

Fix package name

Result:

Correctly export packages when using OSGI.


normanmaurer:
Cherry-picked into master (fed2c76318bd092d091ef499368bdaf6f6d657dd)

PR #255, opened 2017-03-23
Remove dependency on apr by using c++11 / c++14 features

normanmaurer:
Motivation:

We had some limited use of apr left in our tcnative fork which complicated the deployment. This apr usages can be replaced by using c++.

Modification:

Remove the dependency on apr by using c++11 / c++14 features

Result:

No apr dependeny


normanmaurer:
not 100 % done yet. Will ask for review once I am done

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -732,43 +699,33 @@ static void ssl_dyn_lock_function(int mode, struct CRYPTO_dynlock_value *l,
...
-     * convenience.
-     */
-    dynlockpool = p;
     CRYPTO_set_dynlock_create_callback(ssl_dyn_create_function);
     CRYPTO_set_dynlock_lock_callback(ssl_dyn_lock_function);
     CRYPTO_set_dynlock_destroy_callback(ssl_dyn_destroy_function);
 
-    apr_pool_cleanup_register(p, NULL, ssl_thread_cleanup,
-                              apr_pool_cleanup_null);
+    //apr_pool_cleanup_register(p, NULL, ssl_thread_cleanup,
```
nmittler: remove?

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -864,18 +817,18 @@ TCN_IMPLEMENT_CALL(jint, SSL, initialize)(TCN_STDARGS, jstring engine)
...
+        tcn_ThrowException(e, "Unable to init SSL");
+        return -1;
     }
     /*
      * Let us cleanup the ssl library when the library is unloaded
      */
-    apr_pool_cleanup_register(tcn_global_pool, NULL,
-                              ssl_init_cleanup,
-                              apr_pool_cleanup_null);
+    //apr_pool_cleanup_register(tcn_global_pool, NULL,
```
nmittler: remove?

openssl-dynamic/src/main/c/sslutils.c
```diff
@@ -30,8 +30,8 @@
  */
 
 #include "tcn.h"
-
 #include "ssl_private.h"
+#include <string.h>
```
nmittler: Is this needed? Was it previously brought in by another header?

normanmaurer: yep was brought it by another header before.

openssl-dynamic/src/main/c/ssl.c
normanmaurer: Looking for a good way to call these again. So stay tuned

openssl-dynamic/src/main/c/ssl.c
normanmaurer: Looking for a good way to call these again. So stay tuned

normanmaurer:
@nmittler @Scottmitch PTAL... it compiles here and on my linux / windows vm. Need to update the g++ version on the ci.

normanmaurer:
@trustin PTAL as well.

openssl-dynamic/src/main/c/tcn_atomic.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include "tcn_atomic.h"
+#include <atomic>
+
+tcn_atomic_uint32_t tcn_atomic_uint32_new() {
+    auto *atomic = new std::atomic<uint32_t>(0);
+    return (tcn_atomic_uint32_t) atomic;
+}
+
+void tcn_atomic_uint32_free(tcn_atomic_uint32_t atomic) {
+    auto *p = (std::atomic<uint32_t> *) atomic;
```
Scottmitch: consider removing the temp variable.

also consider passing a `**` as a parameter so you can always set `*atomic = NULL;` in these methods.

Scottmitch: this comment applies to all the `cpp` files.

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -222,16 +229,20 @@ TCN_IMPLEMENT_CALL(jlong, SSLContext, make)(TCN_STDARGS, jint protocol, jint mod
         goto cleanup;
     }
 
-    TCN_THROW_IF_ERR(apr_pool_create(&p, tcn_global_pool), p);
-
-    if ((c = apr_pcalloc(p, sizeof(tcn_ssl_ctxt_t))) == NULL) {
-        tcn_ThrowAPRException(e, apr_get_os_error());
+    if ((c = OPENSSL_malloc(sizeof(tcn_ssl_ctxt_t))) == NULL) {
```
Scottmitch: consider just using `calloc` instead of `OPENSSL_malloc`. IIUC OpenSSL is not aware of this variable and therefore we don't need to use OpenSSL allocation.

Scottmitch: also make sure you use `free` instead of `OPENSSL_free`

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -317,7 +328,10 @@ TCN_IMPLEMENT_CALL(jint, SSLContext, free)(TCN_STDARGS, jlong ctx)
     UNREFERENCED_STDARGS;
     TCN_ASSERT(ctx != 0);
 
-    return ssl_context_cleanup(c);
+    ssl_context_cleanup(c);
+
+    // TODO: Fix me
```
Scottmitch: can we fix this now ... just make this function void?

openssl-dynamic/src/main/c/tcn_lock_rw.cpp
```diff
@@ -0,0 +1,49 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+#include "tcn_lock_rw.h"
+#include <shared_mutex>
+
+tcn_lock_rw_t tcn_lock_rw_new() {
+    auto *m = new std::shared_timed_mutex();
```
Scottmitch: add a `TODO` that we should use `std::shared_mutex `when we are able to use c++17 (we don't use the timed component).

openssl-dynamic/src/main/c/tcn_lock_rw.cpp
```diff
@@ -0,0 +1,49 @@
...
+}
+
+void tcn_lock_rw_free(tcn_lock_rw_t lock) {
+    auto *m = (std::shared_timed_mutex *) lock;
+    delete m;
+}
+
+tcn_lock_writer_t tcn_lock_writer_acquire(tcn_lock_rw_t lock) {
+    auto *m = (std::shared_timed_mutex *) lock;
+    auto *writer_lock = new std::lock_guard<std::shared_timed_mutex>(*m);
```
Scottmitch: Did you consider using `std::unique_lock`? It is explicitly called out in the `std::shared_lock` docs (see below). Also the default behavior seems to be the same (constructor locks and destructor unlocks).

http://en.cppreference.com/w/cpp/thread/shared_lock
> Locking a shared_lock locks the associated shared mutex in shared mode (to lock it in exclusive mode, std::unique_lock can be used)

openssl-dynamic/src/main/c/tcn_thread.h
```diff
@@ -0,0 +1,30 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+#ifndef TCN_THREAD_H
+#define TCN_THREAD_H
+
+#include <stddef.h>
```
Scottmitch: is this necessary?

openssl-dynamic/src/main/java/io/netty/internal/tcnative/Library.java
```diff
@@ -119,16 +107,7 @@ public static boolean initialize() throws Exception {
...
-                throw new UnsatisfiedLinkError("Unsupported APR Version (" +
-                                               aprVersionString() + ")");
-            }
-
-            if (!aprHasThreads()) {
-                throw new UnsatisfiedLinkError("Missing APR_HAS_THREADS");
-            }
         }
-        return initialize0() && SSL.initialize(engine) == 0;
+        return SSL.initialize(engine) == 0;
```
Scottmitch: consider just changing `SSL.initialize` to a boolean which returns true if it has already been initialized or this call successfully initialized the `SSL` module.

openssl-dynamic/src/main/native-package/configure.ac
```diff
@@ -29,12 +29,14 @@ AC_CANONICAL_HOST
 AC_CANONICAL_SYSTEM
 
 ${CFLAGS="-O3 -Werror -fno-omit-frame-pointer -Wunused-variable"}
-${CXXFLAGS="-O3 -Werror -fno-omit-frame-pointer -Wunused-variable"}
+# Ensure we support c++11 for atomics etc
```
Scottmitch: `c++11` -> `c++14`

openssl-dynamic/src/main/c/tcn_lock_rw.h
```diff
@@ -0,0 +1,42 @@
...
+ * under the License.
+ */
+#ifndef TCN_LOCK_RW_H
+#define TCN_LOCK_RW_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif // __cplusplus
+
+typedef void* tcn_lock_rw_t;
```
Scottmitch: we have `lock_rw`, `lock_reader`, and `lock_writer`. Consider being consistent/complete with the naming:

- `lock_reader_writer`, `lock_reader`, and `lock_writer`
- `lock_rw`, `lock_r`, and `lock_w`

Also consider making the file name consistent if you go with the more verbose names.

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -676,39 +649,20 @@ static struct CRYPTO_dynlock_value *ssl_dyn_create_function(const char *file,
...
-     * away in the destruction callback.
-     */
-    rv = apr_pool_create(&p, dynlockpool);
-    if (rv != APR_SUCCESS) {
-        /* TODO log that fprintf(stderr, "Failed to create subpool for dynamic lock"); */
-        return NULL;
-    }
-
-    value = (struct CRYPTO_dynlock_value *)apr_palloc(p,
+    value = (struct CRYPTO_dynlock_value *)OPENSSL_malloc(
```
Scottmitch: consider declaring and initializing this variable in the same statement.

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -653,20 +633,13 @@ static void ssl_set_thread_id(CRYPTO_THREADID *id)
     CRYPTO_THREADID_set_numeric(id, ssl_thread_id());
 }
 
-static apr_status_t ssl_thread_cleanup(void *data)
+static void ssl_thread_cleanup()
 {
-    UNREFERENCED(data);
     CRYPTO_set_locking_callback(NULL);
     CRYPTO_THREADID_set_callback(NULL);
     CRYPTO_set_dynlock_create_callback(NULL);
```
Scottmitch: it doesn't look like this method takes ownership of the memory [1]. We should take more care to free this memory, and if we set it check if there already exists something and free it.

[1] https://github.com/openssl/openssl/blob/OpenSSL_1_0_2-stable/crypto/cryptlib.c#L376

Scottmitch: also be sure to use the `ssl_dyn_destroy_function` (should be accessible via `CRYPTO_get_dynlock_destroy_callback`) on the result of `CRYPTO_get_dynlock_create_callback`.

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -732,43 +686,30 @@ static void ssl_dyn_lock_function(int mode, struct CRYPTO_dynlock_value *l,
 static void ssl_dyn_destroy_function(struct CRYPTO_dynlock_value *l,
                           const char *file, int line)
 {
-    apr_status_t rv;
-    rv = apr_thread_mutex_destroy(l->mutex);
-    if (rv != APR_SUCCESS) {
-        /* TODO log that fprintf(stderr, "Failed to destroy mutex for dynamic lock %s:%d", l->file, l->line); */
-    }
+    free((void *) l->file);
```
Scottmitch: is the cast necessary? can we just make the type `char*` in the structure definition if the `const` is confusing things?

normanmaurer: let me see if this works.

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1987,3 +1918,8 @@ TCN_IMPLEMENT_CALL(void, SSL, freeX509Chain)(TCN_STDARGS, jlong x509Chain)
     sk_X509_pop_free(chain, X509_free);
 }
 
+void tcn_ssl_unload() {
+    if (ssl_init_cleanup()) {
```
normanmaurer: sure can do

openssl-dynamic/src/main/c/tcn_atomic.cpp
normanmaurer: good idea

openssl-dynamic/src/main/c/sslcontext.c
normanmaurer: sounds good... this way we can also remove the memset

openssl-dynamic/src/main/c/tcn_lock_rw.cpp
normanmaurer: nope... let me see what is the difference

openssl-dynamic/src/main/c/tcn_thread.h
normanmaurer: yes because of size_t

openssl-dynamic/src/main/java/io/netty/internal/tcnative/Library.java
normanmaurer: I wanted to do this without the need to break the API to make it easier to consume for now.

openssl-dynamic/src/main/native-package/configure.ac
normanmaurer: +1

openssl-dynamic/src/main/c/tcn_lock_rw.h
normanmaurer: +1

openssl-dynamic/src/main/c/ssl.c
normanmaurer: hmm why would this be needed if we not malloc the callbacks before?

openssl-dynamic/src/main/c/sslcontext.c
normanmaurer: like said before I wanted to do the pr without api breakage to make it easier to consume for now.

openssl-dynamic/src/main/c/tcn_lock_rw.cpp
normanmaurer: yep

openssl-dynamic/src/main/c/ssl.c
Scottmitch: can you clarify? previously a `apr_palloc` was used to allocate the `CRYPTO_dynlock_value` structure. This PR uses `OPENSSL_malloc` instead of the `apr_palloc` method. Either way there is dynamic memory being allocated. I'm not sure where/if it was freed before, but unless you find that OpenSSL is taking ownership in this situation (just setting the variable to `NULL`) then we need to ensure this dynamic memory is cleaned up before we lose the reference to it.

openssl-dynamic/src/main/c/sslcontext.c
Scottmitch: Doing a followup PR is fine but the return value is not used in Netty. I don't think this will break anything.

normanmaurer:
Doh brain fart ... you are right let me fix it

openssl-dynamic/src/main/java/io/netty/internal/tcnative/Library.java
Scottmitch: This method is package private, and therefore not used externally. I don't think this will break anything.

normanmaurer:
Good point. Let me change it

openssl-dynamic/src/main/c/tcn_lock_rw.cpp
normanmaurer: You are talking about replacing `lock_guard` here correct ?

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -732,69 +624,100 @@ static void ssl_dyn_lock_function(int mode, struct CRYPTO_dynlock_value *l,
...
+            || !ENGINE_ctrl_cmd_string(e, "LOAD", NULL, 0)) {
+            ENGINE_free(e);
+            e = NULL;
+        }
+    }
+    return e;
 }
+#endif
 
 TCN_IMPLEMENT_CALL(jint, SSL, initialize)(TCN_STDARGS, jstring engine)
```
Scottmitch: https://github.com/netty/netty-tcnative/pull/255#discussion_r108472933

it looks like this can just be void because we throw an exception upon failure anyways.

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -2113,3 +2043,7 @@ TCN_IMPLEMENT_CALL(jbyteArray, SSL, getOcspResponse)(TCN_STDARGS, jlong ssl) {
     return value;
 #endif
 }
+
+void tcn_ssl_unload() {
+    ssl_init_cleanup();
```
Scottmitch: can you combine `tcn_ssl_unload` and `ssl_init_cleanup`? seems unnecessary to have a function which only calls a single function?

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -87,14 +84,19 @@ static apr_status_t ssl_context_cleanup(void *data)
             free(c->password);
             c->password = NULL;
         }
+
+        tcn_atomic_uint32_free(&(c->ticket_keys_new));
```
Scottmitch: `&(c->ticket_keys_new)` -> `&c->ticket_keys_new` ? The parens shouldn't be necessary, right?

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -312,24 +316,24 @@ TCN_IMPLEMENT_CALL(jlong, SSLContext, make)(TCN_STDARGS, jint protocol, jint mod
...
 
-    apr_thread_rwlock_create(&c->mutex, p);
+    c->ticket_keys_lock = tcn_lock_rw_new();
 
     return P2J(c);
 cleanup:
-    if (p != NULL) {
-        apr_pool_destroy(p);
+    if (c != NULL) {
+        OPENSSL_free(c);
```
Scottmitch: `OPENSSL_free(c)` -> `free(c)`

openssl-dynamic/src/main/c/tcn_atomic.cpp
```diff
@@ -0,0 +1,35 @@
...
+    return (tcn_atomic_uint32_t) new std::atomic<uint32_t>(0);
+}
+
+void tcn_atomic_uint32_free(tcn_atomic_uint32_t* atomic) {
+    delete (std::atomic<uint32_t> *) *atomic;
+    *atomic = NULL;
+}
+
+uint32_t tcn_atomic_uint32_get(tcn_atomic_uint32_t atomic) {
+    return  *((std::atomic<uint32_t> *) atomic);
```
Scottmitch: nit: remove double space after `return`

openssl-dynamic/src/main/c/tcn_lock.cpp
```diff
@@ -0,0 +1,34 @@
...
+    return (tcn_lock_t) new std::mutex();
+}
+
+void tcn_lock_free(tcn_lock_t* lock) {
+    delete (std::mutex *) *lock;
+    *lock = NULL;
+}
+
+void tcn_lock_acquire(tcn_lock_t lock) {
+    (* (std::mutex *) lock).lock();
```
Scottmitch: consider simplifying to:
```c
((std::mutex*) lock)->lock();
```

openssl-dynamic/src/main/c/tcn_lock.cpp
```diff
@@ -0,0 +1,34 @@
...
+}
+
+void tcn_lock_release(tcn_lock_t lock) {
+    (* (std::mutex *) lock).unlock();
```
Scottmitch: consider simplifying to:
```c
((std::mutex*) lock)->unlock();
```

openssl-dynamic/src/main/c/tcn_lock.cpp
```diff
@@ -0,0 +1,34 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+#include "tcn_lock.h"
+#include <mutex>
+
+tcn_lock_t tcn_lock_new() {
+    return (tcn_lock_t) new std::mutex();
```
Scottmitch: IIRC in C++ it is ambiguous if you try to use `()` for the no argument constructor without the `new` keyword ... `new` may make the situation unambiguous but consider dropping the `()` which may be more `C++`ish.

http://stackoverflow.com/a/2318731

openssl-dynamic/src/main/c/tcn_lock_rw.cpp
```diff
@@ -0,0 +1,45 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+#include "tcn_lock_rw.h"
+#include <shared_mutex>
+
+tcn_lock_rw_t tcn_lock_rw_new() {
+    // Once we switch to c++17 we should use std::shared_mutex
+    return (tcn_lock_rw_t) new std::shared_timed_mutex();
```
Scottmitch: IIRC in C++ it is ambiguous if you try to use `()` for the no argument constructor without the `new` keyword ... `new` may make the situation unambiguous but consider dropping the `()` which may be more `C++`ish.

http://stackoverflow.com/a/2318731

openssl-dynamic/src/main/c/tcn_lock_rw.cpp
```diff
@@ -0,0 +1,45 @@
...
+}
+
+void tcn_lock_rw_free(tcn_lock_rw_t* lock) {
+    delete (std::shared_timed_mutex *) *lock;
+    *lock = NULL;
+}
+
+tcn_lock_w_t tcn_lock_w_acquire(tcn_lock_rw_t lock) {
+    auto *m = (std::shared_timed_mutex *) lock;
+    return (tcn_lock_w_t) new std::lock_guard<std::shared_timed_mutex>(*m);
```
Scottmitch: https://github.com/netty/netty-tcnative/pull/255#discussion_r108468584

why not `std::unique_lock`?

Scottmitch: also consider removing the temp variable so you don't have to worry about `auto` or types:

```cpp
return (tcn_lock_w_t) new std::lock_guard<std::shared_timed_mutex>(*((std::shared_timed_mutex*) lock));
```

openssl-dynamic/src/main/c/tcn_lock_rw.cpp
```diff
@@ -0,0 +1,45 @@
...
+}
+
+void tcn_lock_w_release(tcn_lock_w_t lock) {
+    delete (std::lock_guard<std::shared_timed_mutex> *) lock;
+}
+
+tcn_lock_r_t tcn_lock_r_acquire(tcn_lock_rw_t lock) {
+    auto *m = (std::shared_timed_mutex *) lock;
```
Scottmitch: consider removing the temp variable so you don't have to worry about auto or types:

```cpp
return (tcn_lock_r_t) new std::shared_lock<std::shared_timed_mutex>(*((std::shared_timed_mutex*) lock));
```

openssl-dynamic/src/main/c/tcn_thread.h
Scottmitch: you may want to be sure you include the cpp headers if you are in cpp, and the c headers if you are in c. They may define types differently or have different behaviors.

```c
#ifdef __cplusplus
#include <cstddef>
extern "C" {
#else
#include <stddef.h>
#endif // __cplusplus
```

openssl-dynamic/src/main/c/ssl.c
Scottmitch: looks like the method return type was changed in the java header, but not here ... seems like there will be a mismatch

openssl-dynamic/src/main/c/tcn_atomic.h
```diff
@@ -0,0 +1,38 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+#ifndef TCN_ATOMIC_H
+#define TCN_ATOMIC_H
+
+#include <stdint.h>
```
Scottmitch: you may want to be sure you include the cpp headers if you are in cpp, and the c headers if you are in c. They may define types differently or have different behaviors.

```c
#ifdef __cplusplus
#include <cstdint>
extern "C" {
#else
#include <stdint.h>
#endif // __cplusplus
```

openssl-dynamic/src/main/c/sslcontext.c
normanmaurer: right

openssl-dynamic/src/main/c/ssl.c
normanmaurer: +1

openssl-dynamic/src/main/c/ssl.c
normanmaurer: yeah... noe done yet.

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -732,69 +624,100 @@ static void ssl_dyn_lock_function(int mode, struct CRYPTO_dynlock_value *l,
...
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
+    ERR_remove_thread_state(NULL);
+#endif
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L && !defined(LIBRESSL_VERSION_NUMBER)
+    free_bio_methods();
+#endif
+
+    CRYPTO_set_locking_callback(NULL);
+    CRYPTO_THREADID_set_callback(NULL);
+    CRYPTO_set_dynlock_create_callback(NULL);
```
normanmaurer: it is not yet, will do today (and it was not freed before in tcnative as well)

normanmaurer: Actually I think all is good now. Please ping me once you are awake and we can talk about it in more detail if you have doubts. Maybe I am just missing something but from my look into the source etc its just fine now

Scottmitch: I didn't track it all the way down but found that there are links to the callbacks retained in the engines ... so `ENGINE_cleanup` *maybe* the place where these callbacks are used to clean stuff up. my concern was we were missing a call which notifies OpenSSL to use these callbacks. @davidben - can you shed some light on this situation?

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -74,7 +71,7 @@ static apr_status_t ssl_context_cleanup(void *data)
         }
         c->alpn_proto_len = 0;
 
-        apr_thread_rwlock_destroy(c->mutex);
+        tcn_lock_rw_free(&(c->ticket_keys_lock));
```
Scottmitch: `&(c->ticket_keys_lock)` -> `&c->ticket_keys_lock`

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -2113,3 +2041,4 @@ TCN_IMPLEMENT_CALL(jbyteArray, SSL, getOcspResponse)(TCN_STDARGS, jlong ssl) {
     return value;
 #endif
 }
+
```
Scottmitch: kill the extra new line

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -106,7 +99,7 @@ struct TCN_bio_bytebuffer {
  * so that the return value for SSL_OP_FUTURE_WORKAROUND will only be
  * reported by versions that actually support that specific workaround.
  */
-static const jint supported_ssl_opts = 0
+static const jlong supported_ssl_opts = 0
```
normanmaurer: +1

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -590,69 +481,100 @@ static void ssl_dyn_lock_function(int mode, struct CRYPTO_dynlock_value *l,
...
         TCN_FREE_CSTRING(engine);
-        return (jint)APR_SUCCESS;
+        return;
     }
 
 #if OPENSSL_VERSION_NUMBER < 0x10100000L
     if (SSLeay() < 0x0090700L) {
         TCN_FREE_CSTRING(engine);
-        tcn_ThrowAPRException(e, APR_EINVAL);
+        tcn_ThrowException(e, "openssl version too old");
```
trustin: How about appending the version number and the minimum required version number?

openssl-dynamic/src/main/native-package/configure.ac
```diff
@@ -29,12 +29,14 @@ AC_CANONICAL_HOST
 AC_CANONICAL_SYSTEM
 
 ${CFLAGS="-O3 -Werror -fno-omit-frame-pointer -Wunused-variable"}
-${CXXFLAGS="-O3 -Werror -fno-omit-frame-pointer -Wunused-variable"}
+# Ensure we support c++14 for atomics etc
+${CXXFLAGS="-O3 -Werror -fno-omit-frame-pointer -Wunused-variable -std=c++14"}
```
trustin: Perhaps should we static-link libgcc and libstdc++? `-static-libgcc -static-libstdc++`
Also, we may eventually want to link against musl or uClibc instead of GLIBC so our binary works on more distros.

normanmaurer: I think we not want to do any static linking in our dynamic build. Not sure about the "-static" ones tho. @Scottmitch @nmittler WDYT ?

trustin: I'm not sure if the situation has changed, but I had an impression that libstdc++ has worse backward ABI compatibility historically. Maybe I'm completely wrong, though.

Scottmitch: Lets handle this as a followup PR if it becomes and issue.

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -675,36 +597,49 @@ TCN_IMPLEMENT_CALL(jint, SSL, initialize)(TCN_STDARGS, jstring engine)
...
+
+    CRYPTO_set_dynlock_create_callback(ssl_dyn_create_function);
+    CRYPTO_set_dynlock_lock_callback(ssl_dyn_lock_function);
+    CRYPTO_set_dynlock_destroy_callback(ssl_dyn_destroy_function);
 
 #ifndef OPENSSL_NO_ENGINE
     if (J2S(engine)) {
         ENGINE *ee = NULL;
-        apr_status_t err = APR_SUCCESS;
+        int err = 0;
```
Scottmitch: just use `bool`?

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -721,19 +656,11 @@ TCN_IMPLEMENT_CALL(jint, SSL, initialize)(TCN_STDARGS, jstring engine)
     if (r) {
```
Scottmitch: (not introduce by this PR, but while we are here) `r` is an integer type ... lets use it like an integer: `r != 0`

openssl-dynamic/src/main/c/tcn_atomic.h
```diff
@@ -0,0 +1,40 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+#ifndef TCN_ATOMIC_H
+#define TCN_ATOMIC_H
+
+
```
Scottmitch: nit: kill line

openssl-dynamic/src/main/c/tcn_thread.h
```diff
@@ -0,0 +1,32 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+#ifndef TCN_THREAD_H
+#define TCN_THREAD_H
+
+
```
Scottmitch: nit: kill line

Scottmitch:
@normanmaurer - are you planning on pulling this in?

normanmaurer:
Yes just wanted to do a few more tests

normanmaurer:
@Scottmitch PTAL again.. rebased on latest master and should are ready to get pulled in.

normanmaurer:
After @Scottmitch mentioned to me that centos 6 glibc may not support c++14 I checked in more detail and found out it even not support c++11 . So I guess we will need to life with apr ... 

@Scottmitch WDYT ?

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -244,17 +245,19 @@ TCN_IMPLEMENT_CALL(jlong, SSLContext, make)(TCN_STDARGS, jint protocol, jint mod
...
+    if ((c = calloc(1, sizeof(tcn_ssl_ctxt_t))) == NULL) {
+        tcn_Throw(e, "Failed to calloc tcn_ssl_ctxt_t");
         goto cleanup;
     }
 
     c->protocol = protocol;
     c->mode     = mode;
     c->ctx      = ctx;
-    c->pool     = p;
+    c->ticket_keys = tcn_atomic_uint32_new();
```
Scottmitch: is this correct? `ticket_keys` is of type `tcn_ssl_ticket_key_t` but this is assigning it to a new atomic uint32 of type `void*`, right?

Scottmitch: Was the intention that this should be `ticket_keys_new` and `ticket_keys` should be initialized in some other way? It is freed above using `OPENSSL_free`.

Scottmitch:
@normanmaurer - I would like to still remove APR ... we should investigate building a glibc shared library, either publish these instructions or the library, and have folks on older distributions depend upon this instead of apr. Newer distributions which have glibc that support c++14 (eventually every one over time) will require not additional dependency. The statically compiled artifacts may be a bit more tricky due to licensing of glibc.

kvc-code:
How does this affect the ability to build against `musl` C library on Alpine? At present we at least have the ability to build for ourselves on Alpine but sounds like this change may lock that out as well.

normanmaurer:
Let me close this for now

PR #258, opened 2017-03-29
Remove supported_ssl_opts and hasOps method

normanmaurer:
Motivation:

    When try to compile on macOS 10.12.4 the build fails due implicit type conversion:

    [INFO] src/ssl.c:235:6: error: implicit conversion from 'long' to 'jint' (aka 'int') changes value from 4292828079 to -2139217 [-Werror,-Wconstant-conversion]
    [INFO]      | 0;
    [INFO]      ^~~

    Modifications:

    Just remove supported_ssl_opts and the hasOps(...) method as we not use it anyway.

    Result:

    Compile works again on macOS 10.12.4 and less code


openssl-dynamic/src/main/c/ssl.c
```diff
@@ -106,7 +106,7 @@ struct TCN_bio_bytebuffer {
  * so that the return value for SSL_OP_FUTURE_WORKAROUND will only be
  * reported by versions that actually support that specific workaround.
  */
-static const jint supported_ssl_opts = 0
+static const jlong supported_ssl_opts = 0
```
Scottmitch: my preference is to remove all this code and the `hasOp` method. The `hasOp` method is not used in Netty, this code is likely incomplete at compile time, and  is very likely inaccurate if dynamic linking is used.

normanmaurer: good point... let me do this. Less code FTW!

normanmaurer:
@Scottmitch done and changed commit message as well.

openssl-dynamic/src/main/c/ssl.c
Scottmitch: +1 for removal! also fyi `jlong` (64 bit) does not necessarily compatible with `long` (at least 32 bit)

normanmaurer: done

Scottmitch:
`Report supported_ssl_opts and hasOps method ` -> `Remove supported_ssl_opts and hasOps method` ?

Scottmitch:
also update the GitHub PR description/title after the comment is updated.

normanmaurer:
@Scottmitch done

normanmaurer:
Cherry-picked into master (659d8c665ed3964892f15fb77fa4360e007c0187)

PR #259, opened 2017-03-29
Allow to compile with openssl 1.0.1

normanmaurer:
Motivation:

After upgrading the gcc version of the CI the build failed with:

[07:48:57][io.netty:netty-tcnative] [INFO] src/sslcontext.c: In function 'cert_requested':
[07:48:57][io.netty:netty-tcnative] [INFO] src/sslcontext.c:1438:13: error: implicit declaration of function 'SSL_add0_chain_cert' [-Werror=implicit-function-declaration]
[07:48:57][io.netty:netty-tcnative] [INFO]          if (SSL_add0_chain_cert(ssl, sk_X509_value(chain, i)) <= 0) {

This is due a missing ifdef condition for openssl < 1.0.2.

Modifications:

Fix ifdef

Result:

Be able to compile against openssl 1.0.1 again with newer gcc versions.

### DIFF
--- a/openssl-dynamic/src/main/c/sslcontext.c
+++ b/openssl-dynamic/src/main/c/sslcontext.c
@@ -1354,7 +1354,7 @@ static jobjectArray principalBytes(JNIEnv* e, const STACK_OF(X509_NAME)* names) 
 }
 
 static int cert_requested(SSL* ssl, X509** x509Out, EVP_PKEY** pkeyOut) {
-#if defined(LIBRESSL_VERSION_NUMBER)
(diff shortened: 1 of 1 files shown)


normanmaurer:
See also:

https://garage.netty.io/teamcity/viewLog.html?buildId=18601&buildTypeId=pulls_netty_tcnative&tab=buildLog#_focus=386

@Scottmitch after this one is merged I also want to investigate in using weak linking for GCC as we discussed when I was in california

normanmaurer:
Also just to make it more clear it failed because `SSL_add0_chain_cert` only exists in openssl >= 1.0.2

normanmaurer:
Cherry-picked into master (058c38a4a1012d1f189eacc5691349eff7067862)

PR #260, opened 2017-03-29
Allow to verify if an error code is valid for CertificateVerifier

normanmaurer:
Motivation:

CertificateVerifier allows to return an errorCode. We should have a static method to validate that the error code that may be returned is valid.

Modifications:

- Make CertificateVerifier an abstract class and add method to check if errorCode is valid.

Result:

Easier way to validate if errorCode is valid.

### DIFF
--- a/openssl-dynamic/pom.xml
+++ b/openssl-dynamic/pom.xml
@@ -107,6 +107,10 @@
           <family>windows</family>
         </os>
       </activation>
+      <properties>
+        <!-- Skip tests on windows as its not easy to load the openssl library we link against in the tests -->
+        <skipTests>true</skipTests>
+      </properties>
       <build>
         <plugins>
           <plugin>
--- a/openssl-dynamic/src/main/java/io/netty/internal/tcnative/CertificateVerifier.java
+++ b/openssl-dynamic/src/main/java/io/netty/internal/tcnative/CertificateVerifier.java
@@ -15,6 +15,10 @@
  */
 package io.netty.internal.tcnative;
 
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
 import static io.netty.internal.tcnative.NativeStaticallyReferencedJniMethods.*;
 
 /**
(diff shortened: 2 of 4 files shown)


openssl-dynamic/src/main/java/io/netty/internal/tcnative/CertificateVerifier.java
```diff
@@ -25,73 +29,155 @@
...
+    public static final int X509_V_ERR_SUITE_B_INVALID_CURVE = x509vErrSuiteBInvalidCurve();
+    public static final int X509_V_ERR_SUITE_B_INVALID_SIGNATURE_ALGORITHM = x509vErrSuiteBInvalidSignatureAlgorithm();
+    public static final int X509_V_ERR_SUITE_B_LOS_NOT_ALLOWED = x509vErrSuiteBLosNotAllowed();
+    public static final int X509_V_ERR_SUITE_B_CANNOT_SIGN_P_384_WITH_P_256 = x509vErrSuiteBCannotSignP384WithP256();
+    public static final int X509_V_ERR_HOSTNAME_MISMATCH = x509vErrHostnameMismatch();
+    public static final int X509_V_ERR_EMAIL_MISMATCH = x509vErrEmailMismatch();
+    public static final int X509_V_ERR_IP_ADDRESS_MISMATCH = x509vErrIpAddressMismatch();
+    public static final int X509_V_ERR_DANE_NO_MATCH = x509vErrDaneNoMatch();
+
+    private static final Set<Integer> ERRORS;
```
netkins: MINOR Reorder the modifiers to comply with the Java Language Specification. [rule](https://garage.netty.io/sonarqube/coding_rules#rule_key=squid%3AModifiersOrderCheck)

normanmaurer:
@Scottmitch PTAL... I will check if I can add a unit test for this easily here

openssl-static/pom.xml
```diff
@@ -35,6 +35,9 @@
     <linkStatic>true</linkStatic>
     <!-- Don't deploy this artifact to Maven Central -->
     <maven.deploy.skip>true</maven.deploy.skip>
+
+    <!-- Skip tests for now as there is something wrong with the static build for openssl when running on macOS -->
+    <skipTests>true</skipTests>
```
normanmaurer: @Scottmitch I will investigate in this as a separate issue. Not super urgent tho as we not publish static builds for openssl anyway

openssl-dynamic/pom.xml
```diff
@@ -107,6 +107,10 @@
           <family>windows</family>
         </os>
       </activation>
+      <properties>
+        <!-- Skip tests on windows as its not easy to load the openssl library we link against in the tests -->
+        <skipTests>true</skipTests>
```
normanmaurer: @Scottmitch there may be a way to fix this, but I am a bit short on time and would love to merge this first. We can do a followup if needed

normanmaurer: Can be removed once https://github.com/netty/netty-tcnative/pull/261 is merged :)

openssl-static/pom.xml
normanmaurer: Can be removed once https://github.com/netty/netty-tcnative/pull/261 is merged :)

openssl-dynamic/src/test/java/io/netty/internal/tcnative/CertificateVerifierTest.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @Test
+    public void testNonValidErrorCode() {
+        // Will overflow on first increment
+        int errorCode = Integer.MAX_VALUE;
+        do {
+            errorCode++;
+            Assert.assertTrue(errorCode < Integer.MAX_VALUE);
+        } while (CertificateVerifier.isValid(errorCode));
```
Scottmitch: this seems like a roundabout way to do:

```java
assertFalse(CertificateVerifier.isValid(Integer.MIN_VALUE));
```

Can we simplify this test?

openssl-dynamic/src/test/java/io/netty/internal/tcnative/CertificateVerifierTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+public class CertificateVerifierTest extends AbstractNativeTest {
+
+    @Test
+    public void testValidErrorCode() throws Exception {
+        Field[] fields = CertificateVerifier.class.getFields();
+        for (Field field : fields) {
+            if (field.isAccessible()) {
+                int errorCode = field.getInt(null);
+                Assert.assertTrue(CertificateVerifier.isValid(errorCode));
```
Scottmitch: add a string description which shows what number `errorCode` is missing.

openssl-dynamic/src/test/java/io/netty/internal/tcnative/CertificateVerifierTest.java
normanmaurer: sure... just let us hope this one will never be valid ;)

Scottmitch: What is the intention of this test? To prove that at least 1 negative value is invalid? Seems strange in general. If we are just trying to demonstrate there is at least 1 error code that is not valid we have a few options:

- create some private markers to indicate "highest" and "lowest" valid value and use these in the test. Negative of this approach is we have to worry about updating these just for testing.
- make the set package private, sort all the values, and pick a value not in the set.
- randomly choose values until we find one which is not in the set and make sure its not valid.

normanmaurer: yeah that it not always return true. Maybe using Integer.MIN_VALUE is good enough as you said.

openssl-dynamic/src/main/java/io/netty/internal/tcnative/CertificateVerifier.java
```diff
@@ -25,73 +29,155 @@
...
+
+    static {
+        Set<Integer> errors = new HashSet<Integer>();
+        errors.add(X509_V_OK);
+        errors.add(X509_V_ERR_UNSPECIFIED);
+        errors.add(X509_V_ERR_UNABLE_TO_GET_ISSUER_CERT);
```
trustin: How about just using reflection to get the fields start with `X509_V_ERR` (and `X509_V_OK`)?

normanmaurer: @trustin I was considering this but did not do because of jdk9 and secruitymanagers. Reflection may be disallowed. This is why I added the test-case which will fail if we miss to at stuff to the map.

trustin: > jdk9 and secruitymanagers

Really? I thought there's no security restriction on accessing the fields that belong to itself reflectively?

normanmaurer: I am not sure :) Better be safe then sorry and like I said we will catch it with the unit test anyway.

trustin: I'm fine with that although I still believe it's gonna be fine. :-)

normanmaurer:
Cherry-picked into master after addressing @Scottmitch s comments

PR #261, opened 2017-03-30
Fix openssl-static build on macOS

normanmaurer:
Motivation:

Previous the openssl-static build was broken and so an exception was thrown when tried to using it.

Modifications:

Use proper Config command which is needed on macOS. See also https://github.com/rbsec/sslscan/issues/59 .

Result:

openssl-static build works on macOS as well.

### DIFF
--- a/openssl-static/pom.xml
+++ b/openssl-static/pom.xml
@@ -179,10 +179,10 @@
       </build>
     </profile>
     <profile>
-      <id>build-openssl-linux-mac</id>
+      <id>build-openssl-linux</id>
       <activation>
         <os>
-          <family>!windows</family>
+          <family>unix</family>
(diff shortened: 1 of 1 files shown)


openssl-static/pom.xml
```diff
@@ -179,10 +179,10 @@
       </build>
     </profile>
     <profile>
-      <id>build-openssl-linux-mac</id>
+      <id>build-openssl-linux</id>
```
Scottmitch: maven doesn't think mac is in the unix family 😢

normanmaurer: yeah... sad! 👎

normanmaurer:
Cherry-picked into master (a08537e)

PR #262, opened 2017-03-31
Cleanup from 563307abc5955194d93d5f9c71caa1c5a3346388

Scottmitch:
Motivation:
563307abc5955194d93d5f9c71caa1c5a3346388 introduced some code which may lead to a double delete of memory. c->sni_hostname_matcher was not always set to NULL which could lead to ssl_context_cleanup double freeing the memory.

Modifications:
- setSniHostnameMatcher should always set c->sni_hostname_matcher and c->sni_hostname_matcher_method

Result:
No more double free.


Scottmitch:
@normanmaurer - PTAL

normanmaurer:
LGTM ... thanks for the cleanup

Scottmitch:
master (ff161b6)

PR #267, opened 2017-04-26
Fix typo.

davidben:
Motivation:

There is a typo. I'm guess this has never been built with asserts on.

Modifications:

Fix typo.

Result:

Typo is fixed.


Scottmitch:
master (37e2e62f)

Thanks @davidben !

PR #270, opened 2017-05-08
Remove SSL_PROTOCOL_ALL

Scottmitch:
Motivation:
SSL_PROTOCOL_ALL includes SSLv2 which has long been deprecated. We should no longer support this definition and instead use SSL_PROTOCOL_TLS.

Modifications:
- Remove SSL_PROTOCOL_ALL

Result:
SSLv2 is not coupled with the default definition SSL_PROTOCOL_ALL.


Scottmitch:
actually let me just close this PR ... this makes supporting sslv2 a bit easier. if we don't want to support sslv2 any more we can consider this in the future.

PR #272, opened 2017-05-17
Support shading of netty-tcnative  

normanmaurer:
Motivation:

If netty-tcnative's class files are renamed and the type references are updated (shaded) the native libraries will not function. netty-tcnative uses implicit JNI bindings which requires the fully qualified java type names to match the method signatures of the native methods. This means netty-tcnative can not be shaded atm.

Modifications:

- Dynamic register JNI methods
- Based on the library name a different package can now be used

Result:

- netty-tcnative can now be shaded.

### DIFF
--- a/openssl-dynamic/src/main/c/bb.c
+++ b/openssl-dynamic/src/main/c/bb.c
@@ -30,6 +30,7 @@
  */
 
 #include "tcn.h"
+#include "bb.h"
 
 TCN_IMPLEMENT_CALL(jlong, Buffer, address)(TCN_STDARGS, jobject bb)
 {
@@ -42,3 +43,21 @@ TCN_IMPLEMENT_CALL(jlong, Buffer, size)(TCN_STDARGS, jobject bb)
     UNREFERENCED(o);
     return (*e)->GetDirectBufferCapacity(e, bb);
 }
+
+// JNI Method Registration Table Begin
+static const JNINativeMethod method_table[] = {
+  { TCN_METHOD_TABLE_ENTRY(address, (Ljava/nio/ByteBuffer;)J, Buffer) },
+  { TCN_METHOD_TABLE_ENTRY(size, (Ljava/nio/ByteBuffer;)J, Buffer) }
+};
+
+static const jint method_table_size = sizeof(method_table) / sizeof(method_table[0]);
+// JNI Method Registration Table End
+
+jint netty_internal_tcnative_Buffer_JNI_OnLoad(JNIEnv* env, const char* packagePrefix) {
+    if (netty_internal_tcnative_util_register_natives(env, packagePrefix, "io/netty/internal/tcnative/Buffer", method_table, method_table_size) != 0) {
+        return JNI_ERR;
+    }
+    return TCN_JNI_VERSION;
+}
+
+void netty_internal_tcnative_Buffer_JNI_OnUnLoad(JNIEnv* env) { }
--- a/openssl-dynamic/src/main/c/jnilib.c
+++ b/openssl-dynamic/src/main/c/jnilib.c
@@ -28,14 +28,26 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#ifndef WIN32
+// It's important to have #define _GNU_SOURCE before any other include as otherwise it will not work.
+// See http://stackoverflow.com/questions/7296963/gnu-source-and-use-gnu
+#define _GNU_SOURCE
+#include <dlfcn.h>
+#else
+#define MAX_DLL_PATH_LEN 2048
+#endif /* WIN32 */
 
 #include "tcn.h"
 #include "apr_version.h"
 #include "apr_atomic.h"
 #include "apr_strings.h"
+#include "bb.h"
+#include "native_constants.h"
+#include "ssl.h"
+#include "sslcontext.h"
 
 #ifndef TCN_JNI_VERSION
-#define TCN_JNI_VERSION JNI_VERSION_1_4
+#define TCN_JNI_VERSION JNI_VERSION_1_6
 #endif
 
 apr_pool_t *tcn_global_pool = NULL;
@@ -48,87 +60,6 @@ static jclass    byteArrayClass;
 static jclass    keyMaterialClass;
 static jfieldID  keyMaterialCertificateChainFieldId;
 static jfieldID  keyMaterialPrivateKeyFieldId;
-
-/* Called by the JVM when APR_JAVA is loaded */
-JNIEXPORT jint JNICALL JNI_OnLoad_netty_tcnative(JavaVM *vm, void *reserved) 
-{
-    JNIEnv *env;
-    apr_version_t apv;
-    int apvn;
-
-    UNREFERENCED(reserved);
-    if ((*vm)->GetEnv(vm, (void **)&env, TCN_JNI_VERSION)) {
-        return JNI_ERR;
-    }
-    tcn_global_vm = vm;
-
-    /* Before doing anything else check if we have a valid
-     * APR version.
-     */
-    apr_version(&apv);
-    apvn = apv.major * 1000 + apv.minor * 100 + apv.patch;
-    if (apvn < 1201) {
-        tcn_Throw(env, "Unsupported APR version (%s)",
-                  apr_version_string());
-        return JNI_ERR;
-    }
-
-
-    /* Initialize global java.lang.String class */
-    TCN_LOAD_CLASS(env, jString_class, "java/lang/String", JNI_ERR);
-
-    TCN_GET_METHOD(env, jString_class, jString_init,
-                   "<init>", "([B)V", JNI_ERR);
-    TCN_GET_METHOD(env, jString_class, jString_getBytes,
-                   "getBytes", "()[B", JNI_ERR);
-
-    // Load the class which makes JNI references available in a static scope before loading any other classes.
-    if ((*env)->FindClass(env, "io/netty/internal/tcnative/NativeStaticallyReferencedJniMethods") == NULL) {
-        return JNI_ERR;
-    }
-
-    TCN_LOAD_CLASS(env, byteArrayClass, "[B", JNI_ERR);
-    TCN_LOAD_CLASS(env, keyMaterialClass, "io/netty/internal/tcnative/CertificateRequestedCallback$KeyMaterial", JNI_ERR);
-
-    TCN_GET_FIELD(env, keyMaterialClass, keyMaterialCertificateChainFieldId,
-                   "certificateChain", "J", JNI_ERR);
-    TCN_GET_FIELD(env, keyMaterialClass, keyMaterialPrivateKeyFieldId,
-                   "privateKey", "J", JNI_ERR);
-
-    return TCN_JNI_VERSION;
-}
-
-/* Called by the JVM when APR_JAVA is loaded */
-JNIEXPORT jint JNICALL JNI_OnLoad(JavaVM *vm, void *reserved)
-{
-    return JNI_OnLoad_netty_tcnative(vm, reserved);
-}
-
-/* Called by the JVM before the APR_JAVA is unloaded */
-JNIEXPORT void JNICALL JNI_OnUnload_netty_tcnative(JavaVM *vm, void *reserved)
-{
-    JNIEnv *env;
-
-    UNREFERENCED(reserved);
-
-    if ((*vm)->GetEnv(vm, (void **)&env, TCN_JNI_VERSION)) {
-        return;
-    }
-
-    if (tcn_global_pool) {
-        TCN_UNLOAD_CLASS(env, jString_class);
-        apr_terminate();
-    }
-
-    TCN_UNLOAD_CLASS(env, byteArrayClass);
-    TCN_UNLOAD_CLASS(env, keyMaterialClass);
-}
-
-/* Called by the JVM before the APR_JAVA is unloaded */
-JNIEXPORT void JNICALL JNI_OnUnload(JavaVM *vm, void *reserved)
-{
-    JNI_OnUnload_netty_tcnative(vm, reserved);
-}
 
 jstring tcn_new_stringn(JNIEnv *env, const char *str, size_t l)
 {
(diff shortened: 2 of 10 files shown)


normanmaurer:
@Scottmitch @nmittler FYI... The second commit also needs to be ported to netty itself, thats why it kept it separate for now.

Also in the long run I wonder if we can share the loading code.

openssl-dynamic/src/main/c/bb.c
```diff
@@ -30,6 +30,7 @@
  */
 
 #include "tcn.h"
+#include "bb.h"
 
 TCN_IMPLEMENT_CALL(jlong, Buffer, address)(TCN_STDARGS, jobject bb)
```
Scottmitch: `TCN_IMPLEMENT_CALL` and `TCN_METHOD_TABLE_ENTRY` should no longer exist nor be used in `netty-tcnative`. All of these methods should be static. We want to be sure the methods can only be registered with JNI via the manual registration, and we never want the implicit name based registration to occur.

openssl-dynamic/src/main/c/jnilib.c
```diff
@@ -225,3 +152,271 @@ jint tcn_get_java_env(JNIEnv **env)
     }
     return JNI_OK;
 }
+
```
Scottmitch: kill line

openssl-dynamic/src/main/c/jnilib.c
```diff
@@ -225,3 +152,271 @@ jint tcn_get_java_env(JNIEnv **env)
...
+
+char* netty_internal_tcnative_util_prepend(const char* prefix, const char* str) {
```
Scottmitch: TODO: share copied code with `netty-unix-common`

openssl-dynamic/src/main/c/jnilib.c
```diff
@@ -225,3 +152,271 @@ jint tcn_get_java_env(JNIEnv **env)
...
+    nettyClassName = NULL;
+    if (nativeCls == NULL) {
+        return JNI_ERR;
+    }
+
+    return (*env)->RegisterNatives(env, nativeCls, methods, numMethods);
+}
+
+static char* netty_internal_tcnative_util_strndup(const char *s, size_t n) {
+// windows don't has strndup
```
Scottmitch: `don't has` -> `does not have`

openssl-dynamic/src/main/c/jnilib.c
```diff
@@ -225,3 +152,271 @@ jint tcn_get_java_env(JNIEnv **env)
...
+                   "certificateChain", "J", JNI_ERR);
+    TCN_GET_FIELD(env, keyMaterialClass, keyMaterialPrivateKeyFieldId,
+                   "privateKey", "J", JNI_ERR);
+
+
+    return TCN_JNI_VERSION;
+}
+
+void netty_internal_tcnative_Library_JNI_OnUnLoad(JNIEnv* env) {
+    if (tcn_global_pool) {
```
Scottmitch: `tcn_global_pool != NULL`

openssl-dynamic/src/main/c/jnilib.c
```diff
@@ -225,3 +152,271 @@ jint tcn_get_java_env(JNIEnv **env)
...
+
+    TCN_UNLOAD_CLASS(env, byteArrayClass);
+    TCN_UNLOAD_CLASS(env, keyMaterialClass);
+
+    netty_internal_tcnative_Buffer_JNI_OnUnLoad(env);
+    netty_internal_tcnative_NativeStaticallyReferencedJniMethods_JNI_OnUnLoad(env);
+    netty_internal_tcnative_SSL_JNI_OnUnLoad(env);
+    netty_internal_tcnative_SSLContext_JNI_OnUnLoad(env);
+}
+
```
Scottmitch: kill line

openssl-dynamic/src/main/c/jnilib.c
```diff
@@ -225,3 +152,271 @@ jint tcn_get_java_env(JNIEnv **env)
...
+    name = dlinfo.dli_fname;
+#else
+    HMODULE module = NULL;
+    if (GetModuleHandleExA(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS | GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT, (void*) parsePackagePrefix, &module) == 0){
+        fprintf(stderr, "FATAL: netty-tcnative JNI call to GetModuleHandleExA failed!\n");
+        return JNI_ERR;
+    }
+
+    char dllPath[2048];
+    if (GetModuleFileNameA(module, dllPath, 2048) == 0) {
```
Scottmitch: `2048` -> `sizeof dllPath / sizeof(dllPath[0])` (or make a constant)

Scottmitch: https://msdn.microsoft.com/en-us/library/windows/desktop/ms683197(v=vs.85).aspx

> Windows XP:  The string is truncated to nSize characters and is not null-terminated.

I guess we should force the last index to NULL otherwise infinite loop ... or we could make sure the string utility methods don't rely upon NULL (e.g. `strlen`)

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -1652,3 +1653,121 @@ TCN_IMPLEMENT_CALL(void, SSLContext, disableOcsp)(TCN_STDARGS, jlong ctx) {
...
+    char* dynamicTypeName = netty_internal_tcnative_util_prepend(packagePrefix, "io/netty/internal/tcnative/CertificateVerifier;");
+    char* tmpMethodSignature = netty_internal_tcnative_util_prepend("(JL", dynamicTypeName);
+    JNINativeMethod* dynamicMethod = &dynamicMethods[fixed_method_table_size];
+    dynamicMethod->name = "setCertVerifyCallback";
+    dynamicMethod->signature = netty_internal_tcnative_util_prepend(tmpMethodSignature, ")V");
+    dynamicMethod->fnPtr = (void *) TCN_FUNCTION_NAME(SSLContext, setCertVerifyCallback);
+    free(dynamicTypeName);
+    free(tmpMethodSignature);
+
+    dynamicTypeName = netty_internal_tcnative_util_prepend(packagePrefix, "io/netty/internal/tcnative/CertificateRequestedCallback;");
```
Scottmitch: `io/netty/internal/tcnative/CertificateRequestedCallback;` -> `io/netty/internal/tcnative/CertificateRequestedCallback;)V`. Then you can get rid of the temporary allocations/copies (`netty_internal_tcnative_util_prepend(tmpMethodSignature, ")V")`) and not have to worry about freeing more memory etc...

openssl-dynamic/src/main/c/bb.c
normanmaurer: Yep 👍

openssl-dynamic/src/main/c/jnilib.c
normanmaurer: +1 good catch

openssl-dynamic/src/main/c/sslcontext.c

openssl-dynamic/src/main/c/bb.c
```diff
@@ -42,3 +43,21 @@ TCN_IMPLEMENT_CALL(jlong, Buffer, size)(TCN_STDARGS, jobject bb)
...
+// JNI Method Registration Table End
+
+jint netty_internal_tcnative_Buffer_JNI_OnLoad(JNIEnv* env, const char* packagePrefix) {
+    if (netty_internal_tcnative_util_register_natives(env, packagePrefix, "io/netty/internal/tcnative/Buffer", method_table, method_table_size) != 0) {
+        return JNI_ERR;
+    }
+    return JNI_VERSION_1_6;
+}
+
+void netty_internal_tcnative_Buffer_JNI_OnUnLoad(JNIEnv* env) { }
```
Scottmitch: nit: newline

openssl-dynamic/src/main/c/jnilib.c
```diff
@@ -225,3 +154,276 @@ jint tcn_get_java_env(JNIEnv **env)
...
+        return JNI_ERR;
+    }
+
+    return (*env)->RegisterNatives(env, nativeCls, methods, numMethods);
+}
+
+static char* netty_internal_tcnative_util_strndup(const char *s, size_t n) {
+// windows does not have strndup
+#ifdef WIN32
+    char* copy = strdup(s);
```
Scottmitch: can we just use `_strdup` instead?

https://msdn.microsoft.com/en-us/library/ms235454.aspx
```
These POSIX functions are deprecated. Use the ISO C++ conformant _strdup, _wcsdup, _mbsdup instead.
```

openssl-dynamic/src/main/c/jnilib.c
```diff
@@ -225,3 +154,276 @@ jint tcn_get_java_env(JNIEnv **env)
...
+    TCN_LOAD_CLASS(env, byteArrayClass, "[B", JNI_ERR);
+
+    char* keyMaterialClassName = netty_internal_tcnative_util_prepend(packagePrefix, "io/netty/internal/tcnative/CertificateRequestedCallback$KeyMaterial");
+    jclass keyMaterialClassLocal = (*env)->FindClass(env, keyMaterialClassName);
+    free(keyMaterialClassName);
+    keyMaterialClassName = NULL;
+    if (keyMaterialClassLocal == NULL) {
+        return JNI_ERR;
+    }
+    keyMaterialClass = (*env)->NewGlobalRef(env, keyMaterialClassLocal);        \
```
Scottmitch: remove the trailing `\`

openssl-dynamic/src/main/c/jnilib.c
normanmaurer: sure

openssl-dynamic/src/main/c/jnilib.c
normanmaurer: was done

normanmaurer:
@Scottmitch happy with it now ?

openssl-dynamic/src/main/c/jnilib.c
```diff
@@ -225,3 +156,276 @@ jint tcn_get_java_env(JNIEnv **env)
...
+    if (keyMaterialClassLocal == NULL) {
+        return JNI_ERR;
+    }
+    keyMaterialClass = (*env)->NewGlobalRef(env, keyMaterialClassLocal);
+
+    TCN_GET_FIELD(env, keyMaterialClass, keyMaterialCertificateChainFieldId,
+                   "certificateChain", "J", JNI_ERR);
+    TCN_GET_FIELD(env, keyMaterialClass, keyMaterialPrivateKeyFieldId,
+                   "privateKey", "J", JNI_ERR);
+
```
Scottmitch: nit: kill line

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -1463,7 +1465,15 @@ TCN_IMPLEMENT_CALL(void, SSLContext, setCertRequestedCallback)(TCN_STDARGS, jlon
...
     } else {
         jclass callback_class = (*e)->GetObjectClass(e, callback);
-        jmethodID method = (*e)->GetMethodID(e, callback_class, "requested", "(J[B[[B)Lio/netty/internal/tcnative/CertificateRequestedCallback$KeyMaterial;");
+
+        char* tmpSignature = netty_internal_tcnative_util_prepend(pPrefix, "io/netty/internal/tcnative/CertificateRequestedCallback$KeyMaterial;");
+        char* requestedSignature = netty_internal_tcnative_util_prepend("(J[B[[B)L", tmpSignature);
+
+        jmethodID method = (*e)->GetMethodID(e, callback_class, "requested", requestedSignature);
+
+        free(tmpSignature);
```
Scottmitch: you can free this immediately after you use it to build `requestedSignature` right?

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -1463,7 +1465,15 @@ TCN_IMPLEMENT_CALL(void, SSLContext, setCertRequestedCallback)(TCN_STDARGS, jlon
         SSL_CTX_set_client_cert_cb(c->ctx, NULL);
     } else {
         jclass callback_class = (*e)->GetObjectClass(e, callback);
-        jmethodID method = (*e)->GetMethodID(e, callback_class, "requested", "(J[B[[B)Lio/netty/internal/tcnative/CertificateRequestedCallback$KeyMaterial;");
+
+        char* tmpSignature = netty_internal_tcnative_util_prepend(pPrefix, "io/netty/internal/tcnative/CertificateRequestedCallback$KeyMaterial;");
+        char* requestedSignature = netty_internal_tcnative_util_prepend("(J[B[[B)L", tmpSignature);
```
Scottmitch: consider just building this signature in the jni_onload, holding a reference to it, and freeing it in the jni_unload instead of building it on demand.

openssl-dynamic/src/main/c/sslcontext.c
normanmaurer: yep sure... it not matters tho

openssl-dynamic/src/main/c/sslcontext.c
normanmaurer: good idea let me do this

normanmaurer:
@Scottmitch addressed your last comments. PTAL again buddy

openssl-dynamic/src/main/c/sslcontext.c
Scottmitch: in general freeing as soon as possible is a good practice to follow ... leaves more memory for the intermediate calls, and less likely to be forgotten if some exceptional code path is taken.

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -40,6 +40,8 @@
 
 extern apr_pool_t *tcn_global_pool;
 
+static char *requestedSignature;
```
Scottmitch: consider clarifying this variable name to provide more context (e.g. `keyMaterialRequestedMethodSignature`)

openssl-dynamic/src/main/c/jnilib.c
normanmaurer: done

openssl-dynamic/src/main/c/jnilib.c
```diff
@@ -225,3 +156,275 @@ jint tcn_get_java_env(JNIEnv **env)
     }
     return JNI_OK;
 }
+
+// TODO: Share code with netty natives utilities.
+char* netty_internal_tcnative_util_prepend(const char* prefix, const char* str) {
+    if (prefix == NULL) {
```
weissi: this method body should be equivalent to

    char *result = NULL;
    asprintf(&result, "%s%s", prefix ?: "", str);
    return result;

if you don't want to use GCC extensions, replace `prefix ?: ""` by `prefix ? prefix : ""`

openssl-dynamic/src/main/c/jnilib.c
```diff
@@ -225,3 +156,275 @@ jint tcn_get_java_env(JNIEnv **env)
...
+    for (; temp != packageNameEnd; ++temp) {
+        if (*temp == '-') {
+            *temp = '/';
+        }
+    }
+    // Make sure packagePrefix is terminated with the '/' JNI package separator.
+    if(*(--temp) != '/') {
+        temp = packagePrefix;
+        packagePrefix = netty_internal_tcnative_util_prepend(packagePrefix, "/");
+        free(temp);
```
weissi: this must be wrong, `temp` is modified above (in `--temp`) and then freed. Should be `free(packagePrefix)`, no?

openssl-dynamic/src/main/c/jnilib.c
```diff
@@ -225,3 +156,275 @@ jint tcn_get_java_env(JNIEnv **env)
...
+
+    if (status == JNI_ERR) {
+        fprintf(stderr, "FATAL: netty-tcnative encountered unexpected library path: %s\n", name);
+        return JNI_ERR;
+    }
+    tcn_global_vm = vm;
+
+    jint ret = netty_internal_tcnative_Library_JNI_OnLoad(env, packagePrefix);
+
+    if (packagePrefix != NULL) {
```
weissi: `free(NULL)` is okay so that check is superfluous

openssl-dynamic/src/main/c/jnilib.c
normanmaurer: Good catch! Let me fix and also fix the original code from which this is ported

openssl-dynamic/src/main/c/jnilib.c
normanmaurer: True let me remove it

openssl-dynamic/src/main/c/jnilib.c
Scottmitch: I would prefer not to use format based based methods when non-format methods can be easily used instead.

openssl-dynamic/src/main/c/jnilib.c
normanmaurer: Ah sorry... @Scottmitch is right all good here

normanmaurer:
Cherry-picked into master (a61c1a897ef13916a70ff1bd7fb676cdae447ff8) .... finally 👍

Xeli:
Sorry to jump on this closed pull request, but is it actually possible to shade tcnative now? I can't seem to get it to work. See https://stackoverflow.com/questions/46650776/after-shading-io-netty-tcnative-stops-working for details.

Any help would be greatly appreciated!

PR #275, opened 2017-05-29
Introduce uber-snapshot and uber-staging profiles which allows to pub…

normanmaurer:
…lish and build uber jars

Motivation:

As part of our release process we need to deploy an uber jar for our boringssl-static module, which was possible before with a property. Unfortunally it was still not possible to deploy an uber jar for a SNAPSHOT version and so also impossible to install it locally. Futhermore how all of this is handled is not consistent with how we build the "uber all" jar.

Modifications:

- Add uber-staging and uber-snapshot profile
- uber-snapshot profile not includes the windows native libs as we not publish snapshots for it (yet)

Result:

Easier deployment / release process for the uber jar

### DIFF
--- a/boringssl-static/pom.xml
+++ b/boringssl-static/pom.xml
@@ -251,24 +251,19 @@
       without a classifier).
 
       To build from the top level, run:
-      mvn clean deploy -pl boringssl-static -P uber -DstagingRepositoryId={repoId}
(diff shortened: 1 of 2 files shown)


normanmaurer:
Cherry-picked into master (b4e2f40)

PR #277, opened 2017-06-06
Correctly detect if we compile on windows

normanmaurer:
Motivation:

a61c1a897ef13916a70ff1bd7fb676cdae447ff8 introduced some platform specific code but not correctly detected if we are on windows in all cases.

Modifications:

Correctly detect.

Result:

Be able to compile on windows as well


normanmaurer:
Cherry-picked into master (83cd901)

PR #279, opened 2017-06-07
Set SSL_CTX_set_allow_unknown_alpn_protos when using boringssl

normanmaurer:
Motivation:

We should set SSL_CTX_set_allow_unknown_alpn_protos when using boringssl to make its behaviour consistent with openssl and libressl.

Modifications:

Set SSL_CTX_set_allow_unknown_alpn_protos if boringssl is used and the SSLContext is created for client-side usage. This is done by a macro-guard so it needs to be linked against boringssl on compile time.

Result:

More consistent behaviour.


normanmaurer:
Cherry-picked into master (a54b661)

PR #281, opened 2017-06-08
Not force to run autoconf and compile multiple times

normanmaurer:
Motivation:

We should not force autoconf and compile as this will result in multiple executions and so slow down the build.

Modifications:

Not force by default.

Result:

Faster build


normanmaurer:
Cherry-picked into master (c07f2bf)

PR #282, opened 2017-06-08
Fix minor lint errors

carl-mastrangelo:
Motivation:

The Google linter complains when compiling this code, and they seem to be out of style with the rest of the code

Modifications:

Minor spacing differences.

Result:

Quieter compilation


Scottmitch:
master (df24571c)

PR #283, opened 2017-06-08
Enable static compilation of Netty TCNative

carl-mastrangelo:
Motivation:

Java8 Allows for rolling multiple JNI modules into a single blob
by using a namespace after JNI_Onload.  This was done for
netty-tcnative in cf8d01b1ec6872c5ae2530cc40cecdd2caa37ff8.

However, in the recent commit to make it shading aware, this
functionality was lost.

Modifications:
This PR reintroduces the ability to load netty-tcnative.
Additionally, since loading native code this way can be done
without shading, the preprocessor define `TCN_NOT_DYNAMIC` is
added to disable the `dlopen` calls to find the true name.

Result:
Java8 compilation continues to work.


carl-mastrangelo:
@Scottmitch @nmittler fyi

normanmaurer:
@carl-mastrangelo thanks! Cherry-picked into master (afc6664)

PR #284, opened 2017-06-12
Cleanup from afc666462b67bb43075fe3b34c8855ae3869ad07

Scottmitch:
Motivaiton:
Commit afc6664 contains an unecessary reference to parsePackagePrefix.

Modificaitons:
- Remove unecessary line of code
- netty_internal_tcnative_util_rstrstr can be static to jnilib.c
- netty_internal_tcnative_util_strndup, netty_internal_tcnative_util_rstrstr, netty_internal_tcnative_util_strstr_last, and parsePackagePrefix can all be excluded if the TCN_NOT_DYNAMIC preprocessor flag is not defined

Result:
Less unecessary code.

### DIFF
--- a/openssl-dynamic/src/main/c/jnilib.c
+++ b/openssl-dynamic/src/main/c/jnilib.c
@@ -170,18 +170,6 @@ char* netty_internal_tcnative_util_prepend(const char* prefix, const char* str) 
     return result;
 }
 
-char* netty_internal_tcnative_util_rstrstr(char* s1rbegin, const char* s1rend, const char* s2) {
-    size_t s2len = strlen(s2);
-    char *s = s1rbegin - s2len;
-
-    for (; s >= s1rend; --s) {
-        if (strncmp(s, s2, s2len) == 0) {
-            return s;
-        }
-    }
-    return NULL;
-}
-
 jint netty_internal_tcnative_util_register_natives(JNIEnv* env, const char* packagePrefix, const char* className, const JNINativeMethod* methods, jint numMethods) {
     char* nettyClassName = netty_internal_tcnative_util_prepend(packagePrefix, className);
     jclass nativeCls = (*env)->FindClass(env, nettyClassName);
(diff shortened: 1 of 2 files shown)


Scottmitch:
@carl-mastrangelo - PTAL

openssl-dynamic/src/main/c/jnilib.c
```diff
@@ -411,7 +411,6 @@ jint JNI_OnLoad_netty_tcnative(JavaVM* vm, void* reserved) {
         return JNI_ERR;
     }
 #else
-    (void)parsePackagePrefix;
```
carl-mastrangelo: This was necessary because if `TCN_NOT_DYNAMIC` is defined, then there are no refs to `parsePackagePrefix` which  results in a compilation failure (we use `[-Werror,-Wunused-function]`).   Wrapping that function in the ifndef makes the compiler complain about no refs to `netty_internal_tcnative_util_strstr_last` and `netty_internal_tcnative_util_strndup`.   At that point I realized it was going to complicate the code aloot to find all transitive calls of `parsePackagePrefix` and check if they only had one reference.  This line was the cleanest way I could come up with. 

I also tried making the function static, but I got more compilation errors: 

> error: static declaration of 'netty_internal_tcnative_util_prepend' follows non-static declaration

I was hoping that if the methods were static, the compiler would just ignore them, but I don't know what C compilers do nowadays.

Scottmitch: thanks for the explanation @carl-mastrangelo ... please check the updated code which should address these issues. Let me know if this approach works for you.

carl-mastrangelo: @Scottmitch works for me.  LGTM

openssl-dynamic/src/main/c/jnilib.c
```diff
@@ -411,7 +415,6 @@ jint JNI_OnLoad_netty_tcnative(JavaVM* vm, void* reserved) {
...
     char* packagePrefix = NULL;
 #endif
```
carl-mastrangelo: I missed adding `/* TCN_NOT_DYNAMIC */`  here, fyi.

Scottmitch: @carl-mastrangelo - just noticed this too .. thanks 👍

Scottmitch:
master (4a0684d7)

PR #285, opened 2017-06-12
Package prefix on windows bug

Scottmitch:
Motivation:
In order to support shading we must derive the package prefix. However the algorithm will not work on windows because it prematurely exits because the package prefix is set to packageNameEnd and then checks if packagePrefix equals packageNameEnd and exits.

Modifications:
- Windows package name should look for a path separator '\' or use the libraryPathName if no path separator exist

Result:
Package prefix parsing on windows should work.

### DIFF
--- a/openssl-dynamic/src/main/c/jnilib.c
+++ b/openssl-dynamic/src/main/c/jnilib.c
@@ -210,6 +210,15 @@ static char* netty_internal_tcnative_util_rstrstr(char* s1rbegin, const char* s1
     return NULL;
 }
 
+static char* netty_internal_tcnative_util_rstrchar(char* s1rbegin, const char* s1rend, const char c2) {
+    for (; s1rbegin >= s1rend; --s1rbegin) {
+        if (*s1rbegin == c2) {
+            return s1rbegin;
+        }
+    }
+    return NULL;
+}
+
 static char* netty_internal_tcnative_util_strstr_last(const char* haystack, const char* needle) {
     char* prevptr = NULL;
     char* ptr = (char*) haystack;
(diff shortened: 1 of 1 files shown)


openssl-dynamic/src/main/c/jnilib.c
```diff
@@ -233,19 +242,23 @@ static char* parsePackagePrefix(const char* libraryPathName, jint* status) {
         return NULL;
     }
 #ifdef WIN32
-    // on windows we don't have a lib prefix
```
normanmaurer: @Scottmitch please preserve this comment (as its still true).

openssl-dynamic/src/main/c/jnilib.c
```diff
@@ -233,19 +242,23 @@ static char* parsePackagePrefix(const char* libraryPathName, jint* status) {
...
-    int incr = 0;
-    char* packagePrefix = packageNameEnd;
+    char* packagePrefix = netty_internal_tcnative_util_rstrchar(packageNameEnd, libraryPathName, '\\');
+    if (packagePrefix == NULL) {
+        // The string does not have to specify a path [1].
+        // [1] https://msdn.microsoft.com/en-us/library/windows/desktop/ms683200(v=vs.85).aspx
```
normanmaurer: windows... pff ;)

openssl-dynamic/src/main/c/jnilib.c
Scottmitch: done (and enhanced the comment)

Scottmitch:
master (4deefa35)

PR #286, opened 2017-06-12
Use standard predefined macros to detect windows

Scottmitch:
Motivation:
We currently use non-standard macros (WIN32) to detect the presence of being built on windows. We also use _WIN64 when it is not necessary because _WIN32 will be defined in 32 and 64 bit environments [1].

[1] https://msdn.microsoft.com/en-us/library/b0084kay.aspx

Modifications:
- Use _WIN32 instead of WIN32
- Remove usage of _WIN64 where just _WIN32 is sufficient

Result:
More correct usage of predefined macros.

### DIFF
--- a/openssl-dynamic/src/main/c/jnilib.c
+++ b/openssl-dynamic/src/main/c/jnilib.c
@@ -28,7 +28,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#if !defined(_WIN32) && !defined (_WIN64)
+#ifndef _WIN32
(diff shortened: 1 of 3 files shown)


Scottmitch:
Only the second commit should be reviewed as part of this PR. First commit is from PR https://github.com/netty/netty-tcnative/pull/285.

normanmaurer:
@Scottmitch I think I had this before and it not worked .... let me re-check

normanmaurer:
@Scottmitch seems like it works... So +1 :)

Scottmitch:
master (a84aa446)

PR #287, opened 2017-06-14
Guard against segfault when user pass in invalid pointer address

normanmaurer:
Motivation:

We had some places where we guarded against invalid poiner addresses provided by the user but it was not done everywhere. Also in some places it was only checked with asserts which means only when compiled with asserts enabled.

Modifications:

- Add new macro which can be used for the invalid pointer address checks
- Use the macro everywhere and so make it more consistent

Result:

Safer and more consistent code

### DIFF
--- a/openssl-dynamic/src/main/c/error.c
+++ b/openssl-dynamic/src/main/c/error.c
@@ -32,23 +32,21 @@
 #include "tcn.h"
 #include "apr_strings.h"
 
+static jclass exceptionClass;
+static jclass nullPointerExceptionClass;
+
 /*
  * Convenience function to help throw an java.lang.Exception.
  */
 void tcn_ThrowException(JNIEnv *env, const char *msg)
 {
-    jclass javaExceptionClass;
-
-    javaExceptionClass = (*env)->FindClass(env, "java/lang/Exception");
-    if (javaExceptionClass == NULL) {
-        fprintf(stderr, "Cannot find java/lang/Exception class\n");
-        return;
-    }
-    (*env)->ThrowNew(env, javaExceptionClass, msg);
-    (*env)->DeleteLocalRef(env, javaExceptionClass);
-
+    (*env)->ThrowNew(env, exceptionClass, msg);
 }
 
+void tcn_ThrowNullPointerException(JNIEnv *env, const char *msg)
+{
+    (*env)->ThrowNew(env, nullPointerExceptionClass, msg);
+}
 void tcn_Throw(JNIEnv *env, const char *fmt, ...)
 {
     char msg[TCN_BUFFER_SZ] = {'\0'};
@@ -67,3 +65,16 @@ void tcn_ThrowAPRException(JNIEnv *e, apr_status_t err)
     apr_strerror(err, serr, 512);
     tcn_ThrowException(e, serr);
 }
+
+jint netty_internal_tcnative_Error_JNI_OnLoad(JNIEnv* env, const char* packagePrefix) {
+
+    TCN_LOAD_CLASS(env, exceptionClass, "java/lang/Exception", JNI_ERR);
+    TCN_LOAD_CLASS(env, nullPointerExceptionClass, "java/lang/NullPointerException", JNI_ERR);
+
+    return TCN_JNI_VERSION;
+}
+
+void netty_internal_tcnative_Error_JNI_OnUnLoad(JNIEnv* env) {
+     TCN_UNLOAD_CLASS(env, exceptionClass);
+     TCN_UNLOAD_CLASS(env, nullPointerExceptionClass);
+ }
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -338,10 +338,7 @@ TCN_IMPLEMENT_CALL(jint, SSL, bioLengthByteBuffer)(TCN_STDARGS, jlong bioAddress
     BIO* bio = J2P(bioAddress, BIO*);
     struct TCN_bio_bytebuffer* bioUserData;
 
-    if (bio == NULL) {
-        tcn_ThrowException(e, "bio is null");
-        return 0;
-    }
+    TCN_CHECK_NULL(bio, bioAddress, 0);
 
     bioUserData = (struct TCN_bio_bytebuffer*) BIO_get_data(bio);
     return bioUserData == NULL ? 0 : bioUserData->bufferLength;
@@ -351,10 +348,7 @@ TCN_IMPLEMENT_CALL(jint, SSL, bioLengthNonApplication)(TCN_STDARGS, jlong bioAdd
     BIO* bio = J2P(bioAddress, BIO*);
     struct TCN_bio_bytebuffer* bioUserData;
 
-    if (bio == NULL) {
-        tcn_ThrowException(e, "bio is null");
-        return 0;
-    }
+    TCN_CHECK_NULL(bio, bioAddress, 0);
 
     bioUserData = (struct TCN_bio_bytebuffer*) BIO_get_data(bio);
     return bioUserData == NULL ? 0 : bioUserData->nonApplicationBufferLength;
@@ -803,14 +797,7 @@ TCN_IMPLEMENT_CALL(jlong /* SSL * */, SSL, newSSL)(TCN_STDARGS,
     int *handshakeCount = NULL;
     tcn_ssl_ctxt_t *c = J2P(ctx, tcn_ssl_ctxt_t *);
 
-    if (c == NULL) {
-        tcn_ThrowException(e, "ssl ctx is null");
-        return 0;
-    }
-    if (c->ctx == NULL) {
-        tcn_ThrowException(e, "ctx is null");
-        return 0;
-    }
+    TCN_CHECK_NULL(c, ctx, 0);
 
     UNREFERENCED_STDARGS;
 
@@ -856,10 +843,7 @@ TCN_IMPLEMENT_CALL(jint, SSL, getError)(TCN_STDARGS,
                                        jint ret) {
     SSL *ssl_ = J2P(ssl, SSL *);
 
-    if (ssl_ == NULL) {
-        tcn_ThrowException(e, "ssl is null");
-        return 0;
-    }
+    TCN_CHECK_NULL(ssl_, ssl, 0);
 
     UNREFERENCED_STDARGS;
 
@@ -874,14 +858,8 @@ TCN_IMPLEMENT_CALL(jint /* status */, SSL, bioWrite)(TCN_STDARGS,
     BIO* bio = J2P(bioAddress, BIO*);
     void* wbuf = J2P(wbufAddress, void*);
 
-    if (bio == NULL) {
-        tcn_ThrowException(e, "bio is null");
-        return 0;
-    }
-    if (wbuf == NULL) {
-        tcn_ThrowException(e, "wbuf is null");
-        return 0;
-    }
+    TCN_CHECK_NULL(bio, bioAddress, 0);
+    TCN_CHECK_NULL(wbuf, wbufAddress, 0);
 
     UNREFERENCED_STDARGS;
 
@@ -896,8 +874,8 @@ TCN_IMPLEMENT_CALL(void, SSL, bioSetByteBuffer)(TCN_STDARGS,
     BIO* bio = J2P(bioAddress, BIO*);
     char* buffer = J2P(bufferAddress, char*);
     struct TCN_bio_bytebuffer* bioUserData = NULL;
-    TCN_ASSERT(bio != NULL);
-    TCN_ASSERT(buffer != NULL);
+    TCN_CHECK_NULL(bio, bioAddress, /* void */);
+    TCN_CHECK_NULL(buffer, bufferAddress, /* void */);
 
     bioUserData = (struct TCN_bio_bytebuffer*) BIO_get_data(bio);
     TCN_ASSERT(bioUserData != NULL);
@@ -944,14 +922,8 @@ TCN_IMPLEMENT_CALL(jint /* status */, SSL, writeToSSL)(TCN_STDARGS,
     SSL *ssl_ = J2P(ssl, SSL *);
     void *w = J2P(wbuf, void *);
 
-    if (ssl_ == NULL) {
-        tcn_ThrowException(e, "ssl is null");
-        return 0;
-    }
-    if (w == NULL) {
-        tcn_ThrowException(e, "wbuf is null");
-        return 0;
-    }
+    TCN_CHECK_NULL(ssl_, ssl, 0);
+    TCN_CHECK_NULL(w, wbuf, 0);
 
     UNREFERENCED_STDARGS;
 
@@ -966,14 +938,8 @@ TCN_IMPLEMENT_CALL(jint /* status */, SSL, readFromSSL)(TCN_STDARGS,
     SSL *ssl_ = J2P(ssl, SSL *);
     void *r = J2P(rbuf, void *);
 
-    if (ssl_ == NULL) {
-        tcn_ThrowException(e, "ssl is null");
-        return 0;
-    }
-    if (r == NULL) {
-        tcn_ThrowException(e, "rbuf is null");
-        return 0;
-    }
+    TCN_CHECK_NULL(ssl_, ssl, 0);
+    TCN_CHECK_NULL(r, rbuf, 0);
 
     UNREFERENCED_STDARGS;
 
@@ -985,10 +951,7 @@ TCN_IMPLEMENT_CALL(jint /* status */, SSL, getShutdown)(TCN_STDARGS,
                                                         jlong ssl /* SSL * */) {
     SSL *ssl_ = J2P(ssl, SSL *);
 
-    if (ssl_ == NULL) {
-        tcn_ThrowException(e, "ssl is null");
-        return 0;
-    }
+    TCN_CHECK_NULL(ssl_, ssl, 0);
 
     UNREFERENCED_STDARGS;
 
@@ -1001,10 +964,7 @@ TCN_IMPLEMENT_CALL(void, SSL, setShutdown)(TCN_STDARGS,
                                            jint mode) {
     SSL *ssl_ = J2P(ssl, SSL *);
 
-    if (ssl_ == NULL) {
-        tcn_ThrowException(e, "ssl is null");
-        return;
-    }
+    TCN_CHECK_NULL(ssl_, ssl, /* void */);
 
     UNREFERENCED_STDARGS;
 
@@ -1018,10 +978,9 @@ TCN_IMPLEMENT_CALL(void, SSL, freeSSL)(TCN_STDARGS,
     tcn_ssl_ctxt_t* c = NULL;
     tcn_ssl_verify_config_t* verify_config = NULL;
     SSL *ssl_ = J2P(ssl, SSL *);
-    if (ssl_ == NULL) {
-        tcn_ThrowException(e, "ssl is null");
-        return;
-    }
+
+    TCN_CHECK_NULL(ssl_, ssl, /* void */);
+
     c = SSL_get_app_data2(ssl_);
     handshakeCount = SSL_get_app_data3(ssl_);
     verify_config = SSL_get_app_data4(ssl_);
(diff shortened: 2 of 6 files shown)


openssl-dynamic/src/main/c/error.c
```diff
@@ -49,6 +49,19 @@ void tcn_ThrowException(JNIEnv *env, const char *msg)
 
 }
 
+void tcn_ThrowIllegalArgumentException(JNIEnv *env, const char *msg)
+{
+    jclass javaExceptionClass;
+
+    javaExceptionClass = (*env)->FindClass(env, "java/lang/IllegalArgumentException");
```
normanmaurer: Also note that I could have cached this class but I didnt feel we really need to as this should not happen frequently

Scottmitch: I would prefer to cache all "static" classes ... but I agree its urgent and we can defer this.

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -803,14 +803,7 @@ TCN_IMPLEMENT_CALL(jlong /* SSL * */, SSL, newSSL)(TCN_STDARGS,
...
 
-    if (c == NULL) {
-        tcn_ThrowException(e, "ssl ctx is null");
-        return 0;
-    }
-    if (c->ctx == NULL) {
-        tcn_ThrowException(e, "ctx is null");
-        return 0;
-    }
+    TCN_CHECK_NULL(c, "ctx is null", 0);
```
Scottmitch: return `NULL` instead of `0`

Scottmitch: general comment where a pointer is returned

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1001,10 +970,7 @@ TCN_IMPLEMENT_CALL(void, SSL, setShutdown)(TCN_STDARGS,
                                            jint mode) {
     SSL *ssl_ = J2P(ssl, SSL *);
 
-    if (ssl_ == NULL) {
-        tcn_ThrowException(e, "ssl is null");
-        return;
-    }
+    TCN_CHECK_NULL(ssl_, "ssl is null",);
```
Scottmitch: consider clarifying this as follow: `TCN_CHECK_NULL(ssl_, "ssl is null", /*void*/);`

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1018,10 +984,9 @@ TCN_IMPLEMENT_CALL(void, SSL, freeSSL)(TCN_STDARGS,
     tcn_ssl_ctxt_t* c = NULL;
     tcn_ssl_verify_config_t* verify_config = NULL;
     SSL *ssl_ = J2P(ssl, SSL *);
-    if (ssl_ == NULL) {
-        tcn_ThrowException(e, "ssl is null");
-        return;
-    }
+
+    TCN_CHECK_NULL(ssl_, "ssl is null",);
```
Scottmitch: consider clarifying this as follow: `TCN_CHECK_NULL(ssl_, "ssl is null", /*void*/);`

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1361,10 +1293,7 @@ TCN_IMPLEMENT_CALL(jlong, SSL, getTimeout)(TCN_STDARGS, jlong ssl)
     SSL *ssl_ = J2P(ssl, SSL *);
     SSL_SESSION *session = NULL;
 
-    if (ssl_ == NULL) {
-        tcn_ThrowException(e, "ssl is null");
-        return 0;
-    }
+    TCN_CHECK_NULL(ssl_, "ssl is null", 0);
```
Scottmitch: `setTimeout` below also needs to be updated ... and change `0` to `NULL` here.

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1408,10 +1337,7 @@ TCN_IMPLEMENT_CALL(void, SSL, setVerify)(TCN_STDARGS, jlong ssl, jint level, jin
     tcn_ssl_ctxt_t* c;
     SSL *ssl_ = J2P(ssl, SSL *);
 
-    if (ssl_ == NULL) {
-        tcn_ThrowException(e, "ssl is null");
-        return;
-    }
+    TCN_CHECK_NULL(ssl_, "ssl is null",);
```
Scottmitch: consider clarifying this as follow: `TCN_CHECK_NULL(ssl_, "ssl is null", /*void*/);`

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1442,10 +1368,7 @@ TCN_IMPLEMENT_CALL(void, SSL, setOptions)(TCN_STDARGS, jlong ssl,
 {
     SSL *ssl_ = J2P(ssl, SSL *);
 
-    if (ssl_ == NULL) {
-        tcn_ThrowException(e, "ssl is null");
-        return;
-    }
+    TCN_CHECK_NULL(ssl_, "ssl is null",);
```
Scottmitch: consider clarifying this as follow: `TCN_CHECK_NULL(ssl_, "ssl is null", /*void*/);`

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1457,10 +1380,7 @@ TCN_IMPLEMENT_CALL(void, SSL, clearOptions)(TCN_STDARGS, jlong ssl,
...
```
Scottmitch: consider clarifying this as follow: `TCN_CHECK_NULL(ssl_, "ssl is null", /*void*/);`

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1666,30 +1560,26 @@ TCN_IMPLEMENT_CALL(void, SSL, setState)(TCN_STDARGS,
                                            jint state) {
     SSL *ssl_ = J2P(ssl, SSL *);
 
-    if (ssl_ == NULL) {
-        tcn_ThrowException(e, "ssl is null");
-        return;
-    }
+    TCN_CHECK_NULL(ssl_, "ssl is null",);
```
Scottmitch: consider clarifying this as follow: `TCN_CHECK_NULL(ssl_, "ssl is null", /*void*/);`

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1747,6 +1635,9 @@ TCN_IMPLEMENT_CALL(void, SSL, setCertificateBio)(TCN_STDARGS, jlong ssl,
                                                          jstring password)
 {
     SSL *ssl_ = J2P(ssl, SSL *);
+
+    TCN_CHECK_NULL(ssl_, "ssl is null",);
```
Scottmitch: add `/*void*/` for the last argument?

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1916,10 +1801,8 @@ TCN_IMPLEMENT_CALL(void, SSL, freeX509Chain)(TCN_STDARGS, jlong x509Chain)
  */
 TCN_IMPLEMENT_CALL(void, SSL, enableOcsp)(TCN_STDARGS, jlong ssl) {
     SSL *ssl_ = J2P(ssl, SSL *);
-    if (ssl_ == NULL) {
-        tcn_ThrowException(e, "ssl is null");
-        return;
-    }
+
+    TCN_CHECK_NULL(ssl_, "ssl is null",);
```
Scottmitch: `/*void*/` as last argument?

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1943,10 +1826,8 @@ TCN_IMPLEMENT_CALL(void, SSL, enableOcsp)(TCN_STDARGS, jlong ssl) {
  */
 TCN_IMPLEMENT_CALL(void, SSL, setOcspResponse)(TCN_STDARGS, jlong ssl, jbyteArray response) {
     SSL *ssl_ = J2P(ssl, SSL *);
-    if (ssl_ == NULL) {
-        tcn_ThrowException(e, "ssl is null");
-        return;
-    }
+
+    TCN_CHECK_NULL(ssl_, "ssl is null",);
```
Scottmitch: `/*void*/` as last argument?

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -348,6 +348,9 @@ TCN_IMPLEMENT_CALL(jlong, SSLContext, make)(TCN_STDARGS, jint protocol, jint mod
 TCN_IMPLEMENT_CALL(jint, SSLContext, free)(TCN_STDARGS, jlong ctx)
```
Scottmitch: did you consider the suggestion in https://github.com/netty/netty/pull/6254#discussion_r121768301 (change return type to jlong and return NULL)?

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -360,9 +363,11 @@ TCN_IMPLEMENT_CALL(void, SSLContext, setContextId)(TCN_STDARGS, jlong ctx,
                                                    jstring id)
 {
     tcn_ssl_ctxt_t *c = J2P(ctx, tcn_ssl_ctxt_t *);
+
+    TCN_CHECK_NULL(c, "ctx is null",);
```
Scottmitch: use `/*void*/` for the last argument? general comment.

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -398,20 +405,23 @@ TCN_IMPLEMENT_CALL(void, SSLContext, clearOptions)(TCN_STDARGS, jlong ctx,
...
 }
 
 TCN_IMPLEMENT_CALL(jboolean, SSLContext, setCipherSuite)(TCN_STDARGS, jlong ctx,
                                                          jstring ciphers)
 {
     tcn_ssl_ctxt_t *c = J2P(ctx, tcn_ssl_ctxt_t *);
-    TCN_ALLOC_CSTRING(ciphers);
     jboolean rv = JNI_TRUE;
 
+    TCN_CHECK_NULL(c, "ctx is null", false);
```
Scottmitch: `false` -> `JNI_FALSE` (return type is jboolean)

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -431,10 +441,12 @@ TCN_IMPLEMENT_CALL(jboolean, SSLContext, setCertificateChainFile)(TCN_STDARGS, j
 {
     tcn_ssl_ctxt_t *c = J2P(ctx, tcn_ssl_ctxt_t *);
     jboolean rv = JNI_FALSE;
+
+    TCN_CHECK_NULL(c, "ctx is null", false);
```
Scottmitch: `false` -> `JNI_FALSE` (return type is jboolean)

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -450,8 +462,9 @@ TCN_IMPLEMENT_CALL(jboolean, SSLContext, setCertificateChainBio)(TCN_STDARGS, jl
     tcn_ssl_ctxt_t *c = J2P(ctx, tcn_ssl_ctxt_t *);
     BIO *b = J2P(chain, BIO *);
 
+    TCN_CHECK_NULL(c, "ctx is null", false);
```
Scottmitch: `false` -> `JNI_FALSE` (return type is jboolean)

Scottmitch: general comment where `false` is returned instead of `JNI_FALSE`

openssl-dynamic/src/main/c/sslcontext.c
normanmaurer: yes but I think I would like to leave it as it is for now to not make it again harder to jump from one version to another for no strong reason

openssl-dynamic/src/main/c/ssl.c
normanmaurer: This does not work an will not compile. If the method signature returns a `jint` or `jlong` I can not return NULL

openssl-dynamic/src/main/c/ssl.c
normanmaurer: sure can do and will do for all of them

openssl-dynamic/src/main/c/sslcontext.c

openssl-dynamic/src/main/c/ssl.c
Scottmitch: @normanmaurer - good point ... `NULL` maybe defined as a pointer ... fine as is .. although it maybe good to have a TCN_JLONG_NULL in the future.

openssl-dynamic/src/main/c/sslcontext.c
Scottmitch: sgtm ... we can delay ... moving tcnative into netty's repo will make this easier later too

openssl-dynamic/src/main/c/error.c
```diff
@@ -49,6 +49,19 @@ void tcn_ThrowException(JNIEnv *env, const char *msg)
...
+    if (javaExceptionClass == NULL) {
+        fprintf(stderr, "Cannot find java/lang/IllegalArgumentException class\n");
```
trustin: Probably better panicking rather than returning?

openssl-dynamic/src/main/c/tcn.h
```diff
@@ -103,6 +104,14 @@ jstring         tcn_new_stringn(JNIEnv *, const char *, size_t);
 
 #define AJP_TO_JSTRING(V)   (*e)->NewStringUTF((e), (V))
 
+#define TCN_CHECK_NULL(V, M, R)                      \
+    TCN_BEGIN_MACRO                                  \
+        if (V == NULL) {                             \
+            tcn_ThrowIllegalArgumentException(e, M); \
```
trustin: I wonder if we can use the `#` operator so that we don't need to require M at all:

    tcn_ThrowIllegalArgumentException(e, #V);

http://en.cppreference.com/w/cpp/preprocessor/replace#.23_and_.23.23_operators

openssl-dynamic/src/main/c/error.c
normanmaurer: @trustin yes! Let me actual cache the class during init and if we can not find it fail the init :)

openssl-dynamic/src/main/c/tcn.h
normanmaurer: maybe but I think "is null" gives a bit more context tho

normanmaurer:
@trustin @Scottmitch PTAL again

openssl-dynamic/src/main/c/sslcontext.c

openssl-dynamic/src/main/c/tcn.h
trustin: What happens if we throw a NPE instead?

normanmaurer: Works for me... @Scottmitch WDYT?

Scottmitch: +1 for NPE

openssl-dynamic/src/main/c/error.c
```diff
@@ -37,18 +37,13 @@
...
-
-    javaExceptionClass = (*env)->FindClass(env, "java/lang/Exception");
-    if (javaExceptionClass == NULL) {
-        fprintf(stderr, "Cannot find java/lang/Exception class\n");
-        return;
-    }
-    (*env)->ThrowNew(env, javaExceptionClass, msg);
-    (*env)->DeleteLocalRef(env, javaExceptionClass);
-
+    (*env)->ThrowNew(env, tcn_get_exception_class(), msg);
```
Scottmitch: the pattern we have been using to partition things is to have the module export a method which is called on ini load and a method for jni unload (e.g. `netty_internal_tcnative_error_JNI_OnLoad` and `netty_internal_tcnative_error_JNI_OnUnLoad`). This will avoid having to expose accessor methods (e.g. `tcn_get_exception_class()` and `tcn_get_illegal_argument_exception_class()`) which are only intended to be used in the "error" module. can we use the same approach here too?

openssl-dynamic/src/main/c/ssl.c
Scottmitch: @normanmaurer - looks like `setTimeout` still needs to be updated

openssl-dynamic/src/main/c/tcn.h
```diff
@@ -103,6 +104,14 @@ jstring         tcn_new_stringn(JNIEnv *, const char *, size_t);
...
+            return R;                                \
```
johnou: return where R is undefined is identical to simple "void" return? pretty cool trick.

normanmaurer: yep :)

openssl-dynamic/src/main/c/tcn.h
normanmaurer: @trustin @Scottmitch I am changing to a NPE but I think we still need `M` as the name of the field may be different (as its really the c field) compared to what the user provide us.

openssl-dynamic/src/main/c/ssl.c
normanmaurer: good catch... also found 3 more places :)

normanmaurer:
@trustin @Scottmitch PTAL again... hopefully the last time 😆

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1629,10 +1520,7 @@ TCN_IMPLEMENT_CALL(jint, SSL, getHandshakeCount)(TCN_STDARGS, jlong ssl)
     int *handshakeCount = NULL;
     SSL *ssl_ = J2P(ssl, SSL *);
 
-    if (ssl_ == NULL) {
-        tcn_ThrowException(e, "ssl is null");
-        return -1;
-    }
+    TCN_CHECK_NULL(ssl_, ssl, 0);
```
johnou: -1 was incorrect before?

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1747,6 +1626,9 @@ TCN_IMPLEMENT_CALL(void, SSL, setCertificateBio)(TCN_STDARGS, jlong ssl,
                                                          jstring password)
 {
     SSL *ssl_ = J2P(ssl, SSL *);
+
+    TCN_CHECK_NULL(ssl_, ssl, /* void */);
+
     BIO *cert_bio = J2P(cert, BIO *);
     BIO *key_bio = J2P(key, BIO *);
```
johnou: TCN_CHECK_NULL cert and key?

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -803,14 +797,7 @@ TCN_IMPLEMENT_CALL(jlong /* SSL * */, SSL, newSSL)(TCN_STDARGS,
...
 
-    if (c == NULL) {
-        tcn_ThrowException(e, "ssl ctx is null");
-        return 0;
-    }
-    if (c->ctx == NULL) {
-        tcn_ThrowException(e, "ctx is null");
-        return 0;
-    }
+    TCN_CHECK_NULL(c, ctx, 0);
```
johnou: c->ctx not checked any more?

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -705,6 +724,9 @@ TCN_IMPLEMENT_CALL(jboolean, SSLContext, setCertificateBio)(TCN_STDARGS, jlong c
                                                          jstring password)
 {
     tcn_ssl_ctxt_t *c = J2P(ctx, tcn_ssl_ctxt_t *);
+
+    TCN_CHECK_NULL(c, ctx, JNI_FALSE);
+
     BIO *cert_bio = J2P(cert, BIO *);
     BIO *key_bio = J2P(key, BIO *);
```
johnou: Cert and key null check?

openssl-dynamic/src/main/c/ssl.c
normanmaurer: not needed

openssl-dynamic/src/main/c/ssl.c
normanmaurer: doesn't really matter. Its just important that we return a number.

openssl-dynamic/src/main/c/ssl.c
normanmaurer: we check this below

openssl-dynamic/src/main/c/sslcontext.c
normanmaurer: we check this below

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -935,19 +961,28 @@ TCN_IMPLEMENT_CALL(jlong, SSLContext, getSessionCacheMode)(TCN_STDARGS, jlong ct
 TCN_IMPLEMENT_CALL(jlong, SSLContext, setSessionCacheTimeout)(TCN_STDARGS, jlong ctx, jlong timeout)
```
Scottmitch: getSessionCacheMode (above) needs null check

normanmaurer: good catch...

normanmaurer:
Cherry-picked into master (cce189a)

PR #288, opened 2017-06-21
Add an alternative library name for Java8 statically linked JNI

carl-mastrangelo:
Motivation:
Java 8 adds a way for JNI code to be loaded by using a suffixed
version of the `JNI_OnLoad` function.  Such a functions looks
like:  `jint JNI_OnLoad_my_code(...)`, and can be loaded with
`System.loadLibrary("my_code")`.  Currently, Netty TCNative uses
dashes "-" instead of underscores in the library name, which makes
it not possible to be defined in C.

A work around for this is just manually call `System.loadLibrary`
with the right name out of band, but since TCNative code can try
multiple library names anyways, it makes since to do it in
`Library`.

Modifcations:
Add a "netty_tcnative" name to try loading in addition to
"netty-tcnative".

Results:
No more out of band initializationl; no more maintaining patches
to get TCNative to work.

### DIFF
--- a/openssl-dynamic/src/main/java/io/netty/internal/tcnative/Library.java
+++ b/openssl-dynamic/src/main/java/io/netty/internal/tcnative/Library.java
@@ -37,7 +37,12 @@ import java.io.File;
 public final class Library {
 
     /* Default library names */
-    private static final String [] NAMES = {"netty-tcnative", "libnetty-tcnative", "netty-tcnative-1", "libnetty-tcnative-1"};
+    private static final String [] NAMES = {
+        "netty-tcnative",
+        "netty_tcnative",
+        "libnetty-tcnative",
+        "netty-tcnative-1",
+        "libnetty-tcnative-1"};
     /*
      * A handle to the unique Library singleton instance.
      */


openssl-dynamic/src/main/java/io/netty/internal/tcnative/Library.java
```diff
@@ -37,7 +37,12 @@
 public final class Library {
 
     /* Default library names */
-    private static final String [] NAMES = {"netty-tcnative", "libnetty-tcnative", "netty-tcnative-1", "libnetty-tcnative-1"};
+    private static final String [] NAMES = {
+        "netty-tcnative",
+        "netty_tcnative",
+        "libnetty-tcnative",
+        "netty-tcnative-1",
```
Scottmitch: @normanmaurer - where is the `-1` suffix used?

we should revisit this when we move tcnative into netty (e.g. just use the `_` version)

carl-mastrangelo:
FYI the corresponding Netty changes are in https://github.com/netty/netty/pull/6884

Scottmitch:
changes generally lgtm ... be nice to reduce the libraries in the future.

openssl-dynamic/src/main/java/io/netty/internal/tcnative/Library.java
normanmaurer: +1

normanmaurer:
Cherry-picked into master (b2ff6891942a4b3ded34560fbfa14379e533bf38)

PR #289, opened 2017-06-22
Correctly handle path seperator when using Windows and try to load na…

normanmaurer:
…tive library.

Motiviation:

Our latest changes broke the build on windows as we only handled \ as path seperator on windows.

Modifications:

When on windows replace \ with / and so allow to load the native library when either \ or / is used.

Result:

Correctly load native lib on windows in all cases and test pass again on windows.

### DIFF
--- a/openssl-dynamic/src/main/c/jnilib.c
+++ b/openssl-dynamic/src/main/c/jnilib.c
@@ -243,8 +243,16 @@ static char* parsePackagePrefix(const char* libraryPathName, jint* status) {
         return NULL;
     }
 #ifdef _WIN32
+    char* tmpLibraryPathName = libraryPathName;
+
+    // replace \ with /
+    for(; *tmpLibraryPathName != '\0'; ++tmpLibraryPathName) {
+        if (*tmpLibraryPathName == '\\') {
+            *tmpLibraryPathName = '/';
+        }
+    }
     // on windows there is no lib prefix so we instead look for the previous path separator or the beginning of the string.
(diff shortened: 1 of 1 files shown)


normanmaurer:
@Scottmitch PTAL... found this while tried to build on windows :/

openssl-dynamic/src/main/c/jnilib.c
```diff
@@ -243,8 +243,16 @@ static char* parsePackagePrefix(const char* libraryPathName, jint* status) {
         return NULL;
     }
 #ifdef _WIN32
+    char* tmpLibraryPathName = libraryPathName;
+
+    // replace \ with /
+    for(; *tmpLibraryPathName != '\0'; ++tmpLibraryPathName) {
```
normanmaurer: @Scottmitch I moved this here to ensure we always have `/` before call `netty_internal_tcnative_util_rstrchar `

openssl-dynamic/src/main/c/jnilib.c
```diff
@@ -418,13 +426,6 @@ jint JNI_OnLoad_netty_tcnative(JavaVM* vm, void* reserved) {
         dllPath[dllPathLen] = '\0';
     }
 
-    char* dllTmpPath = dllPath;
```
Scottmitch: discussed offline with @normanmaurer ... I think getting rid of this path manipulation should be enough.

Scottmitch:
Closing in favor of https://github.com/netty/netty-tcnative/pull/290

openssl-dynamic/src/main/c/jnilib.c
normanmaurer: Yep  👍

PR #290, opened 2017-06-22
Correctly handle loading of native lib on windows when full path is g…

normanmaurer:
…iven

Motivation:

We failed to remove some code that replaced \ with / on windows and so made our parsePackagePrefix function return something invalid on windows.

Modification:

Remove code that replaces \ with /.

Result:

Works again on windows.


normanmaurer:
Cherry-picked into master (9ed3ed8)

PR #293, opened 2017-08-04
Use APR 1.6.2 when static compile

normanmaurer:
Motivation:

We should use APR 1.6.2 (the latest stable version) when compile our static releases. The old release we were using was moved to the archive and so the build fails atm.

Modifications:

Use APR 1.6.2

Result:

Use latest APR version.


normanmaurer:
Pulling this in so the CI not fails anymore.

normanmaurer:
Cherry-picked into master (dbeecde)

PR #294, opened 2017-08-04
Update to libressl 2.5.5

normanmaurer:
Motivation:

libressl 2.5.5 is considered the latest stable release, we should upgrade.

Modifications:

Compile against libressl 2.5.5 in our static builds.

Result:

Use latest libressl version.


normanmaurer:
Cherry-picked into master (dad4150)

PR #295, opened 2017-08-11
Use underscore in native library names to allow loading in all enviro…

normanmaurer:
Motivation:

At the moment we try to load the library using multiple names which includes names using - but also _ . We should just use _ all the time.

This is a followup of cf0c7e0aa4f962b3e65df5fa6ccb34265ab77aec.

Modifications:

Replace - by _

Result:

Make the loading code more consistent.

### DIFF
--- a/boringssl-static/pom.xml
+++ b/boringssl-static/pom.xml
@@ -188,7 +188,7 @@
                       </and>
                       <then>
                         <exec executable="strip" failonerror="true" dir="${nativeLibOnlyDir}/META-INF/native/linux${archBits}/" resolveexecutable="true">
-                          <arg value="libnetty-tcnative.so" />
+                          <arg value="libnetty_tcnative.so" />
                         </exec>
                       </then>
                     </if>
@@ -221,7 +221,7 @@
                 </goals>
                 <phase>compile</phase>
                 <configuration>
-                  <name>netty-tcnative</name>
+                  <name>netty_tcnative</name>
                   <nativeSourceDirectory>${project.basedir}/src/main/c</nativeSourceDirectory>
                   <libDirectory>${nativeLibOnlyDir}</libDirectory>
                   <forceAutogen>${forceAutogen}</forceAutogen>
@@ -342,15 +342,15 @@
                     <mkdir dir="${nativeDir}" />
                     <copy todir="${nativeDir}" flatten="true">
                       <fileset dir="${unpackDir}/osx-x86_64/META-INF/native" />
-                      <globmapper from="libnetty-tcnative.*" to="libnetty-tcnative-osx-x86_64.*" />
+                      <globmapper from="libnetty_tcnative.*" to="libnetty_tcnative_osx_x86_64.*" />
                     </copy>
                     <copy todir="${nativeDir}" flatten="true">
                       <fileset dir="${unpackDir}/linux-x86_64/META-INF/native" />
-                      <globmapper from="libnetty-tcnative.*" to="libnetty-tcnative-linux-x86_64.*" />
+                      <globmapper from="libnetty_tcnative.*" to="libnetty_tcnative_linux_x86_64.*" />
                     </copy>
                     <copy todir="${nativeDir}" flatten="true">
                       <fileset dir="${unpackDir}/windows-x86_64/META-INF/native" />
-                      <globmapper from="netty-tcnative.*" to="netty-tcnative-windows-x86_64.*" />
+                      <globmapper from="netty_tcnative.*" to="netty_tcnative_windows_x86_64.*" />
                     </copy>
                   </target>
                 </configuration>
@@ -375,9 +375,9 @@
                   <instructions>
                     <Export-Package>io.netty.internal.tcnative.*</Export-Package>
                     <Bundle-NativeCode>
-                      META-INF/native/libnetty-tcnative-osx-x86_64.jnilib;osname=macos;osname=macosx;processor=x86_64,
-                      META-INF/native/libnetty-tcnative-linux-x86_64.so;osname=linux;processor=x86_64,
-                      META-INF/native/netty-tcnative-windows-x86_64.dll;osname=win32;processor=x86_64
+                      META-INF/native/libnetty_tcnative_osx_x86_64.jnilib;osname=macos;osname=macosx;processor=x86_64,
+                      META-INF/native/libnetty_tcnative_linux_x86_64.so;osname=linux;processor=x86_64,
+                      META-INF/native/netty_tcnative_windows_x86_64.dll;osname=win32;processor=x86_64
                     </Bundle-NativeCode>
                   </instructions>
                 </configuration>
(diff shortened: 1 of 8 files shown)


normanmaurer:
Also @carl-mastrangelo

nmittler:
@normanmaurer just curious - which environments are problematic? I know that POSIX filenames can't start with a hyphen, but that's not the case here.

normanmaurer:
@nmittler seems like this is problematic on systems that use gsecurity for example:
https://github.com/netty/netty/issues/7069

ejona86:
Huh, that's surprising, especially since `ld-linux.so.2` uses dashes... From a quick reading it seems the problem was the inconsistency; some Netty code expected underscores, some dashes.

Unless I'm missing another piece in the puzzle, the grsecurity comment doesn't seem to be related to underscore vs dash, but instead explaining why the auto-extract doesn't work.

On my Ubuntu 14.04 desktop (with GNOME and such):

```sh
$ ls /lib | grep so | grep -- - | wc -l
7
$ ls /lib | grep so | grep -- _ | wc -l
0
$ ls /usr/lib | grep so | grep -- - | wc -l
217
$ ls /usr/lib | grep so | grep _ | wc -l
24
```

On my Arch "server" (/lib is a symlink to /usr/lib):

```sh
$ ls /usr/lib | grep so | grep -- - | wc -l
418
$ ls /usr/lib | grep so | grep _ | wc -l
269
```

Humorously, some use both, like `libsvn_fs-1.so`.

eun-ice:
Some linux libs use `_` for the name and `-` for the version number.

But you are right, for loading on grsecurity the naming is not the problem. The problem is that the lib tries to extract the prefix by looking at its filename and fails to load if it is not named as expected.

So I too think the inconsistency is the only real problem here.

eun-ice:
With this PR, I can't get tcnative to load at all.

I tried the following names:

libnetty_tcnative.so
libnetty_tcnative_linux-x86_64.so
libnetty-tcnative-linux_x86_64.so
netty_tcnative.so
netty-tcnative.so


22:36:06,021 DEBUG io.netty.util.internal.NativeLibraryLoader:81 - Unable to load the library 'netty-tcnative-linux-x86_64', trying other loading mechanism.
java.lang.UnsatisfiedLinkError: no netty-tcnative-linux-x86_64 in java.library.path
	at io.netty.util.internal.NativeLibraryLoader.loadLibraryByHelper(NativeLibraryLoader.java:209)
22:36:06,028 DEBUG io.netty.util.internal.NativeLibraryLoader:81 - Unable to load the library 'netty-tcnative-linux-x86_64', trying next name...
java.lang.UnsatisfiedLinkError: no netty-tcnative-linux-x86_64 in java.library.path
	at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1867)
	at java.lang.Runtime.loadLibrary0(Runtime.java:870)
	at java.lang.System.loadLibrary(System.java:1122)
	at io.netty.util.internal.NativeLibraryUtil.loadLibrary(NativeLibraryUtil.java:38)
22:36:06,032 DEBUG io.netty.util.internal.NativeLibraryLoader:81 - Unable to load the library 'netty-tcnative-linux-x86_64-fedora', trying other loading mechanism.
java.lang.UnsatisfiedLinkError: no netty-tcnative-linux-x86_64-fedora in java.library.path
	at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1867)
	at java.lang.Runtime.loadLibrary0(Runtime.java:870)
	at java.lang.System.loadLibrary(System.java:1122)
	at io.netty.util.internal.NativeLibraryUtil.loadLibrary(NativeLibraryUtil.java:38)
22:36:06,035 DEBUG io.netty.util.internal.NativeLibraryLoader:81 - Unable to load the library 'netty-tcnative-linux-x86_64-fedora', trying next name...
java.lang.UnsatisfiedLinkError: no netty-tcnative-linux-x86_64-fedora in java.library.path
	at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1867)
	at java.lang.Runtime.loadLibrary0(Runtime.java:870)
	at java.lang.System.loadLibrary(System.java:1122)
	at io.netty.util.internal.NativeLibraryUtil.loadLibrary(NativeLibraryUtil.java:38)
	at io.netty.util.internal.NativeLibraryLoader.loadLibrary(NativeLibraryLoader.java:204)
	at io.netty.util.internal.NativeLibraryLoader.load(NativeLibraryLoader.java:130)
	at io.netty.util.internal.NativeLibraryLoader.loadFirstAvailable(NativeLibraryLoader.java:81)
	at io.netty.handler.ssl.OpenSsl.loadTcNative(OpenSsl.java:423)
22:36:06,039 DEBUG io.netty.util.internal.NativeLibraryLoader:81 - Unable to load the library 'netty-tcnative', trying other loading mechanism.
java.lang.UnsatisfiedLinkError: no netty-tcnative in java.library.path
	at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1867)
	at java.lang.Runtime.loadLibrary0(Runtime.java:870)
	at java.lang.System.loadLibrary(System.java:1122)
	at io.netty.util.internal.NativeLibraryUtil.loadLibrary(NativeLibraryUtil.java:38)
22:36:06,042 DEBUG io.netty.util.internal.NativeLibraryLoader:81 - Unable to load the library 'netty-tcnative', trying next name...
java.lang.UnsatisfiedLinkError: no netty-tcnative in java.library.path
	at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1867)
	at java.lang.Runtime.loadLibrary0(Runtime.java:870)
	at java.lang.System.loadLibrary(System.java:1122)
22:36:06,045 DEBUG io.netty.util.internal.NativeLibraryLoader:81 - Unable to load the library 'netty_tcnative', trying next name...
java.lang.UnsatisfiedLinkError: could not load a native library: netty_tcnative
	at io.netty.util.internal.NativeLibraryLoader.load(NativeLibraryLoader.java:173)
	at io.netty.util.internal.NativeLibraryLoader.loadFirstAvailable(NativeLibraryLoader.java:81)

Scottmitch:
The original motivation for supporting both `_` and `-` was https://github.com/netty/netty-tcnative/pull/288. Basically JDK8+ allows you to call a qualified version of `JNI_OnLoad` which includes the library name, and therefore has to be `_` instead of `-`. I requested that we just support the `_` names (https://github.com/netty/netty/pull/6884#discussion_r123378932) because I would prefer not to attempt to load 2 different libraries. For the following reasons:

- less code complexity in loading
- we have better control over what library name is loaded (we won't accidentally try to load an old crufty library jus because of our load sequencing).

Scottmitch:
@normanmaurer - can you add some of the above comment to your commit message?

normanmaurer:
@Scottmitch yes that is the plan

normanmaurer:
@Scottmitch PTAL again.. also cleaned-up the native names in Library.java

eun-ice:
I think this change is ok, the problem with loading tcnative was related to NativeLibraryLoader in netty.

normanmaurer:
Cherry-picked into master (a773517dd9093a9f69dba160f10c19c557d58f87).

PR #296, opened 2017-08-15
Automatically detect the package prefix when loading the JNI library

Scottmitch:
Motivation:
TCNative supports shading when loading methods, but doesn't support shading for loading the particular JNI library name. This may lead to loading an unintended version of the JNI library.
We should also detect the package prefix automatically the same was as Netty does in Netty's NativeLibraryLoader.

Modifications:
- Copy code from NativeLibraryLoader to detect the package prefix
- Redirect System.loadLibrary calls so we can apply the package prefix

Result:
More correct shading.

### DIFF
--- a/openssl-dynamic/src/main/java/io/netty/internal/tcnative/Library.java
+++ b/openssl-dynamic/src/main/java/io/netty/internal/tcnative/Library.java
@@ -56,7 +56,7 @@ public final class Library {
         StringBuilder err = new StringBuilder();
         for (int i = 0; i < NAMES.length; i++) {
             try {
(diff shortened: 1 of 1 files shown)


Scottmitch:
1 more reason for moving netty-tcnative into Netty's repo.

normanmaurer:
@Scottmitch please ensure to replace `-` by `_` now that https://github.com/netty/netty-tcnative/pull/295 is pulled in :)

normanmaurer:
Cherry-picked into master (3e84155501471251253e138c6b82246a741d7a03).

@Scottmitch thanks!

PR #298, opened 2017-08-16
Ensure we also use underscore for the native library when using windows

normanmaurer:
Motivation:

Commit a773517dd9093a9f69dba160f10c19c557d58f87 broke the build for windows as we missed to adjust the library name in the the VS files.

Modifications:

Replace netty-tcnative with netty_tcnative

Result:

Build works on windows again as well


normanmaurer:
@Scottmitch noticed this will trying to cut the release... we need this fix.

normanmaurer:
Cherry-picked into master (46e8e66)

PR #299, opened 2017-08-17
Do not define JNI_OnLoad when not dynamic

carl-mastrangelo:
Motivation:
Due to an oversight (by myself), linking two JNI modules with
duplicate symbols fails in linking.  This only seems to happen
some of the time (the behavior seems to be different between GCC
and Clang toolchains).  For instance, including both netty tcnative
and netty epoll fails to link because of duplicate JNI_OnLoad
symobols.

Modification:
Do not define the JNI_OnLoad and JNI_OnUnload symbols when
compiling for static linkage, as indicated by the TCN_NOT_DYNAMIC
preprocessor define.  They are never directly called when
statically linked.

Result:
Able to statically compile epoll and tcnative code into a single
binary.

### DIFF
--- a/openssl-dynamic/src/main/c/jnilib.c
+++ b/openssl-dynamic/src/main/c/jnilib.c
@@ -183,7 +183,7 @@ jint netty_internal_tcnative_util_register_natives(JNIEnv* env, const char* pack
     return (*env)->RegisterNatives(env, nativeCls, methods, numMethods);
 }
 
-#ifndef TCN_NOT_DYNAMIC
+#ifndef TCN_BUILD_STATIC
 
 static char* netty_internal_tcnative_util_strndup(const char *s, size_t n) {
 // windows does not have strndup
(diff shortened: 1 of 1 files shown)


openssl-dynamic/src/main/c/jnilib.c
```diff
@@ -441,9 +441,11 @@ jint JNI_OnLoad_netty_tcnative(JavaVM* vm, void* reserved) {
     return ret;
 }
 
+#ifndef TCN_NOT_DYNAMIC
```
Scottmitch: can we avoid double negatives? `#ifdef TCN_DYNAMIC_LINK`?

carl-mastrangelo:
@Scottmitch PTAl

normanmaurer:
Let me merge this one as @Scottmitch approved the same for Netty already.

normanmaurer:
Cherry-picked into master (440154b651ad39e5f9ec32aa207362cd51d9fba1)

PR #302, opened 2017-09-05
Detect processor that is stated in the MANIFEST during compilation

normanmaurer:
Motivation:

At the moment we hardcode x86_64 as processor in the MANIFEST, we should better detect it during compilation to allow people compile their own version of other architectures and still have the correct MANIFEST entry.

Modifications:

Use ${os.detected.arch} to detect the processor

Result:

Easier to build your own version of tcnative on other architectures.

### DIFF
--- a/boringssl-static/pom.xml
+++ b/boringssl-static/pom.xml
@@ -264,6 +264,7 @@
         <libDir>${project.build.directory}/lib</libDir>
         <nativeDir>${project.build.outputDirectory}/META-INF/native</nativeDir>
         <skipTests>true</skipTests>
+        <uberArch>${os.detected.arch}</uberArch>
       </properties>
 
       <repositories>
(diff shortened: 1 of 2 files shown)


boringssl-static/pom.xml
```diff
@@ -375,9 +375,9 @@
                   <instructions>
                     <Export-Package>io.netty.internal.tcnative.*</Export-Package>
                     <Bundle-NativeCode>
-                      META-INF/native/libnetty_tcnative_osx_x86_64.jnilib;osname=macos;osname=macosx;processor=x86_64,
-                      META-INF/native/libnetty_tcnative_linux_x86_64.so;osname=linux;processor=x86_64,
-                      META-INF/native/netty_tcnative_windows_x86_64.dll;osname=win32;processor=x86_64
+                      META-INF/native/libnetty_tcnative_osx_x86_64.jnilib;osname=macos;osname=macosx;processor=${os.detected.arch},
+                      META-INF/native/libnetty_tcnative_linux_x86_64.so;osname=linux;processor=${os.detected.arch},
```
carl-mastrangelo: Is the library name still x86_64.so, even if it my not be 64 bit?

normanmaurer: good catch! Let me fix this 👍

normanmaurer:
@carl-mastrangelo can you have a look again ?

normanmaurer:
Cherry-picked into master (372b3e3f7ede6c0341f7636e5e702cedf4ffbf24)

PR #305, opened 2017-09-19
Add missing include

carl-mastrangelo:
Motivation:
sslcontext.c uses HMAC_Init_ex(), so it should include hmac.h.

In boringssl, ssl.h used to include hmac.h, but in the future it
will not.

Modification:
Add missing hmac.h

Result:
Proper includes achieved


normanmaurer:
Cherry-picked into master (66d14e4d7f764bb07cb68ef274e5599d8671cb4d).

@carl-mastrangelo thanks!

PR #310, opened 2017-10-17
Fix shading support on Windows

trask:
When shading (based on the work from #272), I'm getting the following error on Windows:
```
java.lang.NoClassDefFoundError: shaded/nio/netty/internal/tcnative/Library
```
Note the extra **n** in the above class name (shaded/**n**io/netty/internal/tcnative/Library, when it should be looking for shaded/io/netty/internal/tcnative/Library).

Everything works works as expected on Linux.

I haven't gone through the build steps to test this, but from reading the source, it looks like a simple off-by-one error, hoping you'll agree.

Thanks.


normanmaurer:
@trask can you please sign our icla and let me know once done ? Thanks:

https://netty.io/s/icla

trask:
@normanmaurer sure, signed

normanmaurer:
@trask thanks!

Cherry-picked into master (d7fd591)

Issue #308, opened 2017-10-09
netty-tcnative-parent-2.0.5.Final build issue


dineshappavoo:
I am trying to build netty-tcnative-parent-2.0.5.Final with openssl-static. I am getting build error while fetching apr 1.5.2

```

[INFO] --- maven-antrun-plugin:1.8:run (build-apr-linux-mac) @ netty-tcnative-openssl-static ---
[INFO] Executing tasks

main:
      [get] Getting: http://www.us.apache.org/dist/apr/apr-1.5.2.tar.gz
      [get] To: /Users/ad/Documents/docs/ssl/openssl/version-2-0-5/netty-tcnative/openssl-static/target/apr-1.5.2.tar.gz
      [get] Error opening connection java.io.FileNotFoundException: http://www.us.apache.org/dist/apr/apr-1.5.2.tar.gz
      [get] Error opening connection java.io.FileNotFoundException: http://www.us.apache.org/dist/apr/apr-1.5.2.tar.gz
      [get] Error opening connection java.io.FileNotFoundException: http://www.us.apache.org/dist/apr/apr-1.5.2.tar.gz
      [get] Can't get http://www.us.apache.org/dist/apr/apr-1.5.2.tar.gz to /Users/ad/Documents/docs/version-2-0-5/netty-tcnative/openssl-static/target/apr-1.5.2.tar.gz
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
```

It looks like pom file has apr `1.5.2` version [ https://github.com/netty/netty-tcnative/blob/netty-tcnative-parent-2.0.5.Final/pom.xml#L52 ] from `http://www.us.apache.org/dist/apr/${aprArchiveFile}` URL. [ ref: https://github.com/netty/netty-tcnative/blob/7a72db574db3b4df20d48b5092baa633aa773964/pom.xml#L322 ]. 

But this version is not available in the apache distribution http://www.us.apache.org/dist/apr/ (atleast in this URL).

I am not sure there is fix for this to use different apache dist URL. But when I changed the apr version to 1.6.2 in netty-tcnative/pom.xml , it worked. I wanted to notify the team.

floragunncom:
See https://mail-archives.apache.org/mod_mbox/apr-dev/201710.mbox/%3cCACsi2532W69+QrKoYUwB--pqOF1f_OfmzY6A8u2AybsWYVdDyg@mail.gmail.com%3e

[...]

PR #314, opened 2017-10-23
Fetch apr from archive.apache.org

normanmaurer:
Motivation:

We should fetch apr from archive.apache.org and so also allow to build
later tags.

Modifications:

Change apr url to archive.apache.org

Result:

Fixes [#308].


normanmaurer:
Let me just merge this as it is only a change in the pom.xml which is tiny

normanmaurer:
Cherry-picked into master (d41c6e1758913333a5007c8cd3c198c4cc7980c3)

Issue #307, opened 2017-09-27
Netty OpenSsl.isAvailable() check doesn't handle lib loading issue on 32bit JVM


doom369:
Details are here : https://github.com/netty/netty/issues/7150#issuecomment-332542876

Brief summary:

Right now `tcnative` library has name `libnetty_tcnative.so` and this name doesn't include OS classifier that prevents netty from loading this library if system architecture doesn't match with packed version. For example, epoll native library name look like this `libnetty_transport_native_epoll_x86_64.so`.

normanmaurer:
@doom369 PTAL https://github.com/netty/netty-tcnative/pull/315

PR #315, opened 2017-10-24
Include architecture in native lib name.

normanmaurer:
Motivation:

We should include the architecture in the native lib name to allow not load it if the architecture not match.
This is related to https://github.com/netty/netty/pull/7163.

Modifications:

Add architecture to native lib name.

Result:

Fixes [#307].

### DIFF
--- a/boringssl-static/pom.xml
+++ b/boringssl-static/pom.xml
@@ -188,7 +188,7 @@
                       </and>
                       <then>
                         <exec executable="strip" failonerror="true" dir="${nativeLibOnlyDir}/META-INF/native/linux${archBits}/" resolveexecutable="true">
-                          <arg value="libnetty_tcnative.so" />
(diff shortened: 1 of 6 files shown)


carl-mastrangelo:
There won't be any arch present when statically compiling tcnative into the java binary.   The name of the library has to match the JNI_OnLoad suffix.  

At some point I'll send out a unit test, btu I'm not quite sure yet how to do it in a non bazel environment.

normanmaurer:
@carl-mastrangelo so you say this patch is not correct at all ? We are doing the same for epoll / kqueue transport and also the boringssl uber jar contains the arch since forever.

normanmaurer:
@carl-mastrangelo so ok to merge this one ?

normanmaurer:
@carl-mastrangelo for sure...

normanmaurer:
Cherry-picked into master (1eeb4e11cc9890bdc76caa8f921b67c459c573b6)

PR #319, opened 2017-10-31
Use correct CXXFLAGS depending on OS.

normanmaurer:
Motivation:

72b45657c99986bfbd14164cbbe31fc709d0fcd0 added some extra CXXFLAGS for the build which fixed the build on linux but broke it on macOS.

Modifications:

Use correct CXXFLAGS depending on OS.

Result:

Build works again on all platforms.

### DIFF
--- a/boringssl-static/pom.xml
+++ b/boringssl-static/pom.xml
@@ -141,28 +141,35 @@
                       <equals arg1="${os.detected.name}" arg2="windows" />
                       <then>
                         <!-- On Windows, build with /MT for static linking -->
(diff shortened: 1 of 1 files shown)


normanmaurer:
@Scottmitch PTAL... now the build works again on all platforms

normanmaurer:
Let me just merge this so it builds again on all platforms. @Scottmitch let me know if you have any followup concerns.

normanmaurer:
Cherry-picked into master (02b7acb045fccbe0be2e4fe03f4997f5c97a8d55)

PR #320, opened 2017-11-03
Use openssl 1.0.2m in static build

normanmaurer:
Motivation:

We need to use openssl 1.0.2m in the static build as it was released.

Modifications:

Update openssl version.

Result:

Using latest openssl release.


normanmaurer:
Cherry-picked into master (e6e3fc5) so it compiles again

Issue #321, opened 2017-11-06
netty-tcnative-parent-2.0.7.Final openssl-static build issue


dineshappavoo:
I am trying to run openssl-static with netty-tcnative-parent-2.0.7.Final release. I am getting the following error while opening `netty-tcnative/openssl-static/target/old/1.0.2/openssl-1.0.2l.tar.gz`

```

     [exec] /usr/bin/install -c -m 755 apr-config.out /Users/dinesh/Documents/docs/ssl/openssl/version-2-0-7/netty-tcnative/openssl-static/target/apr/bin/apr-1-config
[INFO] Executed tasks
[INFO]
[INFO] --- maven-antrun-plugin:1.8:run (build-openssl) @ netty-tcnative-openssl-static ---
[INFO] Executing tasks

main:
      [ftp] getting files
      [ftp] transferring old/1.0.2/openssl-1.0.2l.tar.gz to /Users/dinesh/Documents/docs/ssl/openssl/version-2-0-7/netty-tcnative/openssl-static/target/old/1.0.2/openssl-1.0.2l.tar.gz
      [ftp] 1 files retrieved
     [exec] tar: Error opening archive: Failed to open 'openssl-1.0.2l.tar.gz'
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 59.966 s
[INFO] Finished at: 2017-11-06T15:14:24-08:00
[INFO] Final Memory: 14M/256M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-antrun-plugin:1.8:run (build-openssl) on project netty-tcnative-openssl-static: An Ant BuildException has occured: exec returned: 1
[ERROR] around Ant part ...<exec resolveexecutable="true" failonerror="true" dir="/Users/dinesh/Documents/docs/ssl/openssl/version-2-0-7/netty-tcnative/openssl-static/target/" executable="tar">... @ 9:171 in /Users/dinesh/Documents/docs/ssl/openssl/version-2-0-7/netty-tcnative/openssl-static/target/antrun/build-main.xml
[ERROR] -> [Help 1]
[ERROR]
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR]
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[...]

PR #322, opened 2017-11-13
Allow to also compile from older tags when new openssl version was re…

normanmaurer:
…leased.

Motivation:

When a new version of openssl is released they move the old release to a sub-directory which will fail the build from a tag which depends on an older version.

Modifications:

Add an extra ant steps to move the tar.gz file to the correct directory if it was an old release.

Result:

Build works even if it lists an older version of openssl. Fixes [#321].

### DIFF
--- a/openssl-static/pom.xml
+++ b/openssl-static/pom.xml
@@ -138,6 +138,13 @@
                         <include name="**/openssl-${opensslVersion}.tar.gz" />
                       </fileset>
                     </ftp>
+                    <if>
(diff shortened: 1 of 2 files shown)


normanmaurer:
Let me just merge this as its only build config stuff ;)

normanmaurer:
Cherry-picked into master (97dd02e930351bb931f55268ad8c602df38e4f85)

Scottmitch:
+1 for merging ... if the build breaks in the future we will fix it

Issue #263, opened 2017-04-06
Remove usage of SSL_set_state


snazy:
According to the [OpenSSL changelog](https://www.openssl.org/news/changelog.html), _SSL_set_state() has been removed altogether._ in 1.1.0.

Therefore, compiling netty-tcnative against 1.1.0e results in a compile error.

This is related to https://github.com/netty/netty/issues/6320

```
[INFO] --- maven-hawtjni-plugin:1.11:build (build-native-lib) @ netty-tcnative-openssl-static ---
[INFO] executing: /bin/sh -c ./autogen.sh
[INFO] executing: /bin/sh -c ./configure --disable-ccache --prefix=/Users/snazy/devel/netty/netty-tcnative/openssl-static/target/native-build/target --with-ssl=/Users/snazy/devel/netty/netty-tcnative/openssl-static/target/ssl --with-apr=/Users/snazy/devel/netty/netty-tcnative/openssl-static/target/apr
[INFO] executing: /bin/sh -c make install
[INFO] /bin/sh ./libtool  --tag=CC   --mode=compile gcc -DHAVE_CONFIG_H -I. -I./src     -O3 -Werror -fno-omit-frame-pointer -Wunused-variable -DHAVE_OPENSSL -I/Users/snazy/devel/netty/netty-tcnative/openssl-static/target/ssl/include  -I/Users/snazy/devel/netty/netty-tcnative/openssl-static/target/apr/include/apr-1  -D_LARGEFILE64_SOURCE -I/Library/Java/JavaVirtualMachines/jdk1.8.0_121.jdk/Contents/Home/include -I/Library/Java/JavaVirtualMachines/jdk1.8.0_121.jdk/Contents/Home/include/darwin -c -o src/bb.lo src/bb.c
[...]

PR #325, opened 2017-12-24
Remove calls to SSL_set_state

FauxFaux:
See https://github.com/netty/netty/issues/6320 for the only user of this function, and the related discussion.

After this change, netty-tcnative will build on Debian sid/unstable, against openssl 1.1.

---

### Motivation:

`SSL_set_state` has gone from openssl 1.1. Calling it is, and probably always has been, incorrect. It has only ever been used to drive an incorrect renegotiation algorithm. Doing renegotiation in this manner is potentially insecure. There have been at least two insecure renegotiation vulnerabilities in users of the OpenSSL library.

Renegotiation is not necessary for correct operation of the TLS protocol.

### Modifications:

Remove the method, and its implementation.

### Result:

Fixes: #263
Requires: https://github.com/netty/netty/issues/6320

### DIFF
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -1546,18 +1546,6 @@ TCN_IMPLEMENT_CALL(jint, SSL, renegotiate)(TCN_STDARGS,
     return SSL_renegotiate(ssl_);
 }
 
-TCN_IMPLEMENT_CALL(void, SSL, setState)(TCN_STDARGS,
-                                           jlong ssl, /* SSL * */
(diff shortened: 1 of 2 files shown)


normanmaurer:
Fine with pulling this in once we are all in agreement here: https://github.com/netty/netty/pull/7544

Scottmitch:
Followup PR: https://github.com/netty/netty-tcnative/pull/326

floragunncom:
@Scottmitch Any chance to get this released soon?

normanmaurer:
I will cut a release next week

PR #335, opened 2018-03-26
Pass a string length to X509_VERIFY_PARAM_set1_host.

agl:
BoringSSL is going to make using the “zero length means call strlen” an
error for various reasons so this change will allow the code to continue
working with BoringSSL. Additionally, by passing a length, OpenSSL can
catch cases where there's an embedded NUL in the string. Lastly, it's
not clear that JNI guarantees that the result of GetStringUTFChars is
NUL terminated[1].

This change also makes an empty hostname an error. OpenSSL considers
setting an empty hostname to disable all checks, but that seems like a
surprising behaviour when this function already supports `null` to skip
hostname checking.

[1] https://docs.oracle.com/javase/6/docs/technotes/guides/jni/spec/functions.html#wp17265


normanmaurer:
@agl this looks good! Thanks a lot!

PR #338, opened 2018-04-13
We need to check if the java methods threw an exception and if so sto…

normanmaurer:
…p executing our native method.

Motivation:

If call java methods via `Call*Method` we need to check for possible exceptions that were thrown by the java code.

Modifications:

Check for exceptions after calling java methods and if some are pending return early.

Result:

More correct and safe JNI code.


normanmaurer:
Related to https://github.com/netty/netty/issues/7508

carl-mastrangelo:
LGTM

PR #341, opened 2018-05-14
Use SSL_MODE_AUTO_RETRY by default

normanmaurer:
Motivation:

We should enable SSL_MODE_AUTO_RETRY in all cases to guard against all kind of bugs. In fact its enabled by BoringSSL by default already.

Modifications:

Add SSL_MODE_AUTO_RETRY to the mode.

Result:

More correct impl.


normanmaurer:
@agl @davidben @Lukasa @nmittler PTAL as well.

carl-mastrangelo:
heh, Am I on the reviewers list twice?

normanmaurer:
@carl-mastrangelo not sure how this even works... Pressure 🤣

PR #342, opened 2018-05-21
Compile libressl-static against libressl 2.7.3

normanmaurer:
Motivation:

Change the version of libressl to 2.7.3 for netty-tcnative-libressl-static and so ensure we can also compile with newer versions of libressl.

Modifications:

Slightly change the ifdef to only define OPENSSL_VERSION if not using libressl 2.7.x

Result:

Use latest libressl

### DIFF
--- a/openssl-dynamic/src/main/c/ssl_private.h
+++ b/openssl-dynamic/src/main/c/ssl_private.h
@@ -135,7 +135,13 @@ extern const char* TCN_UNKNOWN_AUTH_METHOD;
 #define TLS_method SSLv23_method
 #define TLS_client_method SSLv23_client_method
 #define TLS_server_method SSLv23_server_method
+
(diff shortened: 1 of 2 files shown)


openssl-dynamic/src/main/c/ssl_private.h
```diff
@@ -135,7 +135,11 @@ extern const char* TCN_UNKNOWN_AUTH_METHOD;
 #define TLS_method SSLv23_method
 #define TLS_client_method SSLv23_client_method
 #define TLS_server_method SSLv23_server_method
+
+#if !defined(LIBRESSL_VERSION_NUMBER) || LIBRESSL_VERSION_NUMBER < 0x20700000L
```
trustin: Add some comment, e.g. why we need `< 0x20700000L`?

PR #344, opened 2018-05-30
Providing docker-compose files for centos6 / debian7

normanmaurer:
Motivation:

We should just use docker-compose to allow easy building via docker for the linux distributions we support. This will allow us to enhance to use different java versions easily and we will be able to re-use these for the CI.

Modifications:

Replace old dockerfiles with new ones that are used via docker-compose.

Result:

Easily be able to build netty-tcnative for supported distributions.

### DIFF
--- a/docker/Dockerfile.centos
+++ b/docker/Dockerfile.centos
@@ -1,11 +1,9 @@
-FROM centos:6
-MAINTAINER netty@googlegroups.com
-
-ENTRYPOINT /bin/bash
+ARG centos_version=6
+FROM centos:$centos_version
+# needed to do again after FROM due to docker limitation
+ARG centos_version
 
 ENV SOURCE_DIR /root/source
-ENV MAVEN_VERSION 3.5.0
-ENV JAVA_VERSION 1.8.0
 ENV CMAKE_VERSION_BASE 3.8
 ENV CMAKE_VERSION $CMAKE_VERSION_BASE.2
 ENV NINJA_VERSION 1.7.2
@@ -20,7 +18,6 @@ RUN yum install -y \
  git \
  glibc-devel \
  gnupg \
- java-$JAVA_VERSION-openjdk-devel \
  libapr1-dev \
  libtool \
  lsb-core \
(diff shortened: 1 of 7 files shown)


normanmaurer:
This mirrors the distributions that are used to release as well.

Also /cc @tomerd

docker/Dockerfile.centos
```diff
@@ -42,13 +41,18 @@ RUN wget -q http://storage.googleapis.com/golang/go$GO_VERSION.linux-amd64.tar.g
 
 RUN wget -q http://archive.apache.org/dist/maven/maven-3/$MAVEN_VERSION/binaries/apache-maven-$MAVEN_VERSION-bin.tar.gz && tar xfz apache-maven-$MAVEN_VERSION-bin.tar.gz && mv apache-maven-$MAVEN_VERSION /opt/ && echo 'PATH=/opt/apache-maven-$MAVEN_VERSION/bin:$PATH' >> ~/.bashrc
 
-RUN echo 'export JAVA_HOME="/usr/lib/jvm/java-$JAVA_VERSION/"' >> ~/.bashrc
```
tomerd: JAVA_VERSION not required any more?

docker/Dockerfile.debian
```diff
@@ -1,16 +1,17 @@
...
-MAINTAINER netty@googlegroups.com
-ENTRYPOINT /bin/bash
+ARG debian_version=7
+FROM debian:$debian_version
+# needed to do again after FROM due to docker limitation
+ARG debian_version
 
-ENV SOURCE_DIR $HOME/source
+ENV SOURCE_DIR /root/source
 ENV MAVEN_VERSION 3.5.0
```
tomerd: ditto

docker/docker-compose.centos.yaml
```diff
@@ -0,0 +1,31 @@
...
+  common: &common
+    image: netty-tcnative-centos:default
+    depends_on: [runtime-setup]
+    volumes:
+      - ~/.ssh:/root/.ssh
+      - ~/.gnupg:/root/.gnupg
+      - ..:/code
+    working_dir: /code
+
+  build:
```
tomerd: test?

docker/docker-compose.centos.yaml
```diff
@@ -0,0 +1,31 @@
...
+    <<: *common
+    command: /bin/bash -cl "mvn clean package"
+
+  shell:
```
tomerd: is this for dev use case?

docker/Dockerfile.centos
```diff
@@ -42,13 +41,18 @@ RUN wget -q http://storage.googleapis.com/golang/go$GO_VERSION.linux-amd64.tar.g
 
 RUN wget -q http://archive.apache.org/dist/maven/maven-3/$MAVEN_VERSION/binaries/apache-maven-$MAVEN_VERSION-bin.tar.gz && tar xfz apache-maven-$MAVEN_VERSION-bin.tar.gz && mv apache-maven-$MAVEN_VERSION /opt/ && echo 'PATH=/opt/apache-maven-$MAVEN_VERSION/bin:$PATH' >> ~/.bashrc
```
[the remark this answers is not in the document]
normanmaurer: nope... maybe I should port the wrapper first :)

docker/docker-compose.centos.yaml
```diff
@@ -0,0 +1,31 @@
...
+      context: .
+      dockerfile: Dockerfile.centos
+
+  common: &common
+    image: netty-tcnative-centos:default
+    depends_on: [runtime-setup]
+    volumes:
+      - ~/.ssh:/root/.ssh
+      - ~/.gnupg:/root/.gnupg
+      - ..:/code
```
tomerd: also bind .m2? or at least settings.xml?

docker/docker-compose.centos.yaml
normanmaurer: yeah... when doing a release.

docker/docker-compose.centos.yaml
normanmaurer: @tomerd I only wanted to bind it when using `shell` as I want to refetch maven in case we install something during build.

docker/docker-compose.centos.yaml
normanmaurer: not sure... I thought build would be more fitting as it is really compiling everything and package it.

docker/docker-compose.centos.yaml
normanmaurer: binding `settings.xml` sounds good tho... let me do this

docker/Dockerfile.centos
normanmaurer: nope...

docker/Dockerfile.centos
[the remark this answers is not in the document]
normanmaurer: done

normanmaurer:
@ejona86 fair enough, I will update it. Basically it allows easy configuration of running tests etc with different java versions . The ultimate goal will be to use the same docker config for the CI as well once everything is in place. This will allow people to locally basically have the same setup as on the CI

Issue #353, opened 2018-06-11
Possible memory leak in cert_requested()?


rgs1:
We are seeing a pretty fast growing memory leak when using `clientAuth(ClientAuth.REQUIRE)` in the SSLServerContext.

A very simple self-contained repro (client & server included) can be found here:

https://github.com/rgs1/tlsleak

After looking around for a while, we found that we are leaking certificates (by inspecting the memory of the repro process). It's unclear which library is triggering the leak.

We came up with a very simple — albeit probably incorrect — patch that fixes this for us:

https://github.com/rgs1/netty-tcnative/commit/00d20713924f34ac683e96cb2f5ef868593ed61f

This works for the repro from above. We are also in the process of testing it more exhaustively in canary & production environments, while we keep looking for the definitive root cause of this leak.

We are using an old-ish version of Finagle (6.41), which means that we are pulling in Netty 3 and 4 — perhaps there are known issues when doing this? Maybe that causes an interaciton in which the first certificate in the chain gets an extra ref up?

Has anyone seen something similar?

cc @fishcakez @jogo

normanmaurer:
Interesting... will need to look into it. So the leak is happen with tcnative used by Netty 4 ? Or is it with Netty 3

fishcakez:
@normanmaurer we are using netty 3 but we use netty 4's ssl handler with tcnative 2 to create the SSLEngine to be used by netty 3. However I believe we also tried the latest versions so that we were using netty4 to handle the channels (so no netty3) and still experienced the leak. If needed we can provide a version of the repro that uses the latest of all dependencies (netty/tcnative/finagle).

normanmaurer:
That would be awesome

normanmaurer:
@rgs1 actually after looking into the code a bit with @Scottmitch we think your fix is correct (doh!). Can you submit a PR ?

normanmaurer:
Please also sign our icla:

https://netty.io/s/icla

And check:

https://netty.io/wiki/writing-a-commit-message.html

Scottmitch:
+1. Looks like you found a leak. If we want to submit a PR also consider removing the `cert` [1] local variable and just directly assigning `*x509Out = sk_X509_value(chain, 0);` [2]

[1] https://github.com/netty/netty-tcnative/blob/master/openssl-dynamic/src/main/c/sslcontext.c#L1499
[...]

PR #354, opened 2018-06-11
Fix memory leak in cert_requested()

rgs1:
Motivation:

There is a memory leak in the first certificate of the chain parsed in
cert_requested(). How to reproduce this is described in #353. It's
possible that this was introduced by 0572eabf06.

Modifications:

Remove the extra ref up on the first certificate of the chain.

Result:

Fixes https://github.com/netty/netty-tcnative/issues/353


carl-mastrangelo:
I am almost certainly not the right person to review SSL code!   cc: @flooey

normanmaurer:
@carl-mastrangelo lol why not 🤣 Pinged @flooey and @nmittler on the original issue already.

PR #355, opened 2018-06-12
Remove unused function tcn_X509_up_ref

rgs1:
Motivation:

This function isn't used, it is a leftover from the fix for #353.

Modifications:

Remove the unused function.

Result:

No more dead code.


normanmaurer:
@rgs1 thanks a lot again... as you may have noticed we love PRs ;)

PR #357, opened 2018-06-18
Add explict cast to  (void*) as needed when compiling against OpenSSL…

normanmaurer:
… 1.1.1

Motivation:

OpenSSL 1.1.1 expect to receive an argument of type  void*, so the build currently fails as we treat warnings as errors during compilation.

Modifications:

Add cast to  (void*).

Result:

Build also passes when compiling against OpenSSL 1.1.1


normanmaurer:
Preparation for OpenSSL 1.1.1 support

PR #360, opened 2018-06-19
Allow to set key material without reparsing certificate chain / priva…

normanmaurer:
…te key.

Motiviation:

During profiling one of the bottlenecks that could be observed was that we always needed to re-parse the certificate chain / private key on each handshake. This is kind of wasteful and it may be prefered to just use the same instance all the time.

Modification:

- Add two new methods that allow to directly set the keymaterial without re-parsing it all the time.
- Change the CertificateRequestCallback to not return the KeyMaterial but to have the user directly call a method to set the keymaterial. This allows for better life-time management of the keymaterial, as it may be re-used by the user for performance reasons.

Result:

More flexible and performant way of handling keymaterial (cert chain + private key).

### DIFF
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -1777,6 +1777,141 @@ TCN_IMPLEMENT_CALL(void, SSL, freeX509Chain)(TCN_STDARGS, jlong x509Chain)
     sk_X509_pop_free(chain, X509_free);
 }
 
+TCN_IMPLEMENT_CALL(void, SSL, setKeyMaterialServerSide)(TCN_STDARGS, jlong ssl, jlong chain, jlong key)
+{
+#if OPENSSL_VERSION_NUMBER < 0x10002000L || defined(LIBRESSL_VERSION_NUMBER)
+    // Only supported on openssl 1.0.2+
+    tcn_Throw(e, "Not supported");
+#else
+    SSL *ssl_ = J2P(ssl, SSL *);
+
+    TCN_CHECK_NULL(ssl_, ssl, /* void */);
+
+    EVP_PKEY* pkey = J2P(key, EVP_PKEY *);
+    STACK_OF(X509) *cchain = J2P(chain, STACK_OF(X509) *);
+    int numCerts;
+    char err[ERR_LEN];
+    int i;
+
+    UNREFERENCED(o);
+    TCN_ASSERT(ssl != NULL);
+
+    TCN_CHECK_NULL(cchain, chain, /* void */);
+
+    numCerts = sk_X509_num(cchain);
+
+    // SSL_use_certificate will increment the reference count of the cert.
+    if (numCerts <= 0 || SSL_use_certificate(ssl_, sk_X509_value(cchain, 0)) <= 0) {
+        ERR_error_string_n(ERR_get_error(), err, ERR_LEN);
+        ERR_clear_error();
+        tcn_Throw(e, "Error setting certificate (%s)", err);
+        return;
+    }
+
+    if (pkey != NULL) {
+        // SSL_use_PrivateKey will increment the reference count of the key.
+        if (SSL_use_PrivateKey(ssl_, pkey) <= 0) {
+            ERR_error_string_n(ERR_get_error(), err, ERR_LEN);
+            ERR_clear_error();
+            tcn_Throw(e, "Error setting private key (%s)", err);
+            return;
+        }
+        if (SSL_check_private_key(ssl_) <= 0) {
+            ERR_error_string_n(ERR_get_error(), err, ERR_LEN);
+            ERR_clear_error();
+
+            tcn_Throw(e, "Private key does not match the certificate public key (%s)", err);
+            return;
+        }
+    }
+
+    // The first cert was loaded via SSL_use_certificate so skip it.
+    for (i = 1; i < numCerts; ++i) {
+
+        // SSL_add1_chain_cert will increment the reference count of the cert.
+        if (SSL_add1_chain_cert(ssl_, sk_X509_value(cchain, i)) != 1) {
+            ERR_error_string_n(ERR_get_error(), err, ERR_LEN);
+            ERR_clear_error();
+
+            tcn_Throw(e, "Could not add certificate to chain (%s)", err);
+            return;
+        }
+    }
+#endif
+}
+
+TCN_IMPLEMENT_CALL(void, SSL, setKeyMaterialClientSide)(TCN_STDARGS, jlong ssl, jlong certOut, jlong keyOut, jlong chain, jlong key)
+{
+#if OPENSSL_VERSION_NUMBER < 0x10002000L || defined(LIBRESSL_VERSION_NUMBER)
+    // Only supported on openssl 1.0.2+
+    tcn_Throw(e, "Not supported");
+#else
+    SSL *ssl_ = J2P(ssl, SSL *);
+
+    TCN_CHECK_NULL(ssl_, ssl, /* void */);
+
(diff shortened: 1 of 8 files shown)


normanmaurer:
@Scottmitch @trustin @ejona86 @mosesn  PTAL.... This is the patch for tcnative that is needed to allow caching the key / chain as during profiling we saw a huge amount of time spend to parse these on every handshake. 

I will shortly post another PR on netty that uses these APIs. Just to give you an idea when using caching the handshake times improved by 30 % in some benchmarks.

openssl-dynamic/src/main/java/io/netty/internal/tcnative/CertificateRequestedCallback.java
```diff
@@ -38,47 +38,17 @@
...
-        /**
-         * Returns a x509 chain ({@code STACK_OF(X509)} pointer)
-         *
-         * @return thex509 chain ({@code STACK_OF(X509)} pointer)
-         */
-        public final long certificateChain() {
-            return certificateChain;
-        }
-    }
+    void requested(long ssl, long certOut, long keyOut, byte[] keyTypeBytes, byte[][] asn1DerEncodedPrincipals)
```
Scottmitch: this is a breaking change ... fine for the way we currently do things in Netty but we should do a major version rev + namespace change for this module in future Netty.

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1777,6 +1777,142 @@ TCN_IMPLEMENT_CALL(void, SSL, freeX509Chain)(TCN_STDARGS, jlong x509Chain)
...
+    SSL *ssl_ = J2P(ssl, SSL *);
+
+    TCN_CHECK_NULL(ssl_, ssl, /* void */);
+
+    EVP_PKEY* pkey = J2P(key, EVP_PKEY *);
+    STACK_OF(X509) *cchain = J2P(chain, STACK_OF(X509) *);
+    int numCerts;
+    X509* x509 = NULL;
+    char err[ERR_LEN];
+    int n;
```
Scottmitch: nit: we use `i` for an increment variable in the other method ... consider being consistent.

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1777,6 +1777,142 @@ TCN_IMPLEMENT_CALL(void, SSL, freeX509Chain)(TCN_STDARGS, jlong x509Chain)
...
+        return;
+    }
+    *x509Out = x509;
+
+    if (tcn_EVP_PKEY_up_ref(pkey) < 1) {
+        // We could not increment the reference count
+        ERR_error_string_n(ERR_get_error(), err, ERR_LEN);
+        ERR_clear_error();
+        tcn_Throw(e, "Could not add private key (%s)", err);
+        return;
```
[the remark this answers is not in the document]
normanmaurer: @Scottmitch I don't think so as if we keep x509out here it should be freed by openssl later on. I will double-check in the source of openssl tho to ensure this is the case.

openssl-dynamic/src/main/java/io/netty/internal/tcnative/CertificateRequestedCallback.java
normanmaurer: agree... But as you said this is fine for now.

openssl-dynamic/src/main/c/ssl.c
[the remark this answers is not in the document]
normanmaurer: @Scottmitch I reviewed the code in openssl and you are right... I adjusted the code to account for it.

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1777,6 +1777,141 @@ TCN_IMPLEMENT_CALL(void, SSL, freeX509Chain)(TCN_STDARGS, jlong x509Chain)
...
+        // SSL_add1_chain_cert will increment the reference count of the cert.
+        if (SSL_add1_chain_cert(ssl_, sk_X509_value(cchain, i)) != 1) {
+            ERR_error_string_n(ERR_get_error(), err, ERR_LEN);
+            ERR_clear_error();
+
+            tcn_Throw(e, "Could not add certificate to chain (%s)", err);
+            return;
+        }
+    }
+    #endif
```
carl-mastrangelo: nit: `endif` doesnt have same whitespace alignment.

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1777,6 +1777,141 @@ TCN_IMPLEMENT_CALL(void, SSL, freeX509Chain)(TCN_STDARGS, jlong x509Chain)
...
+            tcn_Throw(e, "Private key does not match the certificate public key (%s)", err);
+            return;
+        }
+    }
+
+    // The first cert was loaded via SSL_use_certificate so skip it.
+    for (i = 1; i < numCerts; ++i) {
+
+        // SSL_add1_chain_cert will increment the reference count of the cert.
+        if (SSL_add1_chain_cert(ssl_, sk_X509_value(cchain, i)) != 1) {
```
carl-mastrangelo: dumb question: how does this unref if there is an error?

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -1519,59 +1510,21 @@ static int cert_requested(SSL* ssl, X509** x509Out, EVP_PKEY** pkeyOut) {
     issuers = principalBytes(e,  SSL_get_client_CA_list(ssl));
 
     // Execute the java callback
-    keyMaterial = (*e)->CallObjectMethod(e, c->cert_requested_callback, c->cert_requested_callback_method, P2J(ssl), keyTypes, issuers);
+    (*e)->CallObjectMethod(e, c->cert_requested_callback, c->cert_requested_callback_method,
```
carl-mastrangelo: Should this be CallVoidMethod?

openssl-dynamic/src/main/c/ssl.c
normanmaurer: if there is an error the reference count of the cert is not incremented. All the previous added certificates will be part of the SSL lifetime which means once its teared down the reference count of these will be decremented and so eventually freed.

normanmaurer:
@carl-mastrangelo addressed all your comments

normanmaurer:
@Scottmitch @carl-mastrangelo @ejona86 @mosesn @bryce-anderson I would like to merge this and release a new version and then merge https://github.com/netty/netty/pull/8036 and include it in the next netty release as I see 30 % perf (even more if the chain is large) wins during handshake. CPU usage goes down significantly as well.

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -1587,7 +1540,8 @@ TCN_IMPLEMENT_CALL(void, SSLContext, setCertRequestedCallback)(TCN_STDARGS, jlon
         SSL_CTX_set_client_cert_cb(c->ctx, NULL);
     } else {
         jclass callback_class = (*e)->GetObjectClass(e, callback);
-        jmethodID method = (*e)->GetMethodID(e, callback_class, "requested", keyMaterialRequestedMethodSignature);
+        jmethodID method = (*e)->GetMethodID(e, callback_class, "requested","(JJJ[B[[B)V");
```
carl-mastrangelo: nit: space after the third argument

carl-mastrangelo: Do you need to check if there was an exception after the GetObjectClass call?

normanmaurer: I dont think so (also note this did not change):
https://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/functions.html

At worse we can do a followup.

PR #362, opened 2018-06-21
Support hostname validation also on OpenSSL versions < 1.0.2

normanmaurer:
Motivation:

We can also support hostname validation prior OpenSSL 1.0.2

Modifications:

Rework code to use different code path for different versions of OpenSSL / LibreSSL / BoringSSL so it always works.

Result:

Hostname validation is always supported.


normanmaurer:
This one is not correct... closing.

Issue #359, opened 2018-06-19
Don't strip libs to make debugging / profiling easier.


normanmaurer:
As of today we typically strip our static libs (like boringssl-static). I think this was mainly done in the past to make these smaller. I wonder if we should better not do this as this really makes it hard to debug / profile code once it ends up in a native method call. The only way how to do this atm is to have the user compile his own version and skip striping which I think is problematic as for example the performance problems will not show up directly again on redeployment.

normanmaurer:
@Scottmitch @mosesn @bryce-anderson @ejona86 @carl-mastrangelo I would like to get you opinion here.

Scottmitch:
Assuming there is negligible performance/memory impact this sgtm. If the additional disk consumption becomes a problem in the future:

- folks can always build their own artifact with their desired options
- we can publish both versions (with and without symbols) if demand is high enough

mosesn:
Yeah, seems reasonable to me.  Do we know what the difference in lib size will be?

mosesn:
3MB isn't too big :shipit:

bryce-anderson:
sgtm. 3 MB is a drop in the bucket for the typical JVM app.

ejona86:
I'd prefer for it to remain stripped:
* grpc does use it on clients, clients that may running on some sort of desktop machine (not a server)
* size is a proxy for the amount of code. For example, initially I was quite scared of tcnative when I noticed the large sizes, before I noticed the lack of stripping. Users who aren't aware of strip will likely think tcnative is a bit of a behemoth (3 MB is 6x the size of libssl.so!)
* if you're going to include debug information for Linux, would you include it for OS X and Windows (pdb) as well? Earlier, only Linux had debug information. The uber jar grows quite a bit if you include debug information for all platforms and I think that would discourage the use of the uber jar
* "[split debug info](https://sourceware.org/gdb/onlinedocs/gdb/Separate-Debug-Files.html)" is quite common. With split debug you still ship a stripped binary, but you save the debug information in a separate file that can be "installed" separately or be used by developers to symbolize a backtrace. Virtually every Linux distribution uses this method for debug information, as well as ISVs like Firefox and Chrome, and all Visual Studio builds

[...]

PR #363, opened 2018-06-22
We should only strip debug symbols

normanmaurer:
Motivation:

At the moment we completely strip the netty-tcnative-boringssl-static lib which makes it impossible to profile / debug the native code. We should limit the stripping to debug symbols to still be able to see at least the functions that are colled.

Modifications:

Use strip --strip-debug on linux.

Result:

Fixes https://github.com/netty/netty-tcnative/issues/359. The lib is only ca 100k bigger then before.


normanmaurer:
@carl-mastrangelo did you just volunteer 🤣  ?

ejona86:
@carl-mastrangelo, :nauseated_face:

PR #365, opened 2018-06-22
We should use weak linking if possible to detect if setHostNameValida…

normanmaurer:
…tion is usable

Motivation:

We did use classic ifdef blocks to detect on compile time if all functions are provided by the used openssl version that are needed for hostname validation. This is problematic as we compile on very old debian / centos versions to make the lib compatible with a wide range of dists.

Modifications:

Use weak linking for all functions needed for setHostNameValidation when GCC is used.

Result:

Be able to use setHostNameValidation on linux when the runtime version of OpenSSL provided these functions no matter which version was used during compilation.


normanmaurer:
Verified this locally... This works as expected 🎉

PR #367, opened 2018-06-26
Allow to set FIPS mode if compiled with FIPS support.

normanmaurer:
Motivation:

We had some users that are interested to build their own netty-tcnative artifacts and use FIPS. We should allow to enable it.

Modifications:

Allow to change the fips mode if its compiled with support for it.

Result:

Possible to use netty-tcnative with FIPS.


normanmaurer:
Related to: 

https://groups.google.com/forum/#!topic/netty/Fv-Rd3BinaU

luengnat:
anything holding this off?

Issue #273, opened 2017-05-17
Use weak linking to enable 1.0.2 features for dynamically linked released artifacts


Scottmitch:
Currently we compile the dynamically linked artifacts against OpenSSL 1.0.1, and we have compile time flags which limit the available feature set. We should consider building against OpenSSL 1.0.2 and/or using weak linking so that folks that link with a newer version of OpenSSL at runtime can use these features supported by the newer version.

Scottmitch:
Related:
https://github.com/netty/netty-tcnative/issues/265#issuecomment-297593386
https://github.com/netty/netty-tcnative/issues/269

sebtomba:
Will you include this in release `2.0.9`?

normanmaurer:
No idea yet if I have time for this ... contributions welcome

normanmaurer:
https://github.com/netty/netty-tcnative/pull/368 should be last bit needed to close this 🎉

PR #368, opened 2018-06-26
Support KeyManagerFactory if compiled against OpenSSL < 1.0.2 but usi…

normanmaurer:
…ng OpenSSL >= 1.0.2 at runtime.

Motivation:

On CentOS and Debian we compile against old versions of OpenSSL to archive maximal compatibility. Because of this we should ensure we can make use of features needed to support KeyManagerFactory when a newer version of OpenSSL is used during runtime.

Modifications:

Check on runtime the used OpenSSL version and based on this allow to use the function calls or not.

Result:

Fixes https://github.com/netty/netty-tcnative/issues/273.

### DIFF
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -664,7 +664,7 @@ TCN_IMPLEMENT_CALL(jint, SSL, initialize)(TCN_STDARGS, jstring engine)
     }
 
 #if OPENSSL_VERSION_NUMBER < 0x10100000L
-    if (SSLeay() < 0x0090700L) {
+    if (OpenSSL_version_num() < 0x0090700L) {
         TCN_FREE_CSTRING(engine);
         tcn_ThrowAPRException(e, APR_EINVAL);
         ssl_initialized = 0;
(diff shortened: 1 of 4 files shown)


openssl-dynamic/src/main/c/ssl.c
```diff
@@ -664,7 +664,7 @@ TCN_IMPLEMENT_CALL(jint, SSL, initialize)(TCN_STDARGS, jstring engine)
     }
 
 #if OPENSSL_VERSION_NUMBER < 0x10100000L
-    if (SSLeay() < 0x0090700L) {
+    if (OpenSSL_version_num() < 0x0090700L) {
```
normanmaurer: FYI: OpenSSL_version_num is the same as `SSLeay` but the "new API". If an old version of OpenSSL is used we delegate internally.

normanmaurer:
Verified locally by build on debian 7 and run on recent version of ubuntu. Works now as expected 🎉

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1847,10 +1854,16 @@ TCN_IMPLEMENT_CALL(void, SSL, setKeyMaterialServerSide)(TCN_STDARGS, jlong ssl,
 
 TCN_IMPLEMENT_CALL(void, SSL, setKeyMaterialClientSide)(TCN_STDARGS, jlong ssl, jlong certOut, jlong keyOut, jlong chain, jlong key)
 {
-#if OPENSSL_VERSION_NUMBER < 0x10002000L || defined(LIBRESSL_VERSION_NUMBER)
-    // Only supported on openssl 1.0.2+
-    tcn_Throw(e, "Not supported");
+#if defined(LIBRESSL_VERSION_NUMBER)
+    tcn_Throw(e, "Not supported with LibreSSL");
 #else
+#ifndef OPENSSL_IS_BORINGSSL
```
carl-mastrangelo: this is somewhat confusing.  You might consider putting in a comment which `endif` is associated.  e.g.

```
#ifndef OPENSSL_IS_BORINGSSL
...
#endif // OPENSSL_IS_BORINGSSL
```

normanmaurer: done

PR #369, opened 2018-06-28
We should disable hostname validation if an empty host or null is bas…

normanmaurer:
…ed into SSL.setHostNameValidation(...)

Motivation:

At the moment its not possible to disalbe hostname validation again when it was enabled before. We should allow this as we need to support this in netty.

Modifications:

Correctly handle the case of hostname been null or empty.

Result:

Be able to disable hostname validation again.

### DIFF
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -1567,27 +1567,34 @@ TCN_IMPLEMENT_CALL(void, SSL, setHostNameValidation)(TCN_STDARGS, jlong ssl, jin
     }
 #endif
 
-
 #if (OPENSSL_VERSION_NUMBER >= 0x10002000L && !defined(LIBRESSL_VERSION_NUMBER)) || LIBRESSL_VERSION_NUMBER >= 0x2060000fL || defined(OPENSSL_IS_BORINGSSL) || defined(__GNUC__) || defined(__GNUG__)
+    // If NULL is used we need to pass it down to X509_VERIFY_PARAM_set1_host as it will
+    // disable hostname validation again.
+    jsize hostnameLen;
+    const char *hostname;
     if (hostnameString == NULL) {
-        return;
+        hostnameLen = 0;
+        hostname = NULL;
+    } else {
+        hostnameLen = (*e)->GetStringUTFLength(e, hostnameString);
+        if (hostnameLen == 0) {
+            // If an empty hostname is used we will just return.
+            return;
+        }
+        hostname = (*e)->GetStringUTFChars(e, hostnameString, JNI_FALSE);
     }
+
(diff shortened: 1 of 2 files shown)


ejona86:
For the record, this is an intentional reversion of an intentional behavior change introduced in #335.

normanmaurer:
@ejona86 good Point... let me just use null for that and just ignore an empty string

normanmaurer:
@ejona86 can you check again... maybe also @agl :)

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1567,27 +1567,34 @@ TCN_IMPLEMENT_CALL(void, SSL, setHostNameValidation)(TCN_STDARGS, jlong ssl, jin
...
-        return;
+        hostnameLen = 0;
+        hostname = NULL;
+    } else {
+        hostnameLen = (*e)->GetStringUTFLength(e, hostnameString);
+        if (hostnameLen == 0) {
+            // If an empty hostname is used we will just return.
+            return;
+        }
+        hostname = (*e)->GetStringUTFChars(e, hostnameString, 0);
```
carl-mastrangelo: That final param should be a `NULL` (or JNI_FALSE)

normanmaurer:
@carl-mastrangelo fixed

agl:
No objections to this, but it won't work with BoringSSL: it rejects setting empty names in case, in the translation from C++ strings to C, a NUL slips in.

Also 72b77dacd205fd01dbef0702fbf69fc319a4ef39 seems to have made trying to set an empty sting (as opposed to a null string) a non-error, which may not have been intended.

normanmaurer:
@agl Please re-check bd2fd4e ... I am correct that NULL will work with boringSSL ?

agl:
See [this function in the source](https://boringssl.googlesource.com/boringssl/+/master/crypto/x509/x509_vpm.c#87). It's still very OpenSSL-based code, so not as clean as I would like, but configuring a hostname to verify and then changing your mind isn't supported. (Additionally, it causes any future X.509 verification with that parameter object to fail in case the error return value isn't caught.)

(I'm unclear why Netty wants to support this. Is there some Java code that's behaving like this?)

normanmaurer:
@agl yeah in java you can enable / disable it by passing in `SSLParameters` to the `SSLEngine`.

agl:
> @agl yeah in java you can enable / disable it by passing in SSLParameters to the SSLEngine.

And you can change your mind and update an SSLEngine with new SSLParameters after the fact?

The BoringSSL behaviour was designed based on real issues that we've seen in code so I'm quite hesitant to relax it. Is there a point where the SSLParameters for a connection are fixed (say, once some "connect" function is called) and the final answer can be applied then?

normanmaurer:
@agl any more comments ?

PR #370, opened 2018-06-28
Cleanup some ifdefs to add support for some features when LibreSSL is…

normanmaurer:
… used.

Motivation:

In newer versions of LibreSSL some functions are now included which allows us to support for example hostname verification. Also removed some special handling for BoringSSL which is not needed anymore.

Modifications:

- Cleanup ifdefs for LibreSSL and some for BoringSSL.

Result:

Support more features when using recent version of LibreSSL and cleaner code.


normanmaurer:
@ejona86 yep 🤷‍♂️

PR #372, opened 2018-07-02
Use JNI_FALSE when calling GetStringUTFChars(...) and not 0

normanmaurer:
Motivation:

GetStringUTFChars(...) list JNI_FALSE as its argument in the docs, we should use it and not depend on the fact that 0 == JNI_FALSE.

Modifications:

Replace 0 by JNI_FALSE when using GetStringUTFChars.

Result:

Cleaner code.


normanmaurer:
This was brought up by @carl-mastrangelo in another PR. So fixed all cases :)

PR #376, opened 2018-08-06
pom.xml:adding option for aarch64 to build

ossdev07:
adding option linux-aarch_64 to create jar in aarch64

Signed-off-by: ossdev <ossdev@puresoftware.com>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -421,10 +421,10 @@
                   <rules>
                     <requireProperty>
                       <regexMessage>
-                        Release process must be performed on linux-x86_64.
+                        Release process must be performed on linux-X86_64 or linux-aarch_64.
                       </regexMessage>
                       <property>os.detected.classifier</property>
-                      <regex>^linux-x86_64$</regex>
+                      <regex>^(linux-x86_64|linux-aarch_64)$</regex>
                     </requireProperty>
                     <requireFilesContent>
                       <message>


ossdev07:
@normanmaurer can you please have a look on this change for aarch64 platform

normanmaurer:
@ossdev07 I wonder why this is needed... are you running `mvn release:*` by yourself ?

ossdev07:
@normanmaurer to build netty-transport-native-epol on aarch64l i need to build netty-tcnative  for aarch64 platform that's it is there
will you please check it ..

normanmaurer:
@ossdev07 I am still confused why you need this change tho.. The restrict rule should only apply during `mvn release:*` so you should already be able to just build it via `mvn clean package` without the change. Can you elaborate ?

ossdev07:
@normanmaurer can we use maven plugins in transport-native-epoll to generate aarch64 stuff in x64 platform in same jar as used in netty-tcnative/openssl-dynamic (build-native-lib) . in my opinion this solution will solve the problem of releasing for different architecture.. if it suits you i can raise pull  request to check the changes.

johnou:
@ossdev07 ptal https://github.com/netty/netty-tcnative/pull/517

normanmaurer:
Lets close this in favour of https://github.com/netty/netty-tcnative/pull/517

PR #378, opened 2018-08-14
Call ENGINE_load_builtin_engines() when a custom engine was specified

normanmaurer:
Motivation:

We allow a user to specify which openssl engine should be used. To make it easier and allow the user to just specify OPENSSL_ENGINES=/path/to/the/engines we should call ENGINE_load_builtin_engines() before we try to lookup the engine per id.

Modifications:

Call ENGINE_load_builtin_engines() before trying to lookup the engine and also ensure we call ENGINE_cleanup() when needed.

Result:

Easier to use a custom openssl engine.

### DIFF
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -439,9 +439,13 @@ static apr_status_t ssl_init_cleanup(void *data)
 #endif
     /* Corresponds to SSL_library_init: */
     EVP_cleanup();
-#if HAVE_ENGINE_LOAD_BUILTIN_ENGINES
+
+    // In case we loaded any engine we should also call cleanup. This is especialy important in openssl < 1.1.
+#ifndef OPENSSL_IS_BORINGSSL
+    // This is deprecated since openssl 1.1 but does not exist at all in BoringSSL.
     ENGINE_cleanup();
 #endif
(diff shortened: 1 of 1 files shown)


openssl-dynamic/src/main/c/ssl.c
```diff
@@ -439,9 +439,13 @@ static apr_status_t ssl_init_cleanup(void *data)
 #endif
     /* Corresponds to SSL_library_init: */
     EVP_cleanup();
-#if HAVE_ENGINE_LOAD_BUILTIN_ENGINES
+
+    // In case we loaded any engine we should also call cleanup. This is especialy important in openssl < 1.1.
+#ifndef OPENSSL_IS_BORINGSSL
+    // This is deprecated since openssl 1.1 but does not exist at all in BoringSSL.
     ENGINE_cleanup();
```
ejona86: Is this safe to call even if we didn't call ENGINE_load_builtin_engines()?

normanmaurer: Yes ... Just checked the source

normanmaurer:
@carl-mastrangelo not easily as you will need to write your on engine implementation and compile against the correct openssl version etc. Let me just assure you I tested it and it works for now ;)

PR #380, opened 2018-08-16
Allow to load a private key from the OpenSSL engine.

normanmaurer:
Motivation:

Its possible to plugin different behavior into OpenSSL by providing a custom OpenSSL engine. When doing so it's possible to load a private key from the engine directly.

Modifications:

Add new method which allows to load a private key from the OpenSSL engine directly and so can be used later on.

Result:

Be able to implement custom loading / store mechanism of private keys via a custom OpenSSL engine.

### DIFF
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -31,6 +31,11 @@
 
 #include <stdbool.h>
 #include <openssl/bio.h>
+
+#ifndef OPENSSL_NO_ENGINE
+#include <openssl/ui.h>
+#endif // OPENSSL_NO_ENGINE
+
 #include "tcn.h"
 #include "apr_file_io.h"
 #include "apr_thread_mutex.h"
@@ -43,8 +48,12 @@
 static int ssl_initialized = 0;
 extern apr_pool_t *tcn_global_pool;
 
-ENGINE *tcn_ssl_engine = NULL;
 void *SSL_temp_keys[SSL_TMP_KEY_MAX];
+
+#ifndef OPENSSL_NO_ENGINE
+static ENGINE *tcn_ssl_engine = NULL;
+static UI_METHOD *ui_method = NULL;
+#endif // OPENSSL_NO_ENGINE
 
 /* Global reference to the pool used by the dynamic mutexes */
 static apr_pool_t *dynlockpool = NULL;
@@ -334,6 +343,49 @@ static long bio_java_bytebuffer_ctrl(BIO* bio, int cmd, long num, void* ptr) {
     }
 }
 
+// This code is based on libcurl:
+// https://github.com/curl/curl/blob/curl-7_61_0/lib/vtls/openssl.c#L521
+#ifndef OPENSSL_NO_ENGINE
+/*
+ * Supply default password to the engine user interface conversation.
+ * The password is passed by OpenSSL engine from ENGINE_load_private_key()
+ * last argument to the ui and can be obtained by UI_get0_user_data(ui) here.
+ */
+static int ssl_ui_reader(UI *ui, UI_STRING *uis)
+{
+    const char *password;
+    switch (UI_get_string_type(uis)) {
+    case UIT_PROMPT:
+    case UIT_VERIFY:
+        password = (const char *) UI_get0_user_data(ui);
+        if (password != NULL && (UI_get_input_flags(uis) & UI_INPUT_FLAG_DEFAULT_PWD) != 0) {
+            UI_set_result(ui, uis, password);
+            return 1;
+        }
+        // fall-through
+    default:
+        return (UI_method_get_reader(UI_OpenSSL()))(ui, uis);
+  }
+}
+
+/*
+ * Suppress interactive request for a default password if available.
+ */
+static int ssl_ui_writer(UI *ui, UI_STRING *uis)
+{
+    switch(UI_get_string_type(uis)) {
+    case UIT_PROMPT:
+    case UIT_VERIFY:
+        if (UI_get0_user_data(ui) != NULL && (UI_get_input_flags(uis) & UI_INPUT_FLAG_DEFAULT_PWD) != 0) {
+            return 1;
+        }
+        // fall-through
+    default:
+        return (UI_method_get_writer(UI_OpenSSL()))(ui, uis);
+  }
+}
+#endif // OPENSSL_NO_ENGINE
+
 TCN_IMPLEMENT_CALL(jint, SSL, bioLengthByteBuffer)(TCN_STDARGS, jlong bioAddress) {
     BIO* bio = J2P(bioAddress, BIO*);
     struct TCN_bio_bytebuffer* bioUserData;
@@ -440,12 +492,6 @@ static apr_status_t ssl_init_cleanup(void *data)
     /* Corresponds to SSL_library_init: */
     EVP_cleanup();
 
-    // In case we loaded any engine we should also call cleanup. This is especialy important in openssl < 1.1.
-#ifndef OPENSSL_IS_BORINGSSL
-    // This is deprecated since openssl 1.1 but does not exist at all in BoringSSL.
-    ENGINE_cleanup();
-#endif
-
 #if OPENSSL_VERSION_NUMBER >= 0x00907001
     CRYPTO_cleanup_all_ex_data();
 #endif
@@ -460,6 +506,25 @@ static apr_status_t ssl_init_cleanup(void *data)
 #ifdef OPENSSL_FIPS
      FIPS_mode_set(0);
 #endif
+
+#ifndef OPENSSL_NO_ENGINE
+     /* Free our "structural" reference. */
+     if (tcn_ssl_engine != NULL) {
+         ENGINE_free(tcn_ssl_engine);
+         tcn_ssl_engine = NULL;
+     }
+
+     if (ui_method != NULL) {
+         UI_destroy_method(ui_method);
+         ui_method = NULL;
+     }
+
+// In case we loaded any engine we should also call cleanup. This is especialy important in openssl < 1.1.
+#ifndef OPENSSL_IS_BORINGSSL
+    // This is deprecated since openssl 1.1 but does not exist at all in BoringSSL.
+    ENGINE_cleanup();
+#endif // OPENSSL_IS_BORINGSSL
+#endif // OPENSSL_NO_ENGINE
 
     /* Don't call ERR_free_strings here; ERR_load_*_strings only
      * actually load the error strings once per process due to static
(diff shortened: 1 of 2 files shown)


normanmaurer:
@carl-mastrangelo @ejona86 sadly this is again not easily testable :( I tested it locally tho.

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -725,6 +781,19 @@ TCN_IMPLEMENT_CALL(jint, SSL, initialize)(TCN_STDARGS, jstring engine)
             if (ee)
                 ENGINE_free(ee);
         }
+        if (ee != NULL) {
```
ejona86: This check seem strange, since `ee` will have been freed and it's possible for `ee != NULL` but `err == APR_ENOTIMPL`. Should this block be moved into the previous block?

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -334,6 +343,47 @@ static long bio_java_bytebuffer_ctrl(BIO* bio, int cmd, long num, void* ptr) {
...
+{
+    const char *password;
+    switch (UI_get_string_type(uis)) {
+    case UIT_PROMPT:
+    case UIT_VERIFY:
+        password = (const char *) UI_get0_user_data(ui);
+        if (password != NULL && (UI_get_input_flags(uis) & UI_INPUT_FLAG_DEFAULT_PWD) != 0) {
+            UI_set_result(ui, uis, password);
+            return 1;
+        }
```
ejona86: Add a comment that fallthrough is expected (make it clear there is a missing break on purpose)? Ditto in next function.

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -725,6 +781,19 @@ TCN_IMPLEMENT_CALL(jint, SSL, initialize)(TCN_STDARGS, jstring engine)
             if (ee)
```
Scottmitch: `ee != NULL` while you are here.

normanmaurer:
@ejona86 addressed.

openssl-dynamic/src/main/c/ssl.c
normanmaurer: done

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -334,6 +343,51 @@ static long bio_java_bytebuffer_ctrl(BIO* bio, int cmd, long num, void* ptr) {
...
+ * Supply default password to the engine user interface conversation.
+ * The password is passed by OpenSSL engine from ENGINE_load_private_key()
+ * last argument to the ui and can be obtained by UI_get0_user_data(ui) here.
+ */
+static int ssl_ui_reader(UI *ui, UI_STRING *uis)
+{
+    const char *password;
+    switch (UI_get_string_type(uis)) {
+    case UIT_PROMPT:
+        // fall-through
```
ejona86: FWIW, I was fine without a comment here because the fallthrough was obvious and normal.

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -721,10 +781,24 @@ TCN_IMPLEMENT_CALL(jint, SSL, initialize)(TCN_STDARGS, jstring engine)
                 if (!ENGINE_set_default(ee, ENGINE_METHOD_ALL))
                     err = APR_ENOTIMPL;
             }
-            /* Free our "structural" reference. */
-            if (ee)
+            if (ee != NULL) {
```
ejona86: The error flow of this is still hard to follow, because it is possible for `err == APR_ENOTIMPL` here. It seems like that isn't on purpose. Although it looks like it would work. Is that on purpose?

openssl-dynamic/src/main/c/ssl.c
normanmaurer: I guess it cant hurt :)

openssl-dynamic/src/main/c/ssl.c
normanmaurer: yeah let me just make it more clear.

normanmaurer:
@ejona86 done :)

normanmaurer:
@ejona86 also fixed a possible memory leak

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -46,6 +51,10 @@ extern apr_pool_t *tcn_global_pool;
 ENGINE *tcn_ssl_engine = NULL;
 void *SSL_temp_keys[SSL_TMP_KEY_MAX];
 
+#ifndef OPENSSL_NO_ENGINE
+UI_METHOD *ui_method = NULL;
```
Scottmitch: should this be defined as `static`?

openssl-dynamic/src/main/c/ssl.c
Scottmitch: (I'm also fine without fall through here bcz it is expected stacking of cases with no-code ... having the comment brings attention to an area where attention isn't necessary)

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -721,17 +789,29 @@ TCN_IMPLEMENT_CALL(jint, SSL, initialize)(TCN_STDARGS, jstring engine)
                 if (!ENGINE_set_default(ee, ENGINE_METHOD_ALL))
                     err = APR_ENOTIMPL;
             }
-            /* Free our "structural" reference. */
-            if (ee)
-                ENGINE_free(ee);
```
Scottmitch: looks like this was dropped ... are we now leaking `ee`?

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1711,6 +1791,37 @@ TCN_IMPLEMENT_CALL(void, SSL, setCertificateChainBio)(TCN_STDARGS, jlong ssl,
...
+TCN_IMPLEMENT_CALL(jlong, SSL, loadPrivateKeyFromEngine)(TCN_STDARGS, jstring keyId, jstring password)
+{
+#ifndef OPENSSL_NO_ENGINE
+    char err[ERR_LEN];
+    EVP_PKEY* pkey = NULL;
+
+    TCN_ALLOC_CSTRING(keyId);
+    TCN_ALLOC_CSTRING(password);
+
+    pkey = ENGINE_load_private_key(tcn_ssl_engine, ckeyId, ui_method, (void*) cpassword);
```
Scottmitch: `ckeyId` made me do a double take ... magical macros 😕

openssl-dynamic/src/main/c/ssl.c
normanmaurer: Can’t hurt... let me also make tcn_ssl_engine static while I am here

Scottmitch: (static will prevent the variable from being visible/exported from this file).

openssl-dynamic/src/main/c/ssl.c
normanmaurer: Nope what was done before was not really correct. We should only call free in the ssl_init_cleanup

normanmaurer: So it was moved there

openssl-dynamic/src/main/c/ssl.c
normanmaurer: Ok will remove

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -721,17 +787,29 @@ TCN_IMPLEMENT_CALL(jint, SSL, initialize)(TCN_STDARGS, jstring engine)
...
             }
-            /* Free our "structural" reference. */
-            if (ee)
-                ENGINE_free(ee);
+            if (err == APR_SUCCESS) {
+                // This code is based on libcurl:
+                // https://github.com/curl/curl/blob/curl-7_61_0/lib/vtls/openssl.c#L521
+                ui_method = UI_create_method((char *)"netty-tcnative user interface");
+                if (ui_method != NULL) {
+                    UI_method_set_opener(ui_method, UI_method_get_opener(UI_OpenSSL()));
```
carl-mastrangelo: I'm uncomfortable ignoring error codes from these setters.    Could you make them an assert 1 at least?

normanmaurer: @carl-mastrangelo sure thing...

normanmaurer: that said its actually `0` on success https://www.openssl.org/docs/man1.1.0/crypto/UI_method_set_opener.html

openssl-dynamic/src/main/c/ssl.c
Scottmitch: can you remove the `ee` intermediate variable then? it introduces the possibility of a leak and makes the ownership less clear.

openssl-dynamic/src/main/c/ssl.c
carl-mastrangelo: Well, dang, thats scary.  ui.h says the opposite:  https://github.com/openssl/openssl/blob/master/include/openssl/ui.h#L256

```
   All these functions are expected to return:
        0       on error.
        1       on success.
```

normanmaurer: I guess this is talking about the "functions" that you use as opener etc itself. not the setter.

normanmaurer:
@carl-mastrangelo PTAL again

openssl-dynamic/src/main/java/io/netty/internal/tcnative/SSL.java
```diff
@@ -598,6 +598,20 @@ public static native boolean setCipherSuites(long ssl, String ciphers)
     public static native void setCertificateBio(
             long ssl, long certBio, long keyBio, String password) throws Exception;
 
+    /**
+     * Load a private key from the used OpenSSL ENGINE via the
+     * <a href="https://www.openssl.org/docs/man1.1.0/crypto/ENGINE_load_private_key.html">ENGINE_load_private_key</a>
+     * function.
+     *
+     * <p>Be sure you understand how OpenSsl will behave with respect to reference counting!
```
carl-mastrangelo: This comment is ominous, I don't really know what I should do about it.  Maybe call `unloadPrivateKey` or something like that?  I don't see that method though.  What should a reader do?

normanmaurer: @carl-mastrangelo added some more docs :)

openssl-dynamic/src/main/c/ssl.c
normanmaurer: done

normanmaurer:
@Scottmitch also gave ok offline... will merge once the CI pass.

PR #381, opened 2018-08-17
rtdavies/fix openssl build win32

rtdavies:
Fix Windows 32-bit build. Similar to changes already applied for 64-bit.


normanmaurer:
@rtdavies thanks a lot!

PR #382, opened 2018-08-23
Ensure multiple shaded version of the same netty-tcnative artifact ca…

normanmaurer:
…n be loaded as long as the shaded prefix is different

Motivation:

We should support to load multiple shaded versions of the same netty-tcnative artifact as netty-tcnative is often used in multiple dependencies.

This is related to https://github.com/netty/netty/issues/7272.

Modifications:

- Use -fvisibility=hidden when compiling and use JNIEXPORT for things we really want to have exported
- Ensure fields are declared as static so these are not exported
- Ensure we pass the correct linker flags so functions of the static linked version of Boring|Libre|OpenSSL and APR are not visible

Result:

Be able to use multiple shaded versions of the same netty-tcnative artifact.

### DIFF
--- a/openssl-dynamic/src/main/native-package/m4/tcnative.m4
+++ b/openssl-dynamic/src/main/native-package/m4/tcnative.m4
@@ -276,6 +276,32 @@ dnl as the help string.
 AC_DEFUN([TCN_HELP_STRING],[ifelse(regexp(AC_ACVERSION, 2\.1), -1, AC_HELP_STRING($1,$2),[  ]$1 substr([                       ],len($1))$2)])dnl
 
 dnl
+dnl TCN_CHECK_STATIC
+dnl Will prepare more LDFLAGS that should be set to ensure we do not export any functions from the static compiled APR / OpenSSL libs.
+dnl
+AC_DEFUN([TCN_CHECK_STATIC],[
+    LD_FLAGS_STATIC=""
+
+    AC_ARG_WITH(static-libs,
+      [  --with-static-libs     The libraries we link against are static.],
+      [
+
+      case $host in
+      *-darwin*)
+          LD_FLAGS_STATIC="-Wl,-exported_symbol,_JNI_*"
+          ;;
+      *linux*)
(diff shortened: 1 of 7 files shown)


openssl-dynamic/src/main/native-package/m4/tcnative.m4
```diff
@@ -275,6 +275,32 @@ dnl Note: this define must be on one line so that it can be properly returned
 dnl as the help string.
 AC_DEFUN([TCN_HELP_STRING],[ifelse(regexp(AC_ACVERSION, 2\.1), -1, AC_HELP_STRING($1,$2),[  ]$1 substr([                       ],len($1))$2)])dnl
 
+dnl
+dnl TCN_CHECK_STATIC
+dnl Will prepare more LDFLAGS that should be set to ensure we not export any functions from the static compiled APR / OpenSSL libs.
+dnl
+AC_DEFUN([TCN_CHECK_STATIC],[
```
normanmaurer: I had to do this as an autoconf function as otherwise the flags would have been picked up during running `./configure` and so may fail the test to detect the linker / gcc as the filtered symbols did not exist there yet.

normanmaurer:
Also verified that you can still see all functions in a flame graph after the stripping etc :)

openssl-dynamic/src/main/native-package/m4/tcnative.m4
```diff
@@ -275,6 +275,32 @@ dnl Note: this define must be on one line so that it can be properly returned
 dnl as the help string.
 AC_DEFUN([TCN_HELP_STRING],[ifelse(regexp(AC_ACVERSION, 2\.1), -1, AC_HELP_STRING($1,$2),[  ]$1 substr([                       ],len($1))$2)])dnl
 
+dnl
+dnl TCN_CHECK_STATIC
+dnl Will prepare more LDFLAGS that should be set to ensure we not export any functions from the static compiled APR / OpenSSL libs.
```
Scottmitch: `we not` -> `we do not`

PR #383, opened 2018-08-24
Correctly support DH with all versions of OpenSSL / BoringSSL / LibreSSL

normanmaurer:
Motivation:

We did not support DH with some versions of OpenSSL / BoringSSL / LibreSSL.

Modifications:

- Remove some ifdefs etc and adjust the code to work with all of the SSL flavors we support.

Result:

Fixes https://github.com/netty/netty/issues/8165.

### DIFF
--- a/openssl-dynamic/src/main/c/ssl_private.h
+++ b/openssl-dynamic/src/main/c/ssl_private.h
@@ -64,6 +64,7 @@
 #include <openssl/rand.h>
 #include <openssl/x509v3.h>
 #include <openssl/hmac.h>
+#include <openssl/dh.h>
 
 #define ERR_LEN 256
 
--- a/openssl-dynamic/src/main/c/sslutils.c
+++ b/openssl-dynamic/src/main/c/sslutils.c
@@ -230,8 +230,6 @@ int tcn_SSL_password_callback(char *buf, int bufsiz, int verify,
     return (int)strlen(buf);
 }
 
-#if (OPENSSL_VERSION_NUMBER < 0x10100000L) || defined(OPENSSL_USE_DEPRECATED) || defined(LIBRESSL_VERSION_NUMBER)
-
 static unsigned char dh0512_p[]={
     0xD9,0xBA,0xBF,0xFD,0x69,0x38,0xC9,0x51,0x2D,0x19,0x37,0x39,
     0xD7,0x7D,0x7E,0x3E,0x25,0x58,0x55,0x94,0x90,0x60,0x93,0x7A,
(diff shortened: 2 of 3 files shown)


normanmaurer:
@rkapsi FYI... This should fix https://github.com/netty/netty/issues/8165

openssl-dynamic/src/main/c/ssl_private.h
```diff
@@ -64,6 +64,7 @@
 #include <openssl/rand.h>
 #include <openssl/x509v3.h>
 #include <openssl/hmac.h>
+#include <openssl/dh.h>
```
normanmaurer: This is needed now as in newer version of OpenSSL this is required (and on older version it cant harm).

openssl-dynamic/src/main/c/sslutils.c
```diff
@@ -329,37 +327,51 @@ static unsigned char dhxxx2_g[]={
...
         case SSL_TMP_KEY_DH_4096:
-            dh->p = BN_bin2bn(dh4096_p, sizeof(dh2048_p), NULL);
+            p = BN_bin2bn(dh4096_p, sizeof(dh2048_p), NULL);
         break;
     }
-    dh->g = BN_bin2bn(dhxxx2_g, sizeof(dhxxx2_g), NULL);
-    if ((dh->p == NULL) || (dh->g == NULL)) {
-        DH_free(dh);
-        return NULL;
+    if (p == NULL) {
```
ejona86: Could `p` be uninitialized here?

openssl-dynamic/src/main/c/sslutils.c
```diff
@@ -329,37 +327,51 @@ static unsigned char dhxxx2_g[]={
...
 #else
-static DH *get_dh(int idx)
-{
+        dh->p = p;
+        dh->g = g;
+#endif
+        return dh;
+    }
+error:
+    // All these functions can handle NULL as argument.
```
ejona86: But couldn't p and g be uninitialized?

normanmaurer: @ejona86 sorry I don't get the question... All the free methods can handle `NULL` so it should be fine, no ?

openssl-dynamic/src/main/c/sslutils.c
normanmaurer: yes `BN_bin2bn` will return `NULL` on error... https://www.openssl.org/docs/man1.1.0/crypto/BN_bin2bn.html

openssl-dynamic/src/main/c/sslutils.c
ejona86: Above:
```
BIGNUM *p, *g;
```

I'd expect `p` and `g` need to be set to `NULL` to initialize them.

openssl-dynamic/src/main/c/sslutils.c
ejona86: But there is no `default` case in the switch.

openssl-dynamic/src/main/c/sslutils.c
normanmaurer: ah yeah better be safe here... Thanks!

openssl-dynamic/src/main/c/sslutils.c
normanmaurer: good point... let me just init it above.

normanmaurer:
@ejona86 fixed... PTAL again

openssl-dynamic/src/main/c/sslutils.c
```diff
@@ -328,38 +326,52 @@ static unsigned char dhxxx2_g[]={
 
 static DH *get_dh(int idx)
 {
-    DH *dh;
-    if ((dh = DH_new()) == NULL)
+    DH *dh = NULL;
+    BIGNUM *p, *g = NULL;
```
ejona86: `p` is still uninitialized :smile:

normanmaurer:
@ejona86 stupid me 🤦‍♂️ ... fixed

openssl-dynamic/src/main/c/sslutils.c
```diff
@@ -328,38 +326,53 @@ static unsigned char dhxxx2_g[]={
 
 static DH *get_dh(int idx)
 {
-    DH *dh;
-    if ((dh = DH_new()) == NULL)
+    DH *dh = NULL;
+    BIGNUM *p;
```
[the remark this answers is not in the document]
normanmaurer: Yes... seems like I did not amend correctly

openssl-dynamic/src/main/c/sslutils.c
```diff
@@ -328,38 +326,53 @@ static unsigned char dhxxx2_g[]={
...
-            dh->p = BN_bin2bn(dh1024_p, sizeof(dh1024_p), NULL);
+            p = BN_bin2bn(dh1024_p, sizeof(dh1024_p), NULL);
         break;
         case SSL_TMP_KEY_DH_2048:
-            dh->p = BN_bin2bn(dh2048_p, sizeof(dh2048_p), NULL);
+            p = BN_bin2bn(dh2048_p, sizeof(dh2048_p), NULL);
         break;
         case SSL_TMP_KEY_DH_4096:
-            dh->p = BN_bin2bn(dh4096_p, sizeof(dh2048_p), NULL);
+            p = BN_bin2bn(dh4096_p, sizeof(dh2048_p), NULL);
```
johnou: @normanmaurer should be `sizeof(dh4096_p)` here

normanmaurer: yep... that said it should not matter :)

normanmaurer: Yep... good catch. This bug existed before btw.

openssl-dynamic/src/main/c/sslutils.c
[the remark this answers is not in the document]
johnou: GitHub still shows the null assignment is missing for p.

PR #384, opened 2018-08-29
We should call the UnLoad methods when we detect an error during call…

normanmaurer:
…ing OnLoad

Motivation:

We should ensure we call *UnLoad when we detect an error during calling *OnLoad and previous *OnLoad calls were succesfull.

Modifications:

Correctly call *UnLoad when needed.

Result:

More correct code and no leaks when an error happens during loading the native lib.

### DIFF
--- a/openssl-dynamic/src/main/c/jnilib.c
+++ b/openssl-dynamic/src/main/c/jnilib.c
@@ -291,27 +291,41 @@ static const jint method_table_size = sizeof(method_table) / sizeof(method_table
 // JNI Method Registration Table End
 
 static jint netty_internal_tcnative_Library_JNI_OnLoad(JNIEnv* env, const char* packagePrefix) {
+    int errorOnLoadCalled = 0;
+    int bufferOnLoadCalled = 0;
+    int jniMethodsOnLoadCalled = 0;
+    int sslOnLoadCalled = 0;
+    int contextOnLoadCalled = 0;
+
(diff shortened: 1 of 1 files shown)


normanmaurer:
This came up here https://github.com/netty/netty/pull/8231#discussion_r213451757.

We should do the same in netty as well.

openssl-dynamic/src/main/c/jnilib.c
```diff
@@ -291,27 +291,41 @@ static const jint method_table_size = sizeof(method_table) / sizeof(method_table
 // JNI Method Registration Table End
 
 static jint netty_internal_tcnative_Library_JNI_OnLoad(JNIEnv* env, const char* packagePrefix) {
+    int errorOnLoadCalled = 0;
```
carl-mastrangelo: for coolness points, you can make these labels and goto the correct one:

```
bufferOnLoadCalled: netty_internal_tcnative_Buffer_JNI_OnUnLoad(env);
errorOnLoadCalled: netty_internal_tcnative_Error_JNI_OnUnLoad(env);

```

ejona86: You "fall through" from one label to the next, so it will call multiple unload methods naturally.

normanmaurer: Ah got it... not sure this is clearer tho. I think I would prefer to keep it as it is

Issue #345, opened 2018-05-30
Prepare to be able to link against openssl 1.1.1


normanmaurer:
OpenSSL 1.1.1 will be the new LTS release, so we should be able to link against it:

https://www.openssl.org/blog/blog/2018/05/18/new-lts/

normanmaurer:
@carl-mastrangelo @ejona86 @Scottmitch @trustin FYI

normanmaurer:
https://github.com/netty/netty-tcnative/commit/35cd01cdffddb789ca56188a8c2a08dda33d54d8 is related to this

floragunncom:
OpenSSL 1.1.1 is now finally released

normanmaurer:
@floragunncom will look into it the next weeks

Issue #256, opened 2017-03-24
Update to openssl 1.1.1 with TLS 1.3


snazy:
OpenSSL 1.1.1 is about to add support for TLS 1.3 according to this [blog post](https://www.openssl.org/blog/blog/2016/10/24/f2f-roadmap/).
Would be great to have TLS 1.3 in Netty via OpenSSL!

nename0:
Related to #249 , as 1.1.1 will be have equal api as 1.1.0:

> The only thing missing was being able to commit to a delivery date, so that downstream consumers (such as Akamai) could dependably plan their own TLS 1.3 rollouts.  Because of our sponsorship, OpenSSL will deliver TLS 1.3 on April 5, in a version compatible with 1.1.0.  Akamai will begin incorporating this update into its systems soon thereafter.
**If you have  applications that use OpenSSL, start migrating to OpenSSL 1.1.0 now. Then, when the version supporting TLS 1.3 is available, you can just "drop it in."**  If you don't have applications to migrate, then follow along on the GitHub repository; same basic interop is already working.  Let the countdown to global TLS 1.3 enablement begin!

 [Source](https://blogs.akamai.com/2017/01/tls-13-ftw.html)

floragunncom:
Now it seems the final openssl 1.1.1 is imminent (after 2 pre-releases) and i think it would be great if we get TLS 1.3 support with netty any time soon.

normanmaurer:
Related to https://github.com/netty/netty-tcnative/issues/345

hc-codersatlas:
Is it as simple as just updating OpenSSL or changes need to be made to support TLS 1.3?

normanmaurer:
@hc-codersatlas sadly not ... But its on my to do list

floragunncom:
OpenSSL 1.1.1 is now finally released

normanmaurer:
Will look into it.

normanmaurer:
@snazy @hc-codersatlas @floragunncom @nename0 just for the record I have everything working here locally and just need to do some more cleanup etc but will provide some PRs within the next 1-2 weeks.

PR #389, opened 2018-09-17
Add support for TLSv1.3 when compiled against openssl 1.1.1

normanmaurer:
Motivation:

OpenSSL 1.1.1 was released which supports TLSv1.3 and it is the now the current LTS release. We should be able to compile against it and also support TLSv1.3.

Modifications:

- Add some new native methods to allow to set TLSv1.3 ciphersuites
- Depending on if TLSv1.3 is supported or not set some flags

Result:

Be able to compile against OpenSSL 1.1.1 and make use of TLSv1.3. Fixes https://github.com/netty/netty-tcnative/issues/345 and https://github.com/netty/netty-tcnative/issues/256

### DIFF
--- a/openssl-dynamic/src/main/c/sslcontext.c
+++ b/openssl-dynamic/src/main/c/sslcontext.c
@@ -422,24 +422,40 @@ TCN_IMPLEMENT_CALL(void, SSLContext, clearOptions)(TCN_STDARGS, jlong ctx,
 }
 
 TCN_IMPLEMENT_CALL(jboolean, SSLContext, setCipherSuite)(TCN_STDARGS, jlong ctx,
-                                                         jstring ciphers)
+                                                         jstring ciphers, jboolean tlsv13)
 {
     tcn_ssl_ctxt_t *c = J2P(ctx, tcn_ssl_ctxt_t *);
     jboolean rv = JNI_TRUE;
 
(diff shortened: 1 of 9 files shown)


openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -347,6 +347,13 @@ TCN_IMPLEMENT_CALL(jlong, SSLContext, make)(TCN_STDARGS, jint protocol, jint mod
                               ssl_context_cleanup,
                               apr_pool_cleanup_null);
 
+
+#ifndef OPENSSL_NO_TLS1_3
+    // We specify this as we currently not handle these extra tickets send after the handshake / shutdown
+    // in netty. We may want to remove this call once we do.
+    SSL_CTX_set_num_tickets(ctx, 0);
```
normanmaurer: I suspect this is not needed anymore after my recent changes. I will check this later today as the wifi on the plane is not really great for downloading openssl 1.1.1 ;)

normanmaurer:
Anyone has any comments here ? I would love to pull this in and cut a release and then do the netty stuff :)

SharpMan:
TLS 1.3 final will be released in Chrome in about 45 days. Isn't better to wait and move to boringSSL ?

normanmaurer:
@SharpMan we will for sure also support it in the BoringSSL one once it is released.

PR #391, opened 2018-09-18
Do not install man-pages during openssl-static build to speed up things.

normanmaurer:
Motivation:

We do not care about the openssl manpages so there is no need to install them during the build.

Modifications:

Use make install_sw which will skip manpage installation.

Result:

Faster build


normanmaurer:
@ejona86 its all in my head 🤣

PR #392, opened 2018-09-22
Return 0 for SSL_OP_NO_TLSv1_3 when TLSv1.3 is not supported.

normanmaurer:
Motivation:

We need to use 0 for SSL_OP_NO_TLSv1_3 when TLSv1.3 is not supported to not disable any protocols by mistake when its not supported at all.

Modifications:

Define SSL_OP_NO_TLSv1_3 as 0 when TLSv1.3 is not supported.

Result:

Correctly handle the case when TLSv1.3 is not supported and so be able to use it in this case.

### DIFF
--- a/openssl-dynamic/src/main/c/ssl_private.h
+++ b/openssl-dynamic/src/main/c/ssl_private.h
@@ -137,7 +137,7 @@ extern const char* TCN_UNKNOWN_AUTH_METHOD;
 #define OPENSSL_NO_TLS1_3
 #endif // OPENSSL_NO_TLS1_3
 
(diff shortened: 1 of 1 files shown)


normanmaurer:
I messed this one up ... We need this to be able to correctly handle both cases: when TLSv1.3 is supported and if its not supported.

normanmaurer:
Once this is merged I will cut a release.

normanmaurer:
@netty-bot test this please

normanmaurer:
@netty-bot test this please

PR #394, opened 2018-09-23
Use type-safe wrapper for STACK_OF(SSL_CIPHERS) lookup.

davidben:
Motivation:

The wrappers avoid casting and are more stable. This was mostly done in 94d4c41b98d37dcc26960098bd41c1624852b13a, but missed a spot.

Modifications:

Use sk_SSL_CIPHER_value instead of casting stack types manually.

Result:

Less casting.


normanmaurer:
@netty-bot test this please

Issue #120, opened 2016-02-23
Static Library Build Phase


Scottmitch:
Just initializing the project files for the IDE `mvn eclipse:eclipse` seems to build all of the static libraries. Is this necessary just for setting up the IDE project files (building the static libraries can take some time).

nmittler:
use intellij :) .... feel free to pick another appropriate phase.  So long as everything build we should be fine.

Scottmitch:
is there any reason why `generate-sources` must be used for everything? are there any other dependencies on other phases?

Would the following structure make more sense:
`generate-sources`: ftp code
`compile`: compile the code
`package`: build static libraries

nmittler:
We just need to ensure that APR and SSL are built before the tcnative code.

Scottmitch:
> We just need to ensure that APR and SSL are built before the tcnative code.

Can you explain why? Seems like as long as we have the header files for the libraries tcnative depends upon then tcnative should be able to build. Why can't we delay the build of the 3rd party libraries until we are trying to package/link stuff together?

nmittler:
Sure you can do that ... just not sure what it buys you.  To be honest, I don't think the problem is which phase we do it in, it's that we don't want to re-do what we've already done. If APR and SSL are already there, then skip download/build.

Scottmitch:
> To be honest, I don't think the problem is which phase we do it in, it's that we don't want to re-do what we've already done.

I think these are 2 separate problems :) The second part should be covered by https://github.com/netty/netty-tcnative/issues/121

Scottmitch:
> Sure you can do that ... just not sure what it buys you.

The current approach does not follow standard maven conventions. This means that plugins and tools may not function as expected or additional work may be necessary at unexpected times (e.g. why do I have to build openssl, boringssl, libressl, apr, etc... just to setup my IDE)?

nmittler:
Ok, if you have ideas for what needs to change, throw together a PR and we'll take a look

PR #395, opened 2018-09-24
Use correct maven lifecycle phases during static builds.

normanmaurer:
Motivation:

We should use the correct phases for the native build. Which means we should ownload download and unpack the needed resources during generate-sources but do the compilation during the compile phase.

Modifications:

Break up tasks to execute in the correct phase.

Result:

Fixes https://github.com/netty/netty-tcnative/issues/120.


normanmaurer:
Let me merge this as its only build script xml stuff

PR #396, opened 2018-09-25
Use generated-sources/ for copied source code during build of static …

normanmaurer:
…libraries.

Motivation:

We should not put copied content in an other place then target/ during build of static modules to not confuse things.

Modifications:

Copy src/main/* into target/generated-sources during build of static artifacts.

Result:

Less confusing and more clear build

### DIFF
--- a/boringssl-static/pom.xml
+++ b/boringssl-static/pom.xml
@@ -240,7 +240,8 @@
                 <phase>compile</phase>
                 <configuration>
                   <name>netty_tcnative</name>
-                  <nativeSourceDirectory>${project.basedir}/src/main/c</nativeSourceDirectory>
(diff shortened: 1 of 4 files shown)


normanmaurer:
@ejona86 for whatever f**king reason the hawtjni plugin will not produce the correct libtool command otherwise.

normanmaurer:
@ejona86 after some more debugging I found out howto get rid of the extra "-I" things :) I had to set customPackageDirectory.

normanmaurer:
@ejona86 the source is the documentation ;) https://github.com/fusesource/hawtjni/blob/master/hawtjni-maven-plugin/src/main/java/org/fusesource/hawtjni/maven/GenerateMojo.java

PR #397, opened 2018-09-26
Adjust pom.xml to be able to compile on Java8,9,10,11,12

normanmaurer:
Motivation:

We should be able to compile on recent versions of Java.

Modifications:

Add profiles and configuration to be able to build on Java8,9,10,11,12

Result:

Be able to build on recent versions of java

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -19,7 +19,7 @@
   <parent>
     <groupId>io.netty</groupId>
     <artifactId>netty-parent</artifactId>
-    <version>4.0.18.Final</version>
+    <version>4.0.29.Final</version>
   </parent>
   <artifactId>netty-tcnative-parent</artifactId>
   <version>2.0.18.Final-SNAPSHOT</version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -40,6 +40,11 @@
 
   <properties>
     <checkstyle.skip>true</checkstyle.skip>
+    <enforcer.plugin.version>1.4.1</enforcer.plugin.version>
```
ejona86: Could you put a comment here (or below) why the version is selected vs the newer version used later (I assume it is the last version that works on Java 1.6)?

pom.xml
```diff
@@ -147,12 +152,35 @@
...
-          <source>1.6</source>
-          <target>1.6</target>
+          <source>${maven.compiler.source}</source>
+          <target>${maven.compiler.target}</target>
+        </configuration>
+      </plugin>
+
+      <plugin>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration combine.self="override">>
```
ejona86: I don't think you meant `>>`. I'm surprised it runs.

pom.xml
normanmaurer: Basically because it’s a non final release and I prefer to use a final by default

pom.xml
```diff
@@ -147,12 +152,35 @@
...
+            <exclude>**/*TestUtil*</exclude>
+          </excludes>
+          <runOrder>random</runOrder>
+          <properties>
+            <property>
+              <name>listener</name>
+              <value>io.netty.build.junit.TimedOutTestsListener</value>
+            </property>
+          </properties>
+          <jvm>${testJavaHome}/bin/java</jvm>
```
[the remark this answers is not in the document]
normanmaurer: Actually we do not need this... let’s remove

pom.xml
ejona86: Okay. That's fine.

normanmaurer:
@ejona86 addressed.

PR #398, opened 2018-10-12
Do not return early in certificate_cb if keytypes could not be detected.

normanmaurer:
Motivation:

We must not return early in certificate_cb if keytypes could not be detected as we correctly handle this case in the Java code. This is especially important to correctly support TLSv1.3

Modifications:

Do not return early when the keytypes could not be detected and just let the java callback handle it.

Result:

Be able to use TLSv1.3 correctly.


normanmaurer:
Will cut a release now that this one is merged

PR #399, opened 2018-10-12
Ensure generated-sources/java is included in jar.

normanmaurer:
Motivation:

ef30a10dcb9b21eb14fa1dc0bfcc249c235c809e changed the build to use generated-sources but missed to ensure we also compile the java classes in the generated jars.

Modifications:

Correctly include into jar.

Result:

Jars contain native code and java code.

### DIFF
--- a/boringssl-static/pom.xml
+++ b/boringssl-static/pom.xml
@@ -89,6 +89,24 @@
             </executions>
           </plugin>
 
+          <plugin>
+            <groupId>org.codehaus.mojo</groupId>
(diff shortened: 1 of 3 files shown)


normanmaurer:
Followup of 
https://github.com/netty/netty-tcnative/pull/396

carl-mastrangelo:
Remind me, what generated sources are present?   (We build netty tcnative from a clean checkout from git,  and using a custom bazel BUILD file.  Will it still work after this?)

PR #407, opened 2018-10-22
Cleanup BoringSSL TLSv1.3 support and consistent handle empty ciphers.

normanmaurer:
Motivation:

Some code can be removed and also we should ensure we handle empty ciphers consistent between BoringSSL and OpenSSL.

Modifications:

- Remove not needed called for min protocol version when using BoringSSL
- Return when an empty ciphers string is given as this will fail in BoringSSL while just return on OpenSSL.

Result:

Cleaner and more consistent code.


carl-mastrangelo:
stale branch?

trustin:
Thanks, @carl-mastrangelo :-)

PR #412, opened 2018-11-08
Correctly implement SSL.getSigAlgs(...) for BoringSSL

normanmaurer:
Motivation:

While BoringSSL does not support SSL_get_sig_algs(...) we can make use of SSL_get0_peer_verify_algorithms for our use case.

Modifications:

Use SSL_get0_peer_verify_algorithms when BoringSSL is used and so have a correct implemention there as well (which is needed for ExtendedSSLSession in java).

Result:

Be able to correctly implement ExtendedSSLSession.getPeerSupportedSignatureAlgorithms() when using BoringSSL.

### DIFF
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -2235,10 +2235,35 @@ TCN_IMPLEMENT_CALL(jobjectArray, SSL, getSigAlgs)(TCN_STDARGS, jlong ssl) {
     SSL *ssl_ = J2P(ssl, SSL *);
     TCN_CHECK_NULL(ssl_, ssl, NULL);
 
-// Not supported by BoringSSL and LibreSSL
-// https://boringssl.googlesource.com/boringssl/+/ba16a1e405c617f4179bd780ad15522fb25b0a65%5E%21/
-#if defined(OPENSSL_IS_BORINGSSL) || defined(LIBRESSL_VERSION_NUMBER)
+// Not supported in LibreSSL
+#if defined(LIBRESSL_VERSION_NUMBER)
     return NULL;
+#elif defined(OPENSSL_IS_BORINGSSL)
+    // Using a different API in BoringSSL
+    // https://boringssl.googlesource.com/boringssl/+/ba16a1e405c617f4179bd780ad15522fb25b0a65%5E%21/
+    int i;
+    jobjectArray array;
+    jstring algString;
+    const uint16_t *peer_sigalgs;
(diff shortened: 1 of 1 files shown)


normanmaurer:
@davidben FYI...

normanmaurer:
may also be interesting for Conscrypt...

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -2235,10 +2235,35 @@ TCN_IMPLEMENT_CALL(jobjectArray, SSL, getSigAlgs)(TCN_STDARGS, jlong ssl) {
...
+// Not supported in LibreSSL
+#if defined(LIBRESSL_VERSION_NUMBER)
     return NULL;
+#elif defined(OPENSSL_IS_BORINGSSL)
+    // Using a different API in BoringSSL
+    // https://boringssl.googlesource.com/boringssl/+/ba16a1e405c617f4179bd780ad15522fb25b0a65%5E%21/
+    int i;
+    jobjectArray array;
+    jstring algString;
+    const uint16_t *peer_sigalgs;
```
ejona86: Does this need to be freed?

normanmaurer: @ejona86 nope... _get0_ does not transfer ownership.

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -2235,10 +2235,35 @@ TCN_IMPLEMENT_CALL(jobjectArray, SSL, getSigAlgs)(TCN_STDARGS, jlong ssl) {
...
+    array = (*e)->NewObjectArray(e, num_peer_sigalgs, tcn_get_string_class(), NULL);
+
+    if (array == NULL) {
+        return NULL;
+    }
+
+    for (i = 0; i < num_peer_sigalgs; i++) {
+        algString = (*e)->NewStringUTF(e, SSL_get_signature_algorithm_name(peer_sigalgs[i], SSL_version(ssl_) != TLS1_2_VERSION));
+        if (algString == NULL) {
+            // something is wrong we should better just return here
```
carl-mastrangelo: Doesn't `array` need to be freed here?

normanmaurer: Good point... this problem exists in other places as well (as this code was mostly a copy + modify of existing code). I will do a pr shortly.

normanmaurer: Actually no it does not need to be freed as it will be handled by the GC:

https://stackoverflow.com/a/12207973/1074097

Issue #419, opened 2018-12-20
parsePackagePrefix does not correctly unescape JNI package paths


kennknowles:
In JNI, the sequence `_1` stands for an underscore in a package name or class name. However, the logic in [parsePackagePrefix](https://github.com/netty/netty-tcnative/blob/master/openssl-dynamic/src/main/c/jnilib.c#L266) unconditionally converts all underscores to slashes.

This arose in the context of Apache Beam, which relocated it to `org.apache.beam.vendor.grpc.v1_13_1.netty`. Our workaround is to just not use underscores, but this should also be easy to fix.

More details at https://issues.apache.org/jira/browse/BEAM-6249

normanmaurer:
@kennknowles PRs welcome :)

kennknowles:
@normanmaurer of course! I was just working on it. I don't do a lot of C these days so my emacs config had rotted :-). #420.

normanmaurer:
@kennknowles 🤣 vi FTW ;)

kennknowles:
I tried `vim` first, but my config there is even worse and kept wanting to use tabs :-p (just to bring both of the great debates into this...)

trustin:
Should be fixed in 2.0.52.Final.

PR #420, opened 2018-12-20
Correctly process JNI escape sequence _1 (#419)

kennknowles:
Current logic incorrectly interprets e.g. `a_b_1c` as `a/b/1c` instead of the correct `a/b_c`. After this patch, the `_1` is processed as an underscore.

### DIFF
--- a/openssl-dynamic/src/main/c/jnilib.c
+++ b/openssl-dynamic/src/main/c/jnilib.c
@@ -255,26 +255,48 @@ static char* parsePackagePrefix(const char* libraryPathName, jint* status) {
     if (packagePrefix == packageNameEnd) {
         return NULL;
     }
-    // packagePrefix length is > 0
-    // Make a copy so we can modify the value without impacting libraryPathName.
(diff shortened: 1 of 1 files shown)


kennknowles:
@normanmaurer this is more of a demonstration of what the logic might be... I'm quite sure I am not utilizing the libs I ought to, and I'm not sure where a test for this would fit in. If you point me to the place to add a test, I'll give it a try.

normanmaurer:
@kennknowles we have some tests related to shading netty. So maybe you can add these there ?

https://github.com/netty/netty/tree/4.1/testsuite-shading

normanmaurer:
@kennknowles ping

kennknowles:
Still alive, just got busy!

hyperxpro:
Just checking if you got any update @kennknowles ?

trustin:
Sorry that it took so long. It should be fixed via the following PRs:

- https://github.com/netty/netty-jni-util/pull/13
- https://github.com/netty/netty/pull/12358
- https://github.com/netty/netty-tcnative/pull/726

PR #421, opened 2018-12-21
Introduce a new native method to access the master key

fzakaria:
This change accompanies the broader pull-request introduced by 
https://github.com/netty/netty/pull/8653 to allow access to the SSL session master key for debugging purposes.

Accessing the master key is useful if you ever need to decrypt the TLS session for debugging purposes with tools such as Wireshark.

Access of the field from the SSL session struct is direct as the wrapper method was only introduced in 1.1.0+  https://www.openssl.org/docs/man1.1.0/ssl/SSL_SESSION_get_master_key.html

### DIFF
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -1607,6 +1607,137 @@ TCN_IMPLEMENT_CALL(jboolean, SSL, setCipherSuites)(TCN_STDARGS, jlong ssl,
     return rv;
 }
 
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
+/*
+ * Backport of SSL_SESSION_get_master_key from 1.1
+ */
+size_t SSL_SESSION_get_master_key(const SSL_SESSION *session,
+                                  unsigned char *out, size_t outlen)
+{
+    if (outlen == 0) {
+        return session->master_key_length;
+    }
+    if (outlen > session->master_key_length) {
+        outlen = session->master_key_length;
+    }
+    memcpy(out, session->master_key, outlen);
+    return outlen;
+}
+
+/*
+ * Backport of SSL_get_server_random from 1.1
+ */
+size_t SSL_get_server_random(const SSL *ssl, unsigned char *out, size_t outlen)
+{
+    if (outlen == 0) {
+        return sizeof(ssl->s3->server_random);
+    }
+    if (outlen > sizeof(ssl->s3->server_random)) {
+        outlen = sizeof(ssl->s3->server_random);
+    }
+    memcpy(out, ssl->s3->server_random, outlen);
+    return outlen;
+}
+
+/*
+ * Backport of SSL_get_client_random from 1.1
+ */
+size_t SSL_get_client_random(const SSL *ssl, unsigned char *out, size_t outlen)
+{
+    if (outlen == 0) {
+        return sizeof(ssl->s3->client_random);
+    }
+    if (outlen > sizeof(ssl->s3->client_random)) {
+        outlen = sizeof(ssl->s3->client_random);
+    }
+    memcpy(out, ssl->s3->client_random, outlen);
+    return outlen;
+}
+#endif
+
+TCN_IMPLEMENT_CALL(jbyteArray, SSL, getClientRandom)(TCN_STDARGS, jlong ssl)
+{
+
+    SSL *ssl_ = J2P(ssl, SSL *);
+    TCN_CHECK_NULL(ssl_, ssl, NULL);
+
+    size_t keyLength = SSL_get_client_random(ssl_, NULL, 0);
+    TCN_ASSERT(keyLength <= 0x7FFFFFFF); /* must fit into 32 bit unsigned jsize */
+
+    unsigned char *key = OPENSSL_malloc(sizeof(unsigned char) * keyLength);
+    if (key == NULL) {
+        tcn_ThrowException(e, "OPENSSL_malloc() returned null");
+        return NULL;
+    }
+
+    size_t bytesMoved = SSL_get_client_random(ssl_, key, keyLength);
+    TCN_ASSERT(bytesMoved == keyLength);
+
+    jbyteArray jKey = (*e)->NewByteArray(e, (jsize) bytesMoved);
+    (*e)->SetByteArrayRegion(e, jKey, 0, bytesMoved, (jbyte*) key);
+
+    OPENSSL_free(key);
+
+    return jKey;
+}
+
+TCN_IMPLEMENT_CALL(jbyteArray, SSL, getServerRandom)(TCN_STDARGS, jlong ssl)
+{
+
+    SSL *ssl_ = J2P(ssl, SSL *);
+    TCN_CHECK_NULL(ssl_, ssl, NULL);
+
+    size_t keyLength = SSL_get_server_random(ssl_, NULL, 0);
+    TCN_ASSERT(keyLength <= 0x7FFFFFFF); /* must fit into 32 bit unsigned jsize */
+
+    unsigned char *key = OPENSSL_malloc(sizeof(unsigned char) * keyLength);
+    if (key == NULL) {
+        tcn_ThrowException(e, "OPENSSL_malloc() returned null");
+        return NULL;
+    }
+
+    size_t bytesMoved = SSL_get_server_random(ssl_, key, keyLength);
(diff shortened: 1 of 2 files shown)


normanmaurer:
@netty-bot test this please

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1607,6 +1607,54 @@ TCN_IMPLEMENT_CALL(jboolean, SSL, setCipherSuites)(TCN_STDARGS, jlong ssl,
     return rv;
 }
 
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
+/*
+  Backport of SSL_SESSION_get_master_key from 1.1
+*/
```
normanmaurer: nit: use either `//` or:

```
/*
 * 
 */
```

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1607,6 +1607,54 @@ TCN_IMPLEMENT_CALL(jboolean, SSL, setCipherSuites)(TCN_STDARGS, jlong ssl,
...
+size_t SSL_SESSION_get_master_key(const SSL_SESSION *session,
+                                  unsigned char *out, size_t outlen)
+{
+    if (outlen == 0)
```
normanmaurer: please use `{...}`

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1607,6 +1607,54 @@ TCN_IMPLEMENT_CALL(jboolean, SSL, setCipherSuites)(TCN_STDARGS, jlong ssl,
...
+        return session->master_key_length;
+    if (outlen > session->master_key_length)
```
normanmaurer: please use `{...}`

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1607,6 +1607,54 @@ TCN_IMPLEMENT_CALL(jboolean, SSL, setCipherSuites)(TCN_STDARGS, jlong ssl,
...
+        outlen = session->master_key_length;
+    memcpy(out, session->master_key, outlen);
+    return outlen;
+}
+#endif
+
```
normanmaurer: nit: remove one empty line

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1607,6 +1607,54 @@ TCN_IMPLEMENT_CALL(jboolean, SSL, setCipherSuites)(TCN_STDARGS, jlong ssl,
...
+TCN_IMPLEMENT_CALL(jbyteArray, SSL, getMasterKey)(TCN_STDARGS, jlong ssl)
+{
+
+    SSL *ssl_ = J2P(ssl, SSL *);
+    TCN_CHECK_NULL(ssl_, ssl, NULL);
+
+    SSL_SESSION *session = NULL;
+    if (session == NULL) {
+        return NULL;
+    }
```
normanmaurer: This does not look correct.. session will also be `NULL` here. Did you forget to call `SSL_get0_SESSION(...)` ?

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1607,6 +1607,54 @@ TCN_IMPLEMENT_CALL(jboolean, SSL, setCipherSuites)(TCN_STDARGS, jlong ssl,
...
+
+    jbyteArray jKey = (*e)->NewByteArray(e, (jsize) keyLength);
+    (*e)->SetByteArrayRegion(e, jKey, 0, keyLength, (jbyte*) key);
+
+    OPENSSL_free(key);
+
+    return jKey;
+}
+
+
```
normanmaurer: nit: remove one empty line

fzakaria:
@normanmaurer Thanks for the comments. I'll address the feedback. I'd like to include a unit test, do you have a good example of how to start a TLS session with the code in this JAR? 

I'd only know how to write a test by starting client+server but that would need a cyclic dependency on SslHandler in Netty.

normanmaurer:
@fzakaria unfortually no :( I guess the best bet is to include a test in Netty itself.

fzakaria:
Thanks. I'll work on that for a post PR commit I guess. 
A good test would be if in Netty itself I could also keep a copy of the encrypted messages before the engine grabs it and then manually decrypt it as well.

fzakaria:
@normanmaurer Addressed your feedback -- 
I really wanted to include a test for posterity but I've been stuck on getting it to work.

I've packaged the work on the test case so far:
https://gist.github.com/fzakaria/f1de28943c8932280254dd90ffbc4ebd

fzakaria:
@normanmaurer I've included a test in **netty** that proves the methods work by manually decrypting the TLS application data -- https://github.com/netty/netty/pull/8699

fzakaria:
@normanmaurer *ping*

normanmaurer:
@fzakaria sorry for the delay... Will try to review tomorrow latest.

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1607,6 +1607,137 @@ TCN_IMPLEMENT_CALL(jboolean, SSL, setCipherSuites)(TCN_STDARGS, jlong ssl,
...
+    if (key == NULL) {
+        tcn_ThrowException(e, "OPENSSL_malloc() returned null");
+        return NULL;
+    }
+
+    size_t bytesMoved = SSL_get_client_random(ssl_, key, keyLength);
+    TCN_ASSERT(bytesMoved == keyLength);
+
+    jbyteArray jKey = (*e)->NewByteArray(e, (jsize) bytesMoved);
+    (*e)->SetByteArrayRegion(e, jKey, 0, bytesMoved, (jbyte*) key);
```
normanmaurer: I think we need to guard against when `NewByteArray(...)` returns `NULL` (like if there is no memory left).

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1607,6 +1607,137 @@ TCN_IMPLEMENT_CALL(jboolean, SSL, setCipherSuites)(TCN_STDARGS, jlong ssl,
...
+    unsigned char *key = OPENSSL_malloc(sizeof(unsigned char) * keyLength);
+    if (key == NULL) {
+        tcn_ThrowException(e, "OPENSSL_malloc() returned null");
+        return NULL;
+    }
+
+    size_t bytesMoved = SSL_get_server_random(ssl_, key, keyLength);
+    TCN_ASSERT(bytesMoved == keyLength);
+
+    jbyteArray jKey = (*e)->NewByteArray(e, (jsize) bytesMoved);
```
normanmaurer: same as above.

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1607,6 +1607,137 @@ TCN_IMPLEMENT_CALL(jboolean, SSL, setCipherSuites)(TCN_STDARGS, jlong ssl,
...
+    unsigned char *key = OPENSSL_malloc(sizeof(unsigned char) * keyLength);
+    if (key == NULL) {
+        tcn_ThrowException(e, "OPENSSL_malloc() returned null");
+        return NULL;
+    }
+
+    size_t bytesMoved = SSL_SESSION_get_master_key(session, key, keyLength);
+    TCN_ASSERT(bytesMoved == keyLength);
+
+    jbyteArray jKey = (*e)->NewByteArray(e, (jsize) bytesMoved);
```
normanmaurer: same as above

openssl-dynamic/src/main/c/ssl.c
fzakaria: I can add the NULL checks however I stuck to previous convention where the other JNI calls did no such guard.

fzakaria:
@normanmaurer @Lukasa  Please see https://github.com/netty/netty/pull/8699/files#diff-7a4cbf2f526e2066184fb9ef098cd815R1127 also on a test-case of it being used to then decrypt the TLS data by hand.

fzakaria:
Done.

normanmaurer:
@fzakaria can you do me a favour and squash + adjust the commit message to match our template:

https://netty.io/wiki/writing-a-commit-message.html

normanmaurer:
@fzakaria thanks

fzakaria:
Thanks for merging @normanmaurer 
As soon as new version is released -- I will bump the _pom.xml_ in the netty pull-request and it should be _good-to-go_ as well.

PR #422, opened 2018-12-26
Adding Automatic-Module-Name to all manifests.

TheIndifferent:
Motivation:

When the library is used as a JPMS dependency or transitive dependency
("tcnative-boringssl-static" for example), JVM attempts to trasform it
into a module and fails because "static" is a reserved word.

Modifications:

The change adds "Automatic-Module-Name" to manifests of all jar files.

Result:

JVM will use the "Automatic-Module-Name" instead of deriving the
module name from the jar file name.
The declared module name can also be used as a dependency name
in JPMS module declaration.

### DIFF
--- a/boringssl-static/pom.xml
+++ b/boringssl-static/pom.xml
@@ -30,6 +30,23 @@
     to BoringSSL and Apache APR.
   </description>
 
+  <build>
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <artifactId>maven-jar-plugin</artifactId>
+          <configuration>
+            <archive>
+              <manifestEntries>
+                <Automatic-Module-Name>io.netty.tcnative.boringssl</Automatic-Module-Name>
+              </manifestEntries>
+            </archive>
+          </configuration>
+        </plugin>
+      </plugins>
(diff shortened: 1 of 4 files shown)


normanmaurer:
@TheIndifferent can you please sign our ICLA and adjust the commit message to match our template:

https://netty.io/s/icla
https://netty.io/wiki/writing-a-commit-message.html

normanmaurer:
@netty-bot test this please

normanmaurer:
@TheIndifferent also would be nice if you could add something to the README.md like here:

https://github.com/netty/netty#usage-with-jdk-9

TheIndifferent:
@normanmaurer CLA signed, commit message changed, could you please take a look if that makes sense?

Looks like current readme just redirects to the wiki and is empty otherwise. Would it make more sense to write that on the wiki instead?

normanmaurer:
@TheIndifferent yes please add it to the wiki. I will pull this one in in the meantime.

normanmaurer:
@netty-bot test this please

normanmaurer:
@TheIndifferent merged thanks.

PR #428, opened 2019-02-08
Use correct return value in certificate_cb for errors.

normanmaurer:
Motivation:

As explained in the SSL_set_cert_cb manpage 0 should be used to signal back errors. At the moment we use -1 which should only be used to signal back that the operation should be retried later.

See https://www.openssl.org/docs/man1.0.2/man3/SSL_set_cert_cb.html.

At the moment -1 also works as we not correctly handle SSL_ERROR_WANT_X509_LOOKUP in nettys ReferenceCountedOpenSslEngine and just treat SSL_ERROR_WANT_X509_LOOKUP as error (which is not correct).

Modifications:

Use 0 to signal a handshake error.

Result:

Use correct return value for errors.

### DIFF
--- a/openssl-dynamic/src/main/c/sslcontext.c
+++ b/openssl-dynamic/src/main/c/sslcontext.c
@@ -1609,15 +1609,16 @@ TCN_IMPLEMENT_CALL(void, SSLContext, setCertRequestedCallback)(TCN_STDARGS, jlon
 }
 
 
+// See https://www.openssl.org/docs/man1.0.2/man3/SSL_set_cert_cb.html for return values.
 static int certificate_cb(SSL* ssl, void* arg) {
 #if defined(LIBRESSL_VERSION_NUMBER)
     // Not supported with LibreSSL
-    return -1;
+    return 0;
 #else
 #ifndef OPENSSL_IS_BORINGSSL
     if (OpenSSL_version_num() < 0x10002000L) {
         // Only supported on openssl 1.0.2+
-        return -1;
+        return 0;
     }
 #endif // OPENSSL_IS_BORINGSSL
 
@@ -1645,7 +1646,7 @@ static int certificate_cb(SSL* ssl, void* arg) {
 
     // Check if java threw an exception and if so signal back that we should not continue with the handshake.
     if ((*e)->ExceptionCheck(e)) {
-        return -1;
+        return 0;
     }
 
     // Everything good...


normanmaurer:
The build error is not related to this PR:

```
08:41:56      [exec] In file included from ../crypto/fipsmodule/ec/p256-x86_64_test.cc:22:0:
08:41:56      [exec] ../crypto/fipsmodule/ec/p256-x86_64_test.cc: In function ‘std::string FieldElementToString(const uint64_t*)’:
08:41:56      [exec] ../crypto/../include/openssl/bn.h:156:28: error: expected ‘)’ before ‘PRIx64’
08:41:56      [exec]  #define BN_HEX_FMT2 "%016" PRIx64
08:41:56      [exec]                             ^
08:41:56      [exec] ../crypto/fipsmodule/ec/p256-x86_64_test.cc:175:36: note: in expansion of macro ‘BN_HEX_FMT2’
08:41:56      [exec]      BIO_snprintf(buf, sizeof(buf), BN_HEX_FMT2, a[i]);
08:41:56      [exec]                                     ^
08:41:56      [exec] ../crypto/fipsmodule/ec/p256-x86_64_test.cc:175:53: error: conversion lacks type at end of format [-Werror=format=]
08:41:56      [exec]      BIO_snprintf(buf, sizeof(buf), BN_HEX_FMT2, a[i]);
08:41:56      [exec]                                                      ^
08:41:56      [exec] ../crypto/fipsmodule/ec/p256-x86_64_test.cc:175:53: error: too many arguments for format [-Werror=format-extra-args]
08:41:56      [exec] cc1plus: all warnings being treated as errors
```

davidben:
You probably are building on a system without C++11-compliant headers.
https://boringssl.googlesource.com/boringssl/+/ce455886958b61f4b350e46a6a1a764f325853e1

You can define `__STDC_FORMAT_MACROS` and `__STDC_CONSTANT_MACROS` to work around this, but we have required C++11 for a while now, so you may want to file a bug with the distribution.

normanmaurer:
@davidben thanks will check... this is on centos 6 and the build just recently started to fail so no idea why :/

davidben:
Haven't checked, but I'm probably it only just trickled into chromium-stable. (Huh. Would have thought it'd have done so sooner. I guess not.)

normanmaurer:
@davidben thanks for checking... no worries. I will fix this up here.

normanmaurer:
@davidben https://github.com/netty/netty-tcnative/pull/429 FYI

PR #430, opened 2019-02-11
Support to offload certificate selection to different thread.

normanmaurer:
Motiviation:

Sometimes selection of the certificate can take a long time as it may be stored on a remote system (for example). We should support to offload selection of the certificate to an other thread and be able to retry the SSL operation after it is done.

This will also allow us to support the SSLEngine abstraction of SSLEngine.getDelegatingTask() and SSLEngineResult.HandshakeStatus.NEED_TASK.

Modification:

- Introduce SSLContest.setUseTasks(...) which allows to enable / disable the usage of tasks to offload certificate selection to an other thread.
- Introduce SSL.getTask(...) which allows to retrieve the task that will do the certificate selection ( and also other tasks in the future )

Result:

Be able to execute long running tasks for certificate selection ( and also others in the future) without blocking the I/O thread.

### DIFF
--- a/openssl-dynamic/src/main/c/sslcontext.c
+++ b/openssl-dynamic/src/main/c/sslcontext.c
@@ -1681,6 +1681,12 @@ static int certificate_cb(SSL* ssl, void* arg) {
         jobject task = (*e)->NewObject(e, certificateCallbackTask_class, certificateCallbackTask_init, P2J(ssl), types, issuers, c->certificate_callback);
         sslTask = (tcn_ssl_task_t*) OPENSSL_malloc(sizeof(tcn_ssl_task_t));
         sslTask->task = (*e)->NewGlobalRef(e, task);
+        if (sslTask->task == NULL) {
+            // NewGlobalRef failed because we ran out of memory, free what we malloc'ed and fail the handshake.
+            OPENSSL_free(sslTask);
+
+            return 0;
+        }
         sslTask->consumed = JNI_FALSE;
 
         tcn_SSL_set_app_data5(ssl, sslTask);
--- a/openssl-dynamic/src/main/c/sslcontext.c
+++ b/openssl-dynamic/src/main/c/sslcontext.c
@@ -38,6 +38,14 @@
 #include <stdint.h>
 #include "sslcontext.h"
 
+
+static jclass    sslTask_class;
+static jfieldID  sslTask_returnValue;
+static jfieldID  sslTask_complete;
+
+static jclass    certificateCallbackTask_class;
+static jmethodID certificateCallbackTask_init;
+
 extern apr_pool_t *tcn_global_pool;
 
 static apr_status_t ssl_context_cleanup(void *data)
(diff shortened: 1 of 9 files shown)


openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -1640,17 +1674,33 @@ static int certificate_cb(SSL* ssl, void* arg) {
...
     }
 
-    // Execute the java callback
-    (*e)->CallVoidMethod(e, c->certificate_callback, c->certificate_callback_method,
-             P2J(ssl), types, issuers);
+    // Let's check if we should provide the certificate callback as task that can be run on another Thread.
+    if (c->use_tasks != 0) {
+        // Lets create the CertificateCallbackTask and store it on the SSL object. We then later retrieve it via
+        // SSL.getTask(ssl) and run it.
+        jobject task = (*e)->NewObject(e, certificateCallbackTask_class, certificateCallbackTask_init, P2J(ssl), types, issuers, c->certificate_callback);
```
carl-mastrangelo: I think you need to check for an exception after this line.

openssl-dynamic/src/main/java/io/netty/internal/tcnative/CertificateCallbackTask.java
```diff
@@ -0,0 +1,45 @@
...
+        this.callback = callback;
+    }
+
+    // See https://www.openssl.org/docs/man1.0.2/man3/SSL_set_cert_cb.html.
+    @Override
+    protected int runTask(long ssl) {
+        try {
+            callback.handle(ssl, keyTypeBytes, asn1DerEncodedPrincipals);
+            return 1;
+        } catch (Exception e) {
```
carl-mastrangelo: It would be nice if this was logged at least.

normanmaurer: In netty we do before rethrow... we dont have any dependencies on a logging framework here. We could use java.util.logger ....

openssl-dynamic/src/main/c/sslcontext.c
normanmaurer: the constructor does not throw so I guess the only exception that could happen is a `OOME`.

openssl-dynamic/src/main/java/io/netty/internal/tcnative/CertificateCallbackTask.java
carl-mastrangelo: That's what I was implying.  Else, you could just rethrow it as RuntimeException.

normanmaurer: @carl-mastrangelo the problem is that rethrowing here is really "useless" as we will process it as part of an openssl c callback which needs to return 0 for an error to abort the handshake.

normanmaurer: @carl-mastrangelo also I think we don't want to log as we rethrow it on the Netty level correctly after the `SSL.doHandshake(...)` method returns and so fail the handshake with an `SSLHandshakeException`.

normanmaurer: so to make it short I think it is all fine as it is... Maybe I should add some comments tho ?

openssl-dynamic/src/main/c/sslcontext.c
carl-mastrangelo: Can you put that as a comment in CertificateCallbackTask's ctor, so it's clear it can't throw in the future too?

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -1640,17 +1674,33 @@ static int certificate_cb(SSL* ssl, void* arg) {
...
+        sslTask = (tcn_ssl_task_t*) OPENSSL_malloc(sizeof(tcn_ssl_task_t));
+        sslTask->task = (*e)->NewGlobalRef(e, task);
```
carl-mastrangelo: sorry to be nitpicky, but this can fail too.  Normally in Java I wouldn't care so much, but in C the risk is greater and it's much harder to debug.

openssl-dynamic/src/main/java/io/netty/internal/tcnative/CertificateCallbackTask.java
carl-mastrangelo: Comment is fine.

normanmaurer:
@carl-mastrangelo addressed your comments

openssl-dynamic/src/main/c/sslcontext.c
normanmaurer: done

PR #432, opened 2019-02-21
Use correct return code when exception is thrown in SSL_CTX_set_cert_…

normanmaurer:
…verify_callback callback.

Motivation:

We used -1 as return code but it should be 0 as stated in the manpage of SSL_CTX_set_cert_verify_callback.

See https://www.openssl.org/docs/man1.0.2/man3/SSL_CTX_set_cert_verify_callback.html

Also we should set the error in this case which we did not.

Modifications:

Return correct return code and also set error.

Result:

Use correct return code on exception.


normanmaurer:
Also let me pull this in as its a simple fix.

PR #433, opened 2019-02-21
Correctly init all fields of struct after malloc it.

normanmaurer:
Motivation:

In 11120b2d81fdc7de4619c904fc73622f77824ac8 we introduced the notion of offload operations as tasks but failed to ensure we init all fields of tcn_ssl_task_t. This can lead to strange side-effects if the OPENSSL_malloc(...) implementation does not memset.

Modifications:

Correctly init consumed to JNI_FALSE.

Result:

Correctly init struct.


normanmaurer:
Let me pull this in as its an easy one and may produce some strange side effects.

PR #435, opened 2019-02-22
Allow to offload certificate validation when using BoringSSL

normanmaurer:
Motivation:

BoringSSL supports offloading certificate validation to a different thread. This is useful as it may need to do blocking operations and so may block the EventLoop.

Modification:

Add BoringSSL specific code which allows to offload certificate validation (just as we already have code for certificate selection).

Result:

Be able to offload certificate validation when using BoringSSL.

### DIFF
--- a/openssl-dynamic/src/main/c/ssl_private.h
+++ b/openssl-dynamic/src/main/c/ssl_private.h
@@ -199,6 +199,11 @@ extern void *SSL_temp_keys[SSL_TMP_KEY_MAX];
 #define SSL_CTRL_GET_CLIENT_CERT_TYPES          103
 #endif
 
+#ifndef SSL_ERROR_WANT_CERTIFICATE_VERIFY
+// See https://github.com/google/boringssl/blob/chromium-stable/include/openssl/ssl.h#L538
+#define SSL_ERROR_WANT_CERTIFICATE_VERIFY       -1
+#endif
+
 typedef struct tcn_ssl_ctxt_t tcn_ssl_ctxt_t;
 
 typedef struct {
--- a/openssl-dynamic/src/main/java/io/netty/internal/tcnative/CertificateVerifierTask.java
+++ b/openssl-dynamic/src/main/java/io/netty/internal/tcnative/CertificateVerifierTask.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2019 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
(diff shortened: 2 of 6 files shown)


normanmaurer:
@ejona86 you mentioned that this may be something you want for GRPC as well as you need to do some blocking operation in the `TrustManager`.

openssl-dynamic/src/main/c/ssl_private.h
```diff
@@ -199,6 +199,11 @@ extern void *SSL_temp_keys[SSL_TMP_KEY_MAX];
 #define SSL_CTRL_GET_CLIENT_CERT_TYPES          103
 #endif
 
+#ifndef SSL_ERROR_WANT_CERTIFICATE_VERIFY
+// See https://github.com/google/boringssl/blob/chromium-stable/include/openssl/ssl.h#L538
+#define SSL_ERROR_WANT_CERTIFICATE_VERIFY 16
```
davidben: I'm guessing this is defined just to ease compilation with OpenSSL or something. You might want to use -1 or something that'll never happen. Otherwise you might run into issues later on if OpenSSL allocates 16 for some other status.

normanmaurer: Good point

normanmaurer: fixed

normanmaurer:
@davidben thanks a lot !

openssl-dynamic/src/main/java/io/netty/internal/tcnative/CertificateVerifierTask.java
```diff
@@ -0,0 +1,38 @@
...
+    CertificateVerifierTask(long ssl, byte[][] x509, String authAlgorithm, CertificateVerifier verifier) {
+        super(ssl);
+        this.x509 = x509;
+        this.authAlgorithm = authAlgorithm;
+        this.verifier = verifier;
+    }
+
+    @Override
+    protected int runTask(long ssl) {
+        return verifier.verify(ssl, x509, authAlgorithm);
```
johnou: @normanmaurer shouldn't we catch Exception here and return 0 like we do in CertificateCallbackTask?

normanmaurer: verifier is not allowed to throw so I I think we do not need this.

normanmaurer:
@carl-mastrangelo @ejona86 any comments ? Would love to pull this in before cutting a release

PR #437, opened 2019-02-26
Remove version checks for runtime version of OpenSSL as these do not …

normanmaurer:
…work on RHEL

Motivation:

We tried to detect if KeyManagerFactory can be supported by checking the runtime version of OpenSSL. This unfortunally does not work on RHEL as it always returns the compile time version of OpenSSL.

See https://bugzilla.redhat.com/show_bug.cgi?id=1497859

It turns out we not need these checks at all as we correctly shim all the calls that only exists in certain OpenSSL versions and will fail if these are not supported.

Modifications:

Remove runtime version checks and just depend on the fact that the calls will fail.

Result:

Fixes https://github.com/netty/netty-tcnative/issues/434.


normanmaurer:
@philsttr PTAL this should fix the problem for you.

Issue #387, opened 2018-09-12
Make use of CRYPTO_BUFFER when using BoringSSL


normanmaurer:
We should check if we can make use of CRYPTO_BUFFER when compiling against BoringSSL to improve performance / memory usage:

https://boringssl.googlesource.com/boringssl/+/HEAD/PORTING.md#crypto_buffer

PR #445, opened 2019-03-13
Use CRYPTO_BUFFER when using BoringSSL to reduce memory usage and ove…

normanmaurer:
…rhead of parsing certificates

Motivation:

BoringSSL added support for CRYPTO_BUFFER [1] which allows to reduce the overhead of storing certificates in memory by "de-duplicate" these (as a lot of these are the same across multiple connections). Beside the memory savings using CRYPTO_BUFFER also allows us to reduce the overhead to convert from native SSL certificate representation to the Java based representation as we do not need to serialize from X509 to bytes first all the time.

[1] https://github.com/google/boringssl/blob/chromium-stable/PORTING.md#crypto_buffer

Modifications:

- When using BoringSSL do not use functions that make use of the X509* parts of the library but use the replacements that use CRYPTO_BUFFER
- When using BoringSSL use TLS_with_buffers_method() which guarantees that we do not call any functions that depend on the X509* parts of the lib.

Result:

Less overhead when using BoringSSL

### DIFF
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -1298,14 +1298,21 @@ TCN_IMPLEMENT_CALL(jstring, SSL, getAlpnSelected)(TCN_STDARGS,
 TCN_IMPLEMENT_CALL(jobjectArray, SSL, getPeerCertChain)(TCN_STDARGS,
                                                   jlong ssl /* SSL * */)
 {
-    STACK_OF(X509) *sk;
+#ifdef OPENSSL_IS_BORINGSSL
+    const STACK_OF(CRYPTO_BUFFER) *chain = NULL;
+    const CRYPTO_BUFFER * cert = NULL;
+    const tcn_ssl_ctxt_t* c = NULL;
+#else
+    STACK_OF(X509) *chain = NULL;
+    X509 *cert = NULL;
+    unsigned char *buf = NULL;
+#endif // OPENSSL_IS_BORINGSSL
     int len;
     int i;
-    X509 *cert;
     int length;
-    unsigned char *buf;
-    jobjectArray array;
-    jbyteArray bArray;
+    int offset;
+    jobjectArray array = NULL;
+    jbyteArray bArray = NULL;
     jclass byteArrayClass = tcn_get_byte_array_class();
 
     SSL *ssl_ = J2P(ssl, SSL *);
@@ -1315,37 +1322,66 @@ TCN_IMPLEMENT_CALL(jobjectArray, SSL, getPeerCertChain)(TCN_STDARGS,
     UNREFERENCED(o);
 
     // Get a stack of all certs in the chain.
-    sk = SSL_get_peer_cert_chain(ssl_);
+#ifdef OPENSSL_IS_BORINGSSL
+    c = tcn_SSL_get_app_data2(ssl_);
 
-    len = sk_X509_num(sk);
+    TCN_ASSERT(c != NULL);
+
+    chain = SSL_get0_peer_certificates(ssl_);
+    len = sk_CRYPTO_BUFFER_num(chain);
+
+    if (c->mode == SSL_MODE_SERVER) {
+        // We don't want to include the leaf certificate to mimic the behaviour of SSL_get_peer_cert_chain(...).
+        offset = 1;
+    } else {
+        offset = 0;
+    }
+#else
+    chain = SSL_get_peer_cert_chain(ssl_);
+    len = sk_X509_num(chain);
+    offset = 0;
+#endif // OPENSSL_IS_BORINGSSL
+
+    len -= offset;
     if (len <= 0) {
         // No peer certificate chain as no auth took place yet, or the auth was not successful.
         return NULL;
     }
+
     // Create the byte[][] array that holds all the certs
     array = (*e)->NewObjectArray(e, len, byteArrayClass, NULL);
 
     for(i = 0; i < len; i++) {
-        cert = sk_X509_value(sk, i);
 
-        buf = NULL;
+#ifdef OPENSSL_IS_BORINGSSL
+        cert = sk_CRYPTO_BUFFER_value(chain, i + offset);
+        length = CRYPTO_BUFFER_len(cert);
+#else
+        cert = sk_X509_value(chain, i + offset);
+
         length = i2d_X509(cert, &buf);
         if (length < 0) {
-            if (buf != NULL) {
-                OPENSSL_free(buf);
-            }
+            OPENSSL_free(buf);
             // In case of error just return an empty byte[][]
             return (*e)->NewObjectArray(e, 0, byteArrayClass, NULL);
         }
+#endif // OPENSSL_IS_BORINGSSL
+
         bArray = (*e)->NewByteArray(e, length);
+
+#ifdef OPENSSL_IS_BORINGSSL
+        (*e)->SetByteArrayRegion(e, bArray, 0, length, (jbyte*) CRYPTO_BUFFER_data(cert));
+#else
         (*e)->SetByteArrayRegion(e, bArray, 0, length, (jbyte*) buf);
+        OPENSSL_free(buf);
+        buf = NULL;
+
+#endif // OPENSSL_IS_BORINGSSL
         (*e)->SetObjectArrayElement(e, array, i, bArray);
 
         // Delete the local reference as we not know how long the chain is and local references are otherwise
         // only freed once jni method returns.
         (*e)->DeleteLocalRef(e, bArray);
-
-        OPENSSL_free(buf);
     }
     return array;
 }
@@ -1353,10 +1389,16 @@ TCN_IMPLEMENT_CALL(jobjectArray, SSL, getPeerCertChain)(TCN_STDARGS,
 TCN_IMPLEMENT_CALL(jbyteArray, SSL, getPeerCertificate)(TCN_STDARGS,
                                                   jlong ssl /* SSL * */)
 {
-    X509 *cert;
-    int length;
+#ifdef OPENSSL_IS_BORINGSSL
+    const STACK_OF(CRYPTO_BUFFER) *certs = NULL;
+    const CRYPTO_BUFFER *leafCert = NULL;
+#else
+    X509 *cert = NULL;
     unsigned char *buf = NULL;
-    jbyteArray bArray;
+#endif // OPENSSL_IS_BORINGSSL
+
+    jbyteArray bArray = NULL;
+    int length;
 
     SSL *ssl_ = J2P(ssl, SSL *);
 
@@ -1364,15 +1406,28 @@ TCN_IMPLEMENT_CALL(jbyteArray, SSL, getPeerCertificate)(TCN_STDARGS,
 
     UNREFERENCED(o);
 
-    // Get a stack of all certs in the chain
+#ifdef OPENSSL_IS_BORINGSSL
+    // Get a stack of all certs in the chain, the first is the leaf.
+    certs = SSL_get0_peer_certificates(ssl_);
+    if (certs == NULL || sk_CRYPTO_BUFFER_num(certs) <= 0) {
+        return NULL;
+    }
+    leafCert = sk_CRYPTO_BUFFER_value(certs, 0);
+    length = CRYPTO_BUFFER_len(leafCert);
+#else
     cert = SSL_get_peer_certificate(ssl_);
     if (cert == NULL) {
         return NULL;
     }
 
     length = i2d_X509(cert, &buf);
+#endif // OPENSSL_IS_BORINGSSL
 
     bArray = (*e)->NewByteArray(e, length);
+
+#ifdef OPENSSL_IS_BORINGSSL
+    (*e)->SetByteArrayRegion(e, bArray, 0, length, (jbyte*) CRYPTO_BUFFER_data(leafCert));
+#else
     (*e)->SetByteArrayRegion(e, bArray, 0, length, (jbyte*) buf);
 
     // We need to free the cert as the reference count is incremented by one and it is not destroyed when the
@@ -1381,7 +1436,7 @@ TCN_IMPLEMENT_CALL(jbyteArray, SSL, getPeerCertificate)(TCN_STDARGS,
     X509_free(cert);
 
     OPENSSL_free(buf);
-
+#endif // OPENSSL_IS_BORINGSSL
     return bArray;
 }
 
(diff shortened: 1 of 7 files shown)


normanmaurer:
Also note that I did not call `SSL_CTX_set0_buffer_pool` yet, I wonder if we should make it configurable if a pool should be used or not or if its a good idea to always use a pool anyway. 

/cc @davidben

normanmaurer:
Also note that I did not call `SSL_CTX_set0_buffer_pool` yet, I wonder if we should make it configurable if a pool should be used or not or if its a good idea to always use a pool anyway. 

/cc @davidben

davidben:
Oh wow, nice!

Didn't look at the PR yet but to answer the question on pooling, we enable it on the client stack but not on the server. The thinking was that clients care a lot about memory and don't worry about thread contention. Whereas servers make dramatically more handshakes and really don't want thread contention.

That said, pooling is only used when you receive certificates from the peer. For your own certificates, you already presumably parse them once-ish when you set up the server and then configure them, so there's no need for in-library dedup. That means that pooling doesn't really do much on the server. It's just when you enable client certs.

Though I should say this is more "we never bothered because it seems pointless" than "we tried it and determined it's definitely the wrong thing to do".

normanmaurer:
@davidben ok cool... I guess we may let it disabled then by default and will add a way to let the user turn it on. Even without the pool it will give us some wins tho as we do not need to convert the `X509` to bytes all the time :)

normanmaurer:
@davidben also related to your answer did you consider using a "pool" per thread ? IMHO this should work very well. That said no idea if it is worth it .

davidben:
We hadn't considered that one yet. Though there may still be some contention since the pool would need to be consulted when the `CRYPTO_BUFFER` is released, and the connection may have switched threads by then.

I think servers doing client certs would be better off reducing memory by not retaining the certificate at all and instead remembering just the already-checked "identity" of the peer---the username or email address in the certificate or whatever the application validates. This would need to be integrated with session resumption, since the identity carries over. (We don't currently have such a feature. I've mused on how one might solve this, but there hasn't been a need yet.)

(There's no fundamental reason clients can't do that too. In theory a client only needs to retain the SAN list after verification. But, in practice, people like to [have](https://golang.org/pkg/crypto/tls/#ConnectionState) [APIs](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/HttpsURLConnection.html#getServerCertificates--) to extract the certificate long after the handshake. Maybe it gets routed into random "view certificate" buttons in the browser and such. I've always thought it odd how we retain these moderately-large certificates just for these little-used buttons, but I've also used the buttons on occasion, so I can't look at it too funny. :-) )

normanmaurer:
@davidben hehe fair enough. Let me know once you had a chance to look at the PR in details. I think it should be fine (all 5000+ tests pass in netty with it) but you will know best ;)

normanmaurer:
I have a pending netty update as well that needs to get in. But let me double check

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -2060,13 +2146,33 @@ TCN_IMPLEMENT_CALL(jlong, SSL, parseX509Chain)(TCN_STDARGS, jlong x509ChainBio)
...
         if (sk_X509_push(chain, cert) <= 0) {
+#endif // OPENSSL_IS_BORINGSSL
+
             tcn_Throw(e, "No Certificate specified or invalid format");
             goto cleanup;
         }
+
+#ifndef OPENSSL_IS_BORINGSSL
         cert = NULL;
+#endif // OPENSSL_IS_BORINGSSL
```
trustin: What do you think about duplicating a little bit of common logic instead of ending the `ifdef` block with an unfinished `if` and `while` block?

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -2118,16 +2244,24 @@ TCN_IMPLEMENT_CALL(void, SSL, setKeyMaterial)(TCN_STDARGS, jlong ssl, jlong chai
...
+#ifdef OPENSSL_IS_BORINGSSL
+    for (i = 0; i < numCerts; i++) {
+        certs[i] = sk_CRYPTO_BUFFER_value(cchain, i);
+    }
 
+    if (numCerts <= 0 || SSL_set_chain_and_key(ssl_, certs, numCerts, pkey, NULL) <= 0) {
+#else
     // SSL_use_certificate will increment the reference count of the cert.
     if (numCerts <= 0 || SSL_use_certificate(ssl_, sk_X509_value(cchain, 0)) <= 0) {
+#endif // OPENSSL_IS_BORINGSSL
```
trustin: Ditto

openssl-dynamic/src/main/c/sslutils.c
```diff
@@ -492,20 +500,46 @@ static int SSL_CTX_setup_certs(SSL_CTX *ctx, BIO *bio, bool skipfirst, bool ca)
...
             n++;
         }
+
+#ifdef OPENSSL_IS_BORINGSSL
+        SSL_CTX_set0_client_CAs(ctx, names);
     } else {
+        return -1;
+#else
+    } else {
+
```
trustin: Could be:

```c
#ifdef OPENSSL_IS_BORINGSSL
    SSL_CTX_set0_client_CAs(ctx, names);
#endif
} else {
#ifdef OPENSSL_IS_BORINGSSL
    return -1;
#else
    /* free a perhaps already configured extra chain */
    ...
#endif // OPENSSL_IS_BORINGSSL
}
```

openssl-dynamic/src/main/c/ssl.c
normanmaurer: I actually would prefer to keep the `#ifdef` stuff to ensure we do not miss to fix something because of code duplication.

normanmaurer:
@carl-mastrangelo I just pushed an update ... now this PR should work without the need to change anything in netty at all.

openssl-dynamic/src/main/c/ssl.c
trustin: How does IDE treats you in this case? If IDEs are happy with this, then I think it's OK. But [this one](https://github.com/netty/netty-tcnative/pull/445#discussion_r265394786) still could be addressed since it has no duplicate.

normanmaurer: CLion is happy, but that said I usually use vi for the C stuff ;)

trustin: Haha, OK, then. But could you take care of [this one](https://github.com/netty/netty-tcnative/pull/445#discussion_r265394786) at least?

normanmaurer: @trustin done

openssl-dynamic/src/main/c/sslutils.c
trustin: Thanks!

normanmaurer:
@ryanoneill maybe also interesting for you...

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -2118,16 +2245,24 @@ TCN_IMPLEMENT_CALL(void, SSL, setKeyMaterial)(TCN_STDARGS, jlong ssl, jlong chai
 
     TCN_CHECK_NULL(cchain, chain, /* void */);
 
-    numCerts = sk_X509_num(cchain);
+#ifdef OPENSSL_IS_BORINGSSL
+    for (i = 0; i < numCerts; i++) {
+        certs[i] = sk_CRYPTO_BUFFER_value(cchain, i);
+    }
 
+    if (numCerts <= 0 || SSL_set_chain_and_key(ssl_, certs, numCerts, pkey, NULL) <= 0) {
```
normanmaurer: Once https://github.com/netty/netty-tcnative/pull/446 is merged we need to ensure to pass in the "private key method" if pkey is `NULL` here

normanmaurer:
@davidben do you think you can do a code-review as well ?

normanmaurer:
Fixes #387

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1315,64 +1322,112 @@ TCN_IMPLEMENT_CALL(jobjectArray, SSL, getPeerCertChain)(TCN_STDARGS,
...
+            OPENSSL_free(buf);
             // In case of error just return an empty byte[][]
             return (*e)->NewObjectArray(e, 0, byteArrayClass, NULL);
         }
+#endif // OPENSSL_IS_BORINGSSL
+
         bArray = (*e)->NewByteArray(e, length);
+
+#ifdef OPENSSL_IS_BORINGSSL
+        (*e)->SetByteArrayRegion(e, bArray, 0, length, (jbyte*) CRYPTO_BUFFER_data(cert));
```
davidben: There are several instances of this ifdef. You could probably write a single function which takes an `X509` (OpenSSL) / `CRYPTO_BUFFER` (BoringSSL) and calls `SetByteArrayRegion` on the bytes within.

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -1766,8 +1848,19 @@ static int cert_requested(SSL* ssl, X509** x509Out, EVP_PKEY** pkeyOut) {
 
     tcn_get_java_env(&e);
 
-    types = keyTypes(e, ssl);
-    issuers = principalBytes(e,  SSL_get_client_CA_list(ssl));
+ if (c->mode == SSL_MODE_SERVER) {
+        // TODO: Consider filling these somehow.
```
davidben: Client's don't generally send servers a CA list in TLS. The `SSL_get_client_CA_list` function, as a server, would return back the configured list, rather than the list from the peer, which is a fairly different operation from what you want here.

TLS 1.3 does generalize the mechanism to work in both directions, but I'm not aware of anyone actually using it. (We don't process the extension in BoringSSL.) Usually the client has no end of supported CAs and the server already knows which to use for a given SNI, so there's not much negotiation to be done.

Moreover, although it's called `cert_requested`, this callback is used with `SSL_CTX_set_client_cert_cb`, which:
a) Is only used on a client anyway
b) Isn't what you want for `TLS_with_buffers_method` since it depends on `X509`.

I'm guessing this is why you had to build with `NDEBUG`? It fires the assert because you're not supposed to use it here. :-P You won't get the binary size benefits and, in some cases (though I think not this one right now), might end up leaking because `SSL_CTX`, etc., objects made from a non-`X509` `SSL_METHOD` don't know how to release `X509` objects hanging off of them.

You probably want `SSL_CTX_set_cert_cb` which works more uniformly (client/server, no X509 dependency, lets you set a cert chain, etc). (We actually emulate `SSL_CTX_set_client_cert_cb` using `SSL_CTX_set_cert_cb` anyway.)

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1843,17 +1902,23 @@ TCN_IMPLEMENT_CALL(void, SSL, setHostNameValidation)(TCN_STDARGS, jlong ssl, jin
 
     TCN_CHECK_NULL(ssl_, ssl, /* void */);
 
+#ifdef OPENSSL_IS_BORINGSSL
+    if (flags != 0) {
+        tcn_ThrowException(e, "flags must be 0");
+    }
+    // Let's just ignore this as it is done in the Java level anyway.
```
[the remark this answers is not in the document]
normanmaurer: @davidben yep thats fine as the java level does all of it. We will remove the method in the future.

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1905,6 +1972,9 @@ TCN_IMPLEMENT_CALL(void, SSL, setCertificateBio)(TCN_STDARGS, jlong ssl,
                                                          jlong cert, jlong key,
                                                          jstring password)
 {
+#ifdef OPENSSL_IS_BORINGSSL
+    tcn_Throw(e, "Not supported using BoringSSL");
```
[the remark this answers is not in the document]
normanmaurer: @davidben this method is never used by us when `BoringSSL` is used so I did not bother to implement it just for the sake to have the code ;)

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -2118,16 +2244,24 @@ TCN_IMPLEMENT_CALL(void, SSL, setKeyMaterial)(TCN_STDARGS, jlong ssl, jlong chai
...
         ERR_error_string_n(ERR_get_error(), err, ERR_LEN);
         ERR_clear_error();
         tcn_Throw(e, "Error setting certificate (%s)", err);
         return;
     }
 
+#ifndef OPENSSL_IS_BORINGSSL
     if (pkey != NULL) {
         // SSL_use_PrivateKey will increment the reference count of the key.
         if (SSL_use_PrivateKey(ssl_, pkey) <= 0) {
```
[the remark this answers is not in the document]
normanmaurer: thanks will check and maybe do a followup PR for this then.

openssl-dynamic/src/main/c/ssl.c
normanmaurer: good idea will do.

openssl-dynamic/src/main/c/sslcontext.c
normanmaurer: @davidben this function is never used with BoringSSL, it is just used when compiling against openssl < 1.0.2. 

I did compile with `NDEBUG` to reduce some overhead, it also works without specify it. Do you say I should remove this option ?

normanmaurer:
@danbev I was mostly interested to have you take a look if I correctly handle the life-cycle of `CRYPTO_BUFFER` in generally. Does what I do here look reasonable ?

openssl-dynamic/src/main/c/sslcontext.c
normanmaurer: sorry.. I wanted to say we never call `setCertRequestedCallback` these days as we replaced it with `SSL_CTX_set_cert_cb `. Let me just remove this all together.

openssl-dynamic/src/main/c/ssl.c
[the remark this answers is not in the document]
normanmaurer: I am mostly interested in not invoking the parser atm

openssl-dynamic/src/main/c/ssl.c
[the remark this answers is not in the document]
normanmaurer: @davidben also just to make sure I understand it correct... if I would like to do the hostname check in the JNI level I would need to basically hand-roll all of it. Correct ?

openssl-dynamic/src/main/c/sslcontext.c
davidben: > I did compile with `NDEBUG` to reduce some overhead, it also works without specify it. Do you say I should remove this option?

I guess it's `assert`, so you probably want to build with it in release mode and without it in debug mode.

openssl-dynamic/src/main/c/ssl.c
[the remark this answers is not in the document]
davidben: > @davidben also just to make sure I understand it correct... if I would like to do the hostname check in the JNI level I would need to basically hand-roll all of it. Correct ?

Sadly, yeah. :-/ `CRYPTO_BUFFER` was originally designed for Chromium case where we have an external OS verifier to call out to, so it doesn't have any cert verification logic itself. Conscrypt similarly goes through Java bits for cert verification, so it worked for them too. (I'm not familiar with how Netty works.)

We definitely want to have a verifier available that is less of a mess than `X509`, but that's still pending. :-/ The long-term plan is to move the new verifier being developed in Chromium into BoringSSL, but that hasn't happened yet.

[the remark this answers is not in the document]
normanmaurer: @davidben ok makes sense... netty also goes through the Java level, so this is just a left-over. Was just interested ...

davidben:
> I was mostly interested to have you take a look if I correctly handle the life-cycle of CRYPTO_BUFFER in generally. Does what I do here look reasonable ?

Yeah, I didn't notice any lifetime issues, though it was rather a lot of ifdef-y code, so hopefully I didn't miss anything.

Do you all test with ASan? (Specifically LSan.) ASan is wonderful and really useful for checking that sort of thing.

normanmaurer:
@davidben ok cool... I never used ASan with JNI code. Let me do some research how to do this tho. This is a good idea.

PR #447, opened 2019-03-14
correct init all fields of tcn_ssl_task_t in all cases.

normanmaurer:
Motivation:

We missed to correctly init all fields of tcn_ssl_task_t and so may run into problem if OPENSSL_malloc(...) does not memset when allocating.

Modifications:

- Add two helper functions, one to init tcn_ssl_task_t and one to free tcn_ssl_task_t pointers to ensure we never fail to init all fields again.
- Use these functions

Result:

Always correctly init all fields of tcn_ssl_task_t.


normanmaurer:
That should be an easy one

Issue #448, opened 2019-03-20
IBM 8 + BoringSSL and client certificate signature again


kwart:
This is a follow up for issue #404.

We experience issues when running on IBM Java (we use version 8) with BoringSSL provider on the server side and JSSE engine on the client side.
The problems started with netty version `4.1.31.Final` and are still present in `4.1.34.Final`. Nevertheless, in my opinion it's only a side effect of a problem in the `netty-tcnative-boringssl-static`.
The native `SSL.getSigAlgs(long)` call returns `null` for some reason.

With SSL debug logging enabled I see following entries in the log:
```
ssl: Ignoring alias key: signature does not conform to negotiated signature algorithms
ssl: Ignoring alias key: signature does not conform to negotiated signature algorithms
ssl: Ignoring alias key: signature does not conform to negotiated signature algorithms
ssl: Ignoring alias key: signature does not conform to negotiated signature algorithms
ssl: Ignoring alias key: signature does not conform to negotiated signature algorithms
ssl: Ignoring alias key: signature does not conform to negotiated signature algorithms
```

I've built the latest netty `4.1.35.Final-SNAPSHOT` and tcnative  `2.0.23.Final-SNAPSHOT`, but the problem is still there.

When I switch from BoringSSL (`netty-tcnative-boringssl-static`) to the dynamically linked OpenSSL (`netty-tcnative`). Then the SSL handshake finishes successfully.

Issue reproducer is available in the following branch:
https://github.com/kwart/test-app/tree/netty-reproducer-boringssl-ibm8
it's based on `EchoServer` netty example.

normanmaurer:
@kwart so you are saying it also not work when you use the exact same versions that reported it fixed in #404 ?

kwart:
The fix for #404 mentions netty 4.0.32.Final and netty-tcnative 2.0.20.Final. This issue is present in that combination.

On the other hand, the issue #404 is reported against combination 4.1.30.Final + 2.0.18.Final. This issue is not reproducible on that one.

normanmaurer:
@kwart ok got it... let me have look.

normanmaurer:
@kwart ok I found the problem... working on a fix + unit test now :)

normanmaurer:
Should be fixed by https://github.com/netty/netty-tcnative/pull/449. I will add a unit test to netty itself.

PR #449, opened 2019-03-21
Skip unknown signature algorithms when when collect signatures in SSL…

normanmaurer:
….getSigAlgs(...)

Motivation:

We did not correctly account for unknown signature algorithms when using BoringSSL but just returned NULL if we accountered one (and discard all the known salgorithms). We should just ignore the unknown and return the others.

Modifications:

Skip unkown signature algorithms and correctly return the known.

Result:

Fixes https://github.com/netty/netty-tcnative/issues/448.


normanmaurer:
@kwart this should fix your problem. Can you verify ?

kwart:
I'm going to check it with your branch.

kwart:
The `EchoServer` reproducer works correctly with the fix. Let me check it also in the Hazelcast Enterprise testsuite where we originally hit the issue.

kwart:
The Hazelcast tests passed too :+1:
Thanks for the fix. Looking forward for the new release.

normanmaurer:
This is the unit test for it https://github.com/netty/netty/pull/8965

normanmaurer:
@kwart thanks for testing the fix :)

PR #454, opened 2019-04-02
Use OpenSSL_malloc(...) to allocate array as not all compilers suppor…

normanmaurer:
…t our code otherwise

Motivation:

913df47adc07c9b8b083cdb92789356d3e864e55 introduced code that allocated an array like `CRYPT_BUFFER*[num]` but num is not know at compile time. This does not work on all compilers, so we must use OpenSSL_malloc(...) / OpenSSL_free(...)

Modifications:

Adjust code to use OpenSSL_malloc / OpenSSL_free

Result:

Be able to build on windows as well again.


normanmaurer:
As is basically the same as https://github.com/netty/netty-tcnative/commit/10ff65992a04ff6e875f8148794664e65d19e503 I will just merge it.

PR #457, opened 2019-04-11
Fix possible segfault when calling SSLContext.setCertificateCallback …

normanmaurer:
…and compiled via GCC

Motivation:

When using GCC and we use weak references to detect if we can call the function to be able to use different versions of openssl during runtime. Unfortunally we missed to return after putting an exception on the stack and so cause a segfault if we later on try to call the function which was declared as weak.

Modifications:

Add missing return after putting the exception on the stack

Result:

No more segfault if SSL_CTX_set_cert_cb not exists and GCC is used.


normanmaurer:
@ejona86 🤣 same

PR #458, opened 2019-04-17
Avoid using un-thread-safe static buffers when printing errors

normanmaurer:
Motivation:

ERR_err_string may return an internal shared buffer and so is not thread safe. While our usage seems to be fine (as we always pass in a buffer by ourself we should better just use ERR_err_string_n which will never return a shared buffer.

Modifications:

Replace ERR_err_string with ERR_err_string_n

Result:

More correct usage of APIs to print errors in a lib that may be used by multiple threads.


normanmaurer:
Related to https://github.com/apple/swift-nio-ssl/pull/102 ... Thanks @Lukasa for the idea :)

PR #460, opened 2019-04-29
Update to latest java8 release

normanmaurer:
Motivation:

There was a new java8 release, we should use it.

Modifications:

Update to adopt@1.8.212-03

Result:

Use latest java8 version on the CI


normanmaurer:
Let me just merge this as it is just a java version bump

PR #461, opened 2019-05-12
 Fixed the artifacts causing the native libs to not load

nomstradamus:
OpenSSl.java 

```java
    private static void loadTcNative() throws Exception {
        String os = PlatformDependent.normalizedOs();
        String arch = PlatformDependent.normalizedArch();

        Set<String> libNames = new LinkedHashSet<String>(4);
        String staticLibName = "netty_tcnative";

        // First, try loading the platform-specific library. Platform-specific
        // libraries will be available if using a tcnative uber jar.
        libNames.add(staticLibName + "_" + os + '_' + arch);
        if ("linux".equalsIgnoreCase(os)) {
            // Fedora SSL lib so naming (libssl.so.10 vs libssl.so.1.0.0)..
            libNames.add(staticLibName + "_" + os + '_' + arch + "_fedora");
        }
        libNames.add(staticLibName + "_" + arch);
        libNames.add(staticLibName);

        NativeLibraryLoader.loadFirstAvailable(SSL.class.getClassLoader(),
            libNames.toArray(new String[0]));
    }

```
Here the problem is that the staticLibName is "netty_tcnative" however the maven build generates the jars with a "-" which causes it to not find the native libraries.

### DIFF
--- a/boringssl-static/pom.xml
+++ b/boringssl-static/pom.xml
@@ -21,7 +21,7 @@
     <artifactId>netty-tcnative-parent</artifactId>
     <version>2.0.26.Final-SNAPSHOT</version>
   </parent>
-  <artifactId>netty-tcnative-boringssl-static</artifactId>
+  <artifactId>netty_tcnative_boringssl_static</artifactId>
   <packaging>jar</packaging>
 
   <name>Netty/TomcatNative [BoringSSL - Static]</name>
--- a/openssl-dynamic/pom.xml
+++ b/openssl-dynamic/pom.xml
@@ -21,7 +21,7 @@
     <artifactId>netty-tcnative-parent</artifactId>
     <version>2.0.26.Final-SNAPSHOT</version>
   </parent>
-  <artifactId>netty-tcnative</artifactId>
+  <artifactId>netty_tcnative</artifactId>
   <packaging>jar</packaging>
 
   <name>Netty/TomcatNative [OpenSSL - Dynamic]</name>
--- a/openssl-static/pom.xml
+++ b/openssl-static/pom.xml
@@ -21,7 +21,7 @@
     <artifactId>netty-tcnative-parent</artifactId>
     <version>2.0.26.Final-SNAPSHOT</version>
   </parent>
-  <artifactId>netty-tcnative-openssl-static</artifactId>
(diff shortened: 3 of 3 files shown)


slandelle:
@nomstradamus You're confusing the name of the jar, eg [`netty-tcnative-boringssl-static-2.0.25.Final-linux-x86_64.jar`](https://search.maven.org/artifact/io.netty/netty-tcnative-boringssl-static/2.0.25.Final/jar) and the name of the native lib it contains, eg `libnetty_tcnative.so` (check in `META-INF/native`). Your proposed fix is wrong and things work as expected.

nomstradamus:
I am not confusing it . Please try with and without the patch and you will see it will not load the native library. I spent quite a bit of time just trying to get it to work.

```java
import io.netty.handler.ssl.OpenSsl;

import java.util.Set;

public class NettyFixed {


    public static void main(String[] args) {
        Set<String> strings = OpenSsl.availableOpenSslCipherSuites();
        OpenSsl.ensureAvailability();
        boolean available = OpenSsl.isAvailable();
        System.out.println(available);
        System.out.println(strings);
    }
}
```
```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>

    <groupId>nettyfixed</groupId>
    <artifactId>com.intu.native.speed</artifactId>
    <version>1.0-SNAPSHOT</version>
<dependencies>

    <dependency>
        <groupId>org.apache.tomcat</groupId>
        <artifactId>tomcat-jni</artifactId>
        <version>8.0.1</version>
    </dependency>


    <dependency>
        <groupId>log4j</groupId>
        <artifactId>log4j</artifactId>
        <version>1.2.16</version>
    </dependency>

    <dependency>
        <groupId>io.netty</groupId>
        <artifactId>netty-all</artifactId>
        <version>4.1.31.Final</version>
    </dependency>

</dependencies>

</project>
```

Try this project by include the generated jar and you'll see .

PR #468, opened 2019-06-03
Adjust ifdef for libressl 2.9.2

normanmaurer:
Motiviation:

We recently upgraded to use libressl 2.9.2 when building our static linked version against libressl. We should also ensure we enable / disable features based on this version.

Modifications:

Adjust ifdef for libressl 2.9.2

Result:

Use correct features for libressl 2.9.2


normanmaurer:
@ejona86 we have a different idea of fun 🤣

Issue #235, opened 2017-02-14
LibreSSL 2.5.x updates


Scottmitch:
A hack was introduced for LibreSSL 2.4.x to work around LibreSSL not supporting `X509_V_ERR_UNSPECIFIED` [1]. This hack should be removed when we updated to LibreSSL 2.5.x.

[1] https://github.com/netty/netty-tcnative/pull/234

normanmaurer:
I think this can be closed now

PR #469, opened 2019-06-08
Remove hack used to support old versions of libressl

normanmaurer:
Motivation:

At some point we introduced a hack to be able to compile against libressl 2.4.x. As we now use libressl 2.9 we can just remove the hack and cleanup the code-base.

See https://github.com/netty/netty-tcnative/pull/234.

Modifications:

Remove hack and just use X509_V_ERR_UNSPECIFIED directly.

Result:

Fixes https://github.com/netty/netty-tcnative/issues/235

### DIFF
--- a/openssl-dynamic/src/main/c/sslcontext.c
+++ b/openssl-dynamic/src/main/c/sslcontext.c
@@ -1508,11 +1508,7 @@ static int SSL_cert_verify(X509_STORE_CTX *ctx, void *arg) {
     JNIEnv *e = NULL;
     jstring authMethodString = NULL;
     int ret = 0;
-#ifdef X509_V_ERR_UNSPECIFIED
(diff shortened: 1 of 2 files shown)


openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -1508,11 +1508,7 @@ static int SSL_cert_verify(X509_STORE_CTX *ctx, void *arg) {
     JNIEnv *e = NULL;
     jstring authMethodString = NULL;
     int ret = 0;
-#ifdef X509_V_ERR_UNSPECIFIED
     jint result = X509_V_ERR_UNSPECIFIED;
```
ejona86: Won't this be undefined for "openssl 1.0.2 and earlier"?

normanmaurer:
Nevermind ... I did screw this up. Lets close it.

Issue #40, opened 2015-05-06
Review usage of malloc


normanmaurer:
We should review the usage of malloc and if we check if allocation failed.

Scottmitch:
`OPENSSL_malloc` while we are at it.

normanmaurer:
@Scottmitch +1 :)

johnou:
https://github.com/netty/netty-tcnative/commit/abded8736ac28d6044d1cd8439187f5af0ede0ce

Scottmitch:
FWIW I don't think abded8736ac28d6044d1cd8439187f5af0ede0ce is necessary and may be confusing (openssl will not take ownership of these elements).

PR #487, opened 2019-09-09
Always correctly handle the case when malloc(...) returns NULL

normanmaurer:
Motivation:

malloc(...) may return NULL when there is not enough memory left. We need to handle this cases.

Modifications:

Add missing == NULL checks when using malloc

Result:

Fixes https://github.com/netty/netty-tcnative/issues/40


normanmaurer:
actually do not review yet ... there is some stuff still missing

normanmaurer:
Alright this should be done now :)

normanmaurer:
@ejona86 yeah :/

PR #488, opened 2019-09-09
Correctly free source pointer when realloc(...) fails

normanmaurer:
Motivation:

When realloc(...) fails and so return NULL we need to ensure we free the original (source) pointer to not leak memory.

Modifications:

Correctly call free(...) when realloc(...) fails.

Result:

No more memory leak when realloc fails.


normanmaurer:
Did notice this one when looking through the code :) All the joys of c are mine ;)

PR #489, opened 2019-09-10
Correctly free memory when loading of classes / fields / methods fail…

normanmaurer:
…s during init

Motivation:

We did not correctly free the memory we allocated before on the heap in the case of:

 - failed to load classes
 - failed to find field
 - failed to find method

While this is very unlikely we should still ensure we free all memory.

Modifications:

- Introduce new macros to make it easier to free memory
- Use the macros

Result:

Correctly free allocated memory in all cases

### DIFF
--- a/openssl-dynamic/src/main/c/tcn.h
+++ b/openssl-dynamic/src/main/c/tcn.h
@@ -133,10 +133,13 @@ jstring         tcn_new_stringn(JNIEnv *, const char *, size_t);
         jclass _##C = (*(E))->FindClass((E), N);    \
         if (_##C == NULL) {                         \
             (*(E))->ExceptionClear((E));            \
-            return R;                               \
+            goto R;                                 \
         }                                           \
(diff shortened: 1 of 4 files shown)


openssl-dynamic/src/main/c/tcn.h
```diff
@@ -133,10 +133,13 @@ jstring         tcn_new_stringn(JNIEnv *, const char *, size_t);
         jclass _##C = (*(E))->FindClass((E), N);    \
         if (_##C == NULL) {                         \
             (*(E))->ExceptionClear((E));            \
-            return R;                               \
+            goto R;                                 \
         }                                           \
         C = (*(E))->NewGlobalRef((E), _##C);        \
         (*(E))->DeleteLocalRef((E), _##C);          \
+        if (C == NULL) {                            \
```
normanmaurer: We missed this check completely before.

PR #490, opened 2019-09-10
Correctly handle the case of NewGlobalRef(...) returning NULL because…

normanmaurer:
… of no memory left

Motivation:

We did not correctly all cases of NewGlobalRef(...) returning NULL and so may fall into some "invalid state". Beside this we also did not correctl NULL out fields in some cases.

Modifications:

Correctly handle NewGlobalRef(...) in all cases

Result:

Ensure we take into account when NewGlobalRef(...) returns NULL.

### DIFF
--- a/openssl-dynamic/src/main/c/sslcontext.c
+++ b/openssl-dynamic/src/main/c/sslcontext.c
@@ -1398,8 +1398,8 @@ static tcn_ssl_task_t* tcn_ssl_task_new(JNIEnv* e, jobject task) {
     if (sslTask == NULL) {
         return NULL;
     }
-    sslTask->task = (*e)->NewGlobalRef(e, task);
-    if (sslTask->task == NULL) {
+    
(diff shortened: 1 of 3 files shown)


normanmaurer:
And another one...

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -2343,16 +2383,27 @@ TCN_IMPLEMENT_CALL(void, SSLContext, setSniHostnameMatcher)(TCN_STDARGS, jlong c
         jclass matcher_class = (*e)->GetObjectClass(e, matcher);
         jmethodID method = (*e)->GetMethodID(e, matcher_class, "match", "(JLjava/lang/String;)Z");
         if (method == NULL) {
-            c->sni_hostname_matcher_method = NULL;
```
normanmaurer: This was just not correct at all.

normanmaurer:
Let me merge this as I have a followup....

PR #493, opened 2019-09-18
Share code to free / release tcn_ssl_task* instances

normanmaurer:
Motivation:

We should share the code that is used to free / destory tcn_ssl_task instances to reduce the risk of bugs

Modifications:

Re-use code and so remove duplicates

Result:

Less error-prone code

### DIFF
--- a/openssl-dynamic/src/main/c/ssl_private.h
+++ b/openssl-dynamic/src/main/c/ssl_private.h
@@ -353,6 +353,9 @@ struct tcn_ssl_task_t {
     jobject task;
 };
 
+tcn_ssl_task_t* tcn_ssl_task_new(JNIEnv*, jobject);
+void tcn_ssl_task_free(JNIEnv*, tcn_ssl_task_t*);
+
 /*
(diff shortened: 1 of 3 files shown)


normanmaurer:
@johnou PTAL as well

normanmaurer:
... this is just some cleanup

openssl-dynamic/src/main/c/ssl_private.h
```diff
@@ -353,6 +353,9 @@ struct tcn_ssl_task_t {
     jobject task;
 };
 
+tcn_ssl_task_t* tcn_ssl_task_new(JNIEnv*, jobject);
```
johnou: didn't use tcn_ssl_task_new in the end?

normanmaurer: its only used by `SSLContext.c` atm in which we also define it. That said I thought it is cleaner to have the new and free here :)

normanmaurer:
@johnou so good to go ?

PR #498, opened 2019-09-24
Consolidate state management to make things more clean and less error…

normanmaurer:
…-prone

Motivation:

At the moment we handle "state" in a fragemented way which makes things quite error prone in general and also makes things harder to maintain.

Modifications:

- Just store all extra state in a struct to consolidate things
- Rename function to make it more explict what it does
- Remove functions which are not needed anymore

Result:

Cleaner and less error-prone code

### DIFF
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -829,8 +829,8 @@ TCN_IMPLEMENT_CALL(jint, SSL, initialize)(TCN_STDARGS, jstring engine)
     }
 #endif
 
-    // For tcn_SSL_get_app_data*() at request time
-    tcn_SSL_init_app_data_idx();
+    // For tcn_SSL_get_app_state() at request time
+    tcn_SSL_init_app_state_idx();
 
 #if OPENSSL_VERSION_NUMBER >= 0x10100000L && !defined(LIBRESSL_VERSION_NUMBER)
     init_bio_methods();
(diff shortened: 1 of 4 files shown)


openssl-dynamic/src/main/c/ssl.c
```diff
@@ -889,21 +889,57 @@ TCN_IMPLEMENT_CALL(jint, SSL, getLastErrorNumber)(TCN_STDARGS) {
...
+static tcn_ssl_state_t* new_ssl_state(tcn_ssl_ctxt_t* ctx) {
+    if (ctx == NULL) {
+        return NULL;
+    }
+
+    tcn_ssl_state_t* state = OPENSSL_malloc(sizeof(tcn_ssl_state_t));
+    if (state == NULL) {
+        return NULL;
+    }
+    memset(state, 0, sizeof(tcn_ssl_state_t));
```
carl-mastrangelo: Can you use `OPENSSL_zalloc` instead?

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -889,21 +889,57 @@ TCN_IMPLEMENT_CALL(jint, SSL, getLastErrorNumber)(TCN_STDARGS) {
...
+    state->ctx = ctx;
+
+     // Initially we will share the configuration from the SSLContext.
+    state->verify_config = &ctx->verify_config;
+    // Store the handshakeCount in the SSL instance.
+    state->handshakeCount = 0;
```
carl-mastrangelo: this should be implicit in the memset above right?

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1524,37 +1528,33 @@ TCN_IMPLEMENT_CALL(jlong, SSL, setTimeout)(TCN_STDARGS, jlong ssl, jlong seconds
 
 TCN_IMPLEMENT_CALL(void, SSL, setVerify)(TCN_STDARGS, jlong ssl, jint level, jint depth)
 {
-    tcn_ssl_verify_config_t* verify_config;
-    tcn_ssl_ctxt_t* c;
+    tcn_ssl_state_t* state;
```
carl-mastrangelo: Optional: `= NULL` like on line 1900

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1897,15 +1897,17 @@ TCN_IMPLEMENT_CALL(jbyteArray, SSL, getSessionId)(TCN_STDARGS, jlong ssl)
...
     SSL *ssl_ = J2P(ssl, SSL *);
 
     TCN_CHECK_NULL(ssl_, ssl, 0);
 
     UNREFERENCED(o);
 
-    handshakeCount = tcn_SSL_get_app_data3(ssl_);
-    return handshakeCount != NULL ? *handshakeCount : 0;
+    if ((state = tcn_SSL_get_app_state(ssl_)) == NULL) {
+        return state->handshakeCount;
```
carl-mastrangelo: wont state be NULL here by the condition?

openssl-dynamic/src/main/c/ssl_private.h
```diff
@@ -356,23 +356,30 @@ struct tcn_ssl_task_t {
...
+    tcn_ssl_ctxt_t *ctx;
+    tcn_ssl_task_t* ssl_task;
+    tcn_ssl_verify_config_t* verify_config;
+};
+
+#define TCN_GET_SSL_CTX(ssl, C)                             \
+    TCN_BEGIN_MACRO                                         \
+        tcn_ssl_state_t* _S = tcn_SSL_get_app_state(ssl);   \
+        if (_S == NULL) {                                   \
+            C = NULL;                                       \
```
carl-mastrangelo: I also dont understand this.   If _S is null, wont it still be derefenced unconditionally below?

openssl-dynamic/src/main/c/ssl.c
normanmaurer: as far as I know this was just added in 1.1.0 so to not add more ifdefs we should just memset by ourselves.

openssl-dynamic/src/main/c/ssl.c
normanmaurer: doh... good catch. This should be `!= NULL`

openssl-dynamic/src/main/c/ssl_private.h
normanmaurer: Good catch... missed an else :(

normanmaurer:
@carl-mastrangelo PTAL again

Issue #475, opened 2019-07-19
Detect ninja executable alternate name


morvael:
In my system, CentOS 7.6, ninja executable installed in standard way is actually called ninja-build. It would be nice if pom.xml of boringssl-static subproject would be able to detect and handle that.

normanmaurer:
@morvael PTAL. https://github.com/netty/netty-tcnative/pull/477

morvael:
Project built without problems.

normanmaurer:
@morvael thanks a lot for verify it

morvael:
BTW, I was able to install and run ppc64le system on a "normal" Intel machine using Qemu on my Linux Mint. It isn't fast, but it does it job. So it should be possible to build such version of native library somewhere on Netty project's machines :)

normanmaurer:
@morvael maybe you can write this up ?

morvael:
You mean a tutorial to be placed on project wiki? That would be of help to some other people compiling Netty, but would not result in stock netty-all from Maven having proper .so inside.

I'm not a qemu veteran, so now I'm struggling with sharing files between guest and host (recommended 9p filesystem approach is impossible in CentOS 7, and I would have to reconfigure network interface emulation to connect via ssh)... So for now I'm unable to pull the compiled library from my virtual machine :-)

morvael:
04:24 h to build tcnative using qemu'ed ppc64le :)

PR #514, opened 2019-11-29
Use ninjia-build if possible

wangxiyuan:
Similar with https://github.com/netty/netty-tcnative/issues/475, on CentOS 7, ninjia is called ninjia-build, try to use `ninjia-build` if possible.


wangxiyuan:
Oh, duplicated with https://github.com/netty/netty-tcnative/pull/513

PR #524, opened 2020-02-13
Use git 2.x on centos6 and update maven scm plugin

normanmaurer:
Motivation:

To be able to use the maven scm plugin with a specific revision of a repository we need to have git 2.x installed.

Modifications:

- Ensure we install git 2.x when on centos 6 (as it ships with 1.7)
- Update scm plugin

Result:

Be able to checkout specific revision with scm plugin


normanmaurer:
@netty-bot test this please

PR #525, opened 2020-02-14
Use Brewfile to manage macOS build dependencies via Homebrew

jchambers:
This change adds a simple Brewfile to make it easier to install macOS build dependencies via Homebrew. After this change, folks can install all of the dependencies via `brew bundle`. This also opens the door to requiring specific versions of some packages in the future.

My immediate use case for this is actually a little subtle. I'm trying to build `netty-tcnative` on a CI system under various versions of macOS. The varying CI host images come with different subsets of the build dependencies already installed. `brew install a b c` will actually return a non-success exit code if `a`, `b`, or `c` are already installed, and that can cause CI systems to declare the whole build a failure.

`brew bundle`, by contrast, reports success as long as all of the dependencies are satisfied by the time it exits regardless of whether it actually installed all of the packages, some of the packages, or none of the packages.


normanmaurer:
@jchambers thanks a lot

jchambers:
Awesome! I [updated the wiki](https://github.com/netty/netty/wiki/Forked-Tomcat-Native/_compare/14782cf873650620eba866042f7afdf7e9bcd5c5...49d90e56bc373d627485f18d1a64cd636777f90e) to reflect this change. Cheers!

normanmaurer:
@jchambers <3

PR #526, opened 2020-02-14
Explicit specify build target for boringssl

normanmaurer:
Motivation:

Some tests do not compile on centos6 after the latest update of the chromium-stable branch. To workaround this we just dont compile the tests anymore for boringssl (we didnt run these anyway).

Modifications:

Explicit specify targets when building boringssl

Result:

Be able to build boringssl-static again on cents 6


normanmaurer:
This is also related to https://boringssl-review.googlesource.com/c/boringssl/+/40067

Issue #532, opened 2020-03-17
Possible memory leak in ssl.c


creamsoup:
After PR #498, AddressSanitizer test starts to fail.

```
ERROR: LeakSanitizer: detected memory leaks

Too many leaks! Only the first 5000 leaks encountered will be reported.
Direct leak of 40 byte(s) in 1 object(s) allocated from:
    #0 0x56273d7f14dd in malloc llvm-project/compiler-rt/lib/asan/asan_malloc_linux.cpp:145:3
    #1 0x56274231e008 in OPENSSL_malloc boringssl/src/crypto/mem.c:112:15
    #2 0x56273e99d081 in new_ssl_state netty_tcnative/openssl-dynamic/src/main/c/ssl.c:905:30
    #3 0x56273e99d081 in netty_internal_tcnative_SSL_newSSL netty_tcnative/openssl-dynamic/src/main/c/ssl.c:951:18
    #4 0x7f62f3967282  (<unknown module>)
    #5 0x7f62f3955bcf  (<unknown module>)
    #6 0x7f62f39560f5  (<unknown module>)
    #7 0x7f62f39560f5  (<unknown module>)
    #8 0x7f62f3955e3f  (<unknown module>)
    #9 0x7f62f3955e3f  (<unknown module>)
    #10 0x7f62f3955e3f  (<unknown module>)
    #11 0x7f62f395613a  (<unknown module>)
    #12 0x7f62f39560f5  (<unknown module>)
    #13 0x7f62f39560f5  (<unknown module>)
    #14 0x7f62f3955e3f  (<unknown module>)
    #15 0x7f62f3955f26  (<unknown module>)
    #16 0x7f62f39560f5  (<unknown module>)
    #17 0x7f62f39560f5  (<unknown module>)
    #18 0x7f62f395613a  (<unknown module>)
    #19 0x7f62f39560f5  (<unknown module>)
    #20 0x7f62f39560f5  (<unknown module>)
```

normanmaurer:
Interesting... will have a look

normanmaurer:
@creamsoup PTAL https://github.com/netty/netty-tcnative/pull/533

PR #533, opened 2020-03-18
Correctly free tcn_ssl_state_t to prevent memory leak

normanmaurer:
Motivation:

We failed to free tcn_ssl_state_t via OPENSSL_free(...) which produced a small memory leak. This regression was introduced by https://github.com/netty/netty-tcnative/pull/498.

Modifications:

Correctly call OPENSSL_free(...) on previous allocated tcn_ssl_state_t.

Result:

Fixes https://github.com/netty/netty-tcnative/issues/532.

### DIFF
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -918,6 +918,11 @@ static void free_ssl_state(tcn_ssl_state_t* state) {
     tcn_get_java_env(&e);
     tcn_ssl_task_free(e, state->ssl_task);
     state->ssl_task = NULL;
+
+    // Free the tcn_ssl_state_t itself as it was allocated via OPENSSL_malloc(...) before
+    //
+    // https://github.com/netty/netty-tcnative/issues/532
+    OPENSSL_free(state);
 }
 
 TCN_IMPLEMENT_CALL(jlong /* SSL * */, SSL, newSSL)(TCN_STDARGS,


normanmaurer:
@creamsoup is this something we could run on out CI as well ?

creamsoup:
I am a middle man here, so take it with grain of salt. 
I believe you can set it up since the address sanitizer is an open source project (https://github.com/google/sanitizers/wiki/AddressSanitizer). the test I ran is a simple tls ping pong with address sanitizer. 
iiuc, it is not useful for java in general. but it often finds bugs in jni code.

normanmaurer:
@creamsoup let me see if I can find some docs on how I could hook this up with Java/JNI. If you can provide any pointers that would be great tho :)

normanmaurer:
I will cut a new release tomorrow so we consume this.

johnou:
@creamsoup out of interest did you also try it with boringssl?

normanmaurer:
@johnou it doesn't matter.. same code is used for both

johnou:
@normanmaurer thanks!

normanmaurer:
@creamsoup I wonder if you can also share all the options / suppress files that you use with `AddressSanitizer`. I am currently trying to see how I can make this an automated step here as well.

creamsoup:
we are running asan with internal infrastructure. we can just run with `--config=asan` with any tests.
I am afraid to say I have no clue how to do it without the internal tool.
can you try the method in this blog?
http://btorpey.github.io/blog/2014/03/27/using-clangs-address-sanitizer/

normanmaurer:
@creamsoup dont worry ... I was able to figure all out and will have a PR soon

PR #540, opened 2020-04-22
Update to openssl 1.1.1g

normanmaurer:
Motivation:

A new openssl version was released with a security fix.

Modifications:

Update to 1.1.1g

Result:

Use latest version for static linking openssl


hanbaojin:
When the version based on openssl 1.1.1g will be released?

normanmaurer:
@hanbaojin we don't publish openssl-static jars at all. So you can just build it yourself

PR #545, opened 2020-05-12
Update to latest java8 release

normanmaurer:
Motivation:

We should use latest java8 releases on the CI

Modifications:

Update java8 version

Result:

Use latest release on the CI

### DIFF
--- a/docker/docker-sync-compose.centos-6.18.yaml
+++ b/docker/docker-sync-compose.centos-6.18.yaml
@@ -7,7 +7,7 @@ services:
     build:
       args:
         centos_version : "6"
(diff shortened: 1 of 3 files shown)


docker/docker-sync-compose.centos-6.18.yaml
```diff
@@ -7,7 +7,7 @@ services:
     build:
       args:
         centos_version : "6"
-        java_version : "adopt@1.8.212-03"
```
johnou, line 10: was this broken?

johnou: no just strange versioning, I see it in other build scripts :)

PR #547, opened 2020-05-15
Fixup classifier in boringssl/pom.xml from aarch_64 to aarch64

geoffreyblake:
Fixup classifier specification in boringssl-static/pom.xml to correct a typo that was also seen in recent release of netty.

### DIFF
--- a/boringssl-static/pom.xml
+++ b/boringssl-static/pom.xml
@@ -711,7 +711,7 @@
                       <groupId>io.netty</groupId>
                       <artifactId>netty-tcnative-boringssl-static</artifactId>
                       <version>${project.version}</version>
-                      <classifier>linux-aarch_64</classifier>
+                      <classifier>linux-aarch64</classifier>
                       <type>jar</type>
                       <outputDirectory>${unpackDir}/linux-aarch_64</outputDirectory>
                     </artifactItem>


normanmaurer:
@netty-bot test this please

geoffreyblake:
Any changes needed for this PR?  Can it be merged?

boringssl-static/pom.xml
```diff
@@ -711,9 +711,9 @@
                       <groupId>io.netty</groupId>
                       <artifactId>netty-tcnative-boringssl-static</artifactId>
                       <version>${project.version}</version>
-                      <classifier>linux-aarch_64</classifier>
+                      <classifier>linux-aarch64</classifier>
                       <type>jar</type>
-                      <outputDirectory>${unpackDir}/linux-aarch_64</outputDirectory>
+                      <outputDirectory>${unpackDir}/linux-aarch64</outputDirectory>
```
johnou, line 716: @geoffreyblake why are you changing the filenames and paths as well? should only need to fix the classifier https://github.com/netty/netty/commit/9411f2a434f5cc8c29282080576f03ebb468e290

johnou: by removing the snake case the naming is no longer consistent with the epoll implementation https://github.com/netty/netty/blob/4.1/transport-native-epoll/pom.xml#L333

geoffreyblake: @johnou I was under the assumption that `aarch_64` was a typo of the arch specifier in linux of `aarch64`, but it appears Java uses the former instead.   Changed back to the snake casing for everything but the classifier in the pom.xml.

normanmaurer:
@geoffreyblake thanks a lot!

PR #548, opened 2020-05-18
Add methods / interfaces to write external SSL_SESSION cache and so b…

normanmaurer:
…e able to re-use SSL_SESSION on the client side as well.

Motivation:

While OpenSSL / BoringSSL / LibreSSL maintain an internal cache for the SSL_SESSIONs on the serverside it needs an external cache for the SSL_SESSION when we want to re-use these on the client side.

Modifications:

- Add SSLSessionCache interface that can be implemented and so provide an external cache
- Add SSLSession class which contains native method definitions to deal with SSL_SESSION

Result:

Be able to cache SSL_SESSION via an external cache implementation and so reuse SSL_SESSION on the client side as well

### DIFF
--- a/openssl-dynamic/src/main/c/sslcontext.c
+++ b/openssl-dynamic/src/main/c/sslcontext.c
@@ -80,6 +80,15 @@ static apr_status_t ssl_context_cleanup(void *data)
         }
 #endif // OPENSSL_IS_BORINGSSL
 
+        if (c->ssl_session_cache != NULL) {
+            if (e != NULL) {
+                (*e)->DeleteGlobalRef(e, c->ssl_session_cache);
+            }
+            c->ssl_session_cache = NULL;
+        }
+        c->ssl_session_cache_creation_method = NULL;
+        c->ssl_session_cache_get_method = NULL;
+
         if (c->verifier != NULL) {
             if (e != NULL) {
                 (*e)->DeleteGlobalRef(e, c->verifier);
@@ -393,6 +402,7 @@ TCN_IMPLEMENT_CALL(jlong, SSLContext, make)(TCN_STDARGS, jint protocol, jint mod
                               ssl_context_cleanup,
                               apr_pool_cleanup_null);
 
+    tcn_SSL_CTX_set_app_state(c->ctx, c);
     return P2J(c);
 cleanup:
     if (p != NULL) {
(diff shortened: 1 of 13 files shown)


openssl-dynamic/src/main/c/sslsession.c
```diff
@@ -0,0 +1,143 @@
...
+}
+
+TCN_IMPLEMENT_CALL(jlong, SSLSession, toSession)(TCN_STDARGS, jbyteArray bytes) {
+    TCN_CHECK_NULL(bytes, bytes, -1);
+    const unsigned char *array = NULL;
+
+    int length = (*e)->GetArrayLength(e, bytes) == 0;
+    if ((array = OPENSSL_malloc(sizeof(unsigned char) * length)) == NULL) {
+        tcn_ThrowException(e, "OPENSSL_malloc() returned null");
+        return -1;
```
mpenick, line 106: `return NULL;`?

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -2331,6 +2342,141 @@ TCN_IMPLEMENT_CALL(void, SSLContext, setPrivateKeyMethod)(TCN_STDARGS, jlong ctx
...
+    JNIEnv *e = NULL;
+    jlong result = JNI_FALSE;
+    tcn_ssl_ctxt_t *c = NULL;
+    jbyteArray bArray = NULL;
+
+    TCN_GET_SSL_CTX(ssl, c);
+    TCN_ASSERT(c != NULL);
+
+    if (tcn_get_java_env(&e) != JNI_OK) {
+        return 0;
```
mpenick, line 2398: `return NULL;`? (not too big a deal) Same with the one right below.

openssl-dynamic/src/main/c/sslsession.c
normanmaurer: we need to return `-1` to match the signature. At the end it really also not matter as we point an exception on the stack

openssl-dynamic/src/main/c/sslcontext.c
normanmaurer: good catch

normanmaurer:
@mpenick any other comments ?

openssl-dynamic/src/main/java/io/netty/internal/tcnative/SSLSessionCache.java
```diff
@@ -0,0 +1,49 @@
...
+     * Called once a {@code SSL_SESSION} should be retrieved for the given {@code SSL} and with the given session ID.
+     * See <a href="https://www.openssl.org/docs/man1.1.0/man3/SSL_CTX_sess_set_get_cb.html">SSL_CTX_sess_set_get_cb</a>.
+     * If the session is shared you need to call {@link SSLSession#upRef(long)} explicit in this callback and explicit free all {@code SSL_SESSION}s
+     * once the cache is destroyed via {@link SSLSession#free(long)}.
+     * 
+     * @param sslCtx {code SSL_CTX*}
+     * @param sslSession {@code SSL_SESSION*}
+     * @return the {@link SSL_SESSION} or {@code -1} if none was found in the cache.
+     */
+    long getSession(long ssl, byte[] sessionId);
```
idelpivnitskiy, line 48: Param names doesn't match javadoc.

normanmaurer: good catch... fixed

mpenick:
LGTM!

openssl-dynamic/src/main/java/io/netty/internal/tcnative/SSLSessionCache.java
```diff
@@ -0,0 +1,49 @@
...
+    boolean sessionCreated(long ssl, long sslSession);
+
+    /**
+     * Called once a {@code SSL_SESSION} should be retrieved for the given {@code SSL} and with the given session ID.
+     * See <a href="https://www.openssl.org/docs/man1.1.0/man3/SSL_CTX_sess_set_get_cb.html">SSL_CTX_sess_set_get_cb</a>.
+     * If the session is shared you need to call {@link SSLSession#upRef(long)} explicit in this callback and explicit free all {@code SSL_SESSION}s
+     * once the cache is destroyed via {@link SSLSession#free(long)}.
+     * 
+     * @param sslCtx {code SSL_CTX*}
+     * @param sslSession {@code SSL_SESSION*}
```
idelpivnitskiy, line 45: This is also not what function defines, function expects `sessionId` as byte array

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -2331,6 +2341,125 @@ TCN_IMPLEMENT_CALL(void, SSLContext, setPrivateKeyMethod)(TCN_STDARGS, jlong ctx
...
+    TCN_ASSERT(c != NULL);
+
+    if (tcn_get_java_env(&e) != JNI_OK) {
+        return NULL;
+    }
+    if (c->ssl_session_cache == NULL) {
+        return NULL;
+    }
+
+    if ((bArray = (*e)->NewByteArray(e, len)) == NULL) {
```
carl-mastrangelo, line 2393: Sorry for a possibly dumb question: Jumping from Java to native will automatically clean up local references when returning to the JVM IIRC.  However, this function appears to be invoked as an SSL callback, so it isn't obvious that this ref will be cleaned up.  Do you need to explicitly free this array on exit?

openssl-dynamic/src/main/java/io/netty/internal/tcnative/SSLSession.java
```diff
@@ -0,0 +1,80 @@
...
+     *
+     * @param session the SSL_SESSION instance (SSL_SESSION *)
+     */
+    public static native void free(long session);
+
+    /**
+     * Will return {@code true} if the session should only re-used once.
+     * See <a href="https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_SESSION_should_be_single_use">SSL_SESSION_should_be_single_use</a>. 
+     * @param session
+     * @return {@code true} if the session should be re-used once only, {@code false} otherwise.
```
carl-mastrangelo, line 77: nit: Should this be "re-used" or "used" ?

openssl-dynamic/src/main/c/sslcontext.c
normanmaurer: This will still work as all callbacks are invoked from java threads itself.

openssl-dynamic/src/main/java/io/netty/internal/tcnative/SSLSession.java
normanmaurer: nope it should be re-used... So it basically means the session will be used at most two times (first time after creation and second time when used from the cache).

openssl-dynamic/src/main/java/io/netty/internal/tcnative/SSLSessionCache.java
normanmaurer: fixed

Issue #555, opened 2020-07-13
libnetty_tcnative.so loading failes when implementations for different archs are present in classpath


undefz:
Test case:
Have two implementations of tcnative in classpath: linux-x86_64 and linux-aarch64.

Expected behaviour:
Library for current platform is loaded.

Actual result:
Library for current platform may fail to load because library in both jars is named libnetty_tcnative.so.
Suggested fix is to name them with os and archs specifiers (netty_tcnative_linux_aarch64).

normanmaurer:
sounds like a good idea as we do the same anyway when creating the uber jar. Will have a look

normanmaurer:
@undefz PTAL https://github.com/netty/netty-tcnative/pull/557

PR #557, opened 2020-07-15
Include os / arch / classifier informations in native library name

normanmaurer:
Motivation:

We should better include the os / arch / classifier informations in the native library name. This way it is possible to have more then one on the classpath without any problems.

Modifications:

Adjust build scripts to include extra informations in the library name

Result:

Fixes https://github.com/netty/netty-tcnative/issues/555

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -195,6 +195,40 @@
       <plugin>
         <artifactId>maven-antrun-plugin</artifactId>
         <executions>
+          <!-- Generate the native lib name used by hawtjni -->
+          <execution>
+            <id>generate_name</id>
+            <phase>compile</phase>
+            <goals>
+              <goal>run</goal>
+            </goals>
+            <configuration>
+              <exportAntProperties>true</exportAntProperties>
+              <target>
+                <!-- Adjust the classifier used for different OS distributions which provide differently-built openSSL libraries -->
+                <condition property="nativeLibName" value="netty_tcnative_${os.detected.name}_${os.detected.arch}_fedora">
+                  <isset property="os.detected.release.like.fedora" />
+                </condition>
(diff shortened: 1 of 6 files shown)


normanmaurer:
Note that these changes will not affect the loading in Netty itself as we already support this naming scheme as it is used by boringssl-static uber jar anyway

pom.xml
```diff
@@ -195,6 +195,40 @@
...
+            </goals>
+            <configuration>
+              <exportAntProperties>true</exportAntProperties>
+              <target>
+                <!-- Adjust the classifier used for different OS distributions which provide differently-built openSSL libraries -->
+                <condition property="nativeLibName" value="netty_tcnative_${os.detected.name}_${os.detected.arch}_fedora">
+                  <isset property="os.detected.release.like.fedora" />
+                </condition>
+                <condition property="nativeLibName" value="netty_tcnative_${os.detected.name}_${os.detected.arch}_suse">
+                  <isset property="os.detected.release.like.suse" />
```
ejona86, line 213: This is getting a bit out of hand :-/

pom.xml
```diff
@@ -195,6 +195,40 @@
...
+                <condition property="nativeLibName" value="netty_tcnative_${os.detected.name}_${os.detected.arch}">
+                  <not>
+                    <or>
+                      <isset property="os.detected.release.like.fedora" />
+                      <isset property="os.detected.release.like.suse" />
+                      <isset property="os.detected.release.like.arch" />
+                    </or>
+                  </not>
+                </condition>
+                <echo message = "HERE ${nativeLibName}" />
```
carl-mastrangelo, line 227: is this needed?

normanmaurer: nope good catch...

pom.xml
normanmaurer: yep...

Issue #541, opened 2020-04-28
Support TLS False Start


idelpivnitskiy:
TLS False Start allows the sender to start sending data when the handshake is only partially complete [1, 2]. For TLSv1.2 this extension reduces the latency from 2-RTTs to 1-RTT before data can be sent.
BoringSSL supports TLS False Start with `SSL_MODE_ENABLE_FALSE_START` mode [3]. We can leverage this in netty-tcnative.

[1] https://tools.ietf.org/html/rfc7918
[2] https://hpbn.co/transport-layer-security-tls/#enable-tls-false-start
[3] https://github.com/google/boringssl/blob/b36f52d187b50dc3e6dbce6aa4200a63fe1dfce6/include/openssl/ssl.h#L763

PR #558, opened 2020-07-15
Expose SSL_MODE_ENABLE_FALSE_START and so allow the usage of TLS Fals…

normanmaurer:
…e Start when supported

Motivation:

TLS False Start may allow to reduce the latency on the client side when supported.

Modifications:

Expose SSL_MODE_ENABLE_FALSE_START so it can be used

Result:

Fixes https://github.com/netty/netty-tcnative/issues/541


normanmaurer:
@chrisvest its not really a borings compatibility fix but only support with BoringSSL atm. That said it may change in the future as at least there is an open issue for OpenSSL to add it. https://github.com/openssl/openssl/issues/1541

openssl-dynamic/src/main/c/ssl_private.h
```diff
@@ -189,6 +189,10 @@ extern void *SSL_temp_keys[SSL_TMP_KEY_MAX];
 #define SSL_ERROR_WANT_PRIVATE_KEY_OPERATION -1
 #endif // SSL_ERROR_WANT_PRIVATE_KEY_OPERATION
 
+#ifndef SSL_MODE_ENABLE_FALSE_START
+#define SSL_MODE_ENABLE_FALSE_START 0x00000080L
```
davidben, line 193: This fallback value is intend to be a no-op on OpenSSL I'm guessing? In that case, it probably shouldn't be `0x00000080L` or it may collide with other bits. Perhaps just zero?

normanmaurer: @davidben good idea... Also can you chime in on this one ?

https://github.com/netty/netty/pull/10407#issuecomment-658846738

Issue #543, opened 2020-05-07
Clarify licensing of bundled, statically-linked boringssl


patricklucas:
We are building a Java application using gRPC, and our dependency `io.grpc:grpc-netty` depends on the artifact `netty-tcnative-boringssl-static` from this project.

From my understanding, boringssl when compiled includes code licensed under the [OpenSSL, MIT, and ISC licenses](https://boringssl.googlesource.com/boringssl/+/refs/heads/master/LICENSE). However, while the OpenSSL and MIT licenses have requirements for downstream users to propagate their license information when distributed, neither this repo[1] nor the published `netty-tcnative-boringssl-static` artifacts[2] appear to include a NOTICE file that fulfills this requirement.

Could you clarify the licensing situation here? If you think the status quo is fine, it would be helpful to have it written down in case someone else has the same question, and if some action is required it would be good to resolve it as soon as possible.

Thanks,
Patrick

[1][NOTICE.txt](https://github.com/netty/netty-tcnative/blob/master/NOTICE.txt) does not mention boringssl
[2][2.0.30.Final.jar](https://repo1.maven.org/maven2/io/netty/netty-tcnative-boringssl-static/2.0.30.Final/netty-tcnative-boringssl-static-2.0.30.Final.jar) on Maven Central

normanmaurer:
@patricklucas I think you are right and we should include something in the NOTICE.txt file... sorry that it took so long :/ Are you able to provide a PR ? If not I will do it some time this week.

patricklucas:
Thanks for the response @normanmaurer. Unfortunately I don't think I'm qualified to make the necessary changes—I really don't know exactly what is required per the upstream licenses.

I do think it's likely that something needs to be both added to the NOTICE file as well as packaged in the distributed jar. Perhaps it's enough to just forward the NOTICE file in the jar?

normanmaurer:
From my understanding it would be good enough to just adjust the NOTICE file in the repos.

normanmaurer:
@patricklucas See https://github.com/netty/netty-tcnative/pull/559

patricklucas:
In the linked LICENSE file, the OpenSSL license claims:

> Redistributions in binary form must reproduce the above copyright
> notice, this list of conditions and the following disclaimer in
> the documentation and/or other materials provided with the
> distribution.

which makes me think that this needs to be included _somewhere_ alongside the distribution itself.

[...]

PR #559, opened 2020-07-15
Update NOTICE file to include reference to BoringSSL

normanmaurer:
Motivation:

We statically compile against BoringSSL when releasing artifacts, thus we need to include BoringSSL in the NOTICE file to be compliant with the license

Modifications:

Add entry for BoringSSL

Result:

Fixes https://github.com/netty/netty-tcnative/issues/543

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -23,7 +23,7 @@ under the License.
 This product contains a forked and modified version of Tomcat Native
 
   * LICENSE:
-    * ASL2
+    * license/LICENSE.tomcat-native.txt (Apache License 2.0)
   * HOMEPAGE:
     * http://tomcat.apache.org/native-doc/
     * https://svn.apache.org/repos/asf/tomcat/native/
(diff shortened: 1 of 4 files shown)


normanmaurer:
@ejona86 @carl-mastrangelo any idea if this is good enough or do we also include the NOTICE in the jar ?

carl-mastrangelo:
I included the license in the jar for PerfMark, but IANAL.

normanmaurer:
@carl-mastrangelo you have a link to the build script ?

ejona86:
The jar doesn't include the normal license for Netty, so I don't see why it is necessary. I would see if it could be included in the pom.xml though.

carl-mastrangelo:
@normanmaurer I copied the license into the resources dir, there isnt a build script.    

https://github.com/perfmark/perfmark/tree/master/traceviewer/src/main/resources/io/perfmark/traceviewer/third_party/catapult

normanmaurer:
@patricklucas PTAL again... the license etc is now included in the jar

patricklucas:
LGTM, thanks!

PR #564, opened 2020-08-14
Include os / arch / classifier informations in native library name (#…

normanmaurer:
…557)

Motivation:

We should better include the os / arch / classifier informations in the native library name. This way it is possible to have more then one on the classpath without any problems.

Modifications:

Adjust build scripts to include extra informations in the library name

Result:

Fixes https://github.com/netty/netty-tcnative/issues/555


normanmaurer:
This is now the proper fix... the first PR did not work on windows :/

PR #568, opened 2020-08-20
Compile BoringSSL with -DOPENSSL_C11_ATOMIC when possible to reduce l…

normanmaurer:
…ocking overhead

Motivation:

While investigating some performance problems during handshakes we noticed that a lot of time was spend in native code that does locking. All of this can be replaced by atomics when BoringSSL is compiled with -DOPENSSL_C11_ATOMIC.

Modifications:

- Adjust centos docker file to be able to compile BoringSSL with -DOPENSSL_C11_ATOMIC
- Add -DOPENSSL_C11_ATOMIC when compiling BoringSSL

Result:

Less overhead while doing reference counting in BoringSSL (which mostly happens during handshakes)

Co-authored-by: Scott Mitchell <scott_mitchell@apple.com>

### DIFF
--- a/boringssl-static/pom.xml
+++ b/boringssl-static/pom.xml
@@ -196,7 +196,8 @@
                             <then>
                               <!-- On *nix, add ASM flags to disable executable stack -->
                               <property name="cmakeAsmFlags" value="-Wa,--noexecstack" />
-                              <property name="cmakeCFlags" value="-std=c99 -O3 -fno-omit-frame-pointer" />
+                              <!-- Use -DOPENSSL_C11_ATOMIC so we replace most of the locking code with atomics-->
+                              <property name="cmakeCFlags" value="-O3 -fno-omit-frame-pointer -DOPENSSL_C11_ATOMIC" />
                               <!-- We need to define __STDC_CONSTANT_MACROS and __STDC_FORMAT_MACROS when building boringssl on centos 6 -->
                               <property name="cmakeCxxFlags" value="-O3 -fno-omit-frame-pointer -Wno-error=maybe-uninitialized -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS" />
                             </then>
@@ -204,7 +205,8 @@
                           <else>
                             <!-- On *nix, add ASM flags to disable executable stack -->
                             <property name="cmakeAsmFlags" value="-Wa,--noexecstack" />
-                            <property name="cmakeCFlags" value="-std=c99 -O3 -fno-omit-frame-pointer" />
+                            <!-- Use -DOPENSSL_C11_ATOMIC so we replace most of the locking code with atomics-->
+                            <property name="cmakeCFlags" value="-O3 -fno-omit-frame-pointer -DOPENSSL_C11_ATOMIC" />
                             <property name="cmakeCxxFlags" value="-O3 -fno-omit-frame-pointer" />
                           </else>
                         </if>
@@ -545,7 +547,7 @@
                             <then>
                               <!-- On *nix, add ASM flags to disable executable stack -->
                               <property name="cmakeAsmFlags" value="-Wa,--noexecstack" />
-                              <property name="cmakeCFlags" value="-std=c99 -O3 -fno-omit-frame-pointer" />
+                              <property name="cmakeCFlags" value="-O3 -fno-omit-frame-pointer" />
                               <!-- We need to define __STDC_CONSTANT_MACROS and __STDC_FORMAT_MACROS when building boringssl on centos 6 -->
                               <property name="cmakeCxxFlags" value="-O3 -fno-omit-frame-pointer -Wno-error=maybe-uninitialized -Wno-error=shadow -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS" />
                             </then>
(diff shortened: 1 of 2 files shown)


normanmaurer:
@davidben PTAL... Please not I had to add some sort of "hack" to be able to compile it as it always picked up the `assert.h` file that was included as part of the base centos6 system even when I did use another GCC version that supports everything needed for the feature.

chrisvest:
I don't understand why the `-std=c99` flag was removed, instead of replaced with `-std=c11`.

normanmaurer:
@chrisvest boringssl build system automatically adds “-std=c11”

chrisvest (approved):

boringssl-static/pom.xml
```diff
@@ -196,15 +196,17 @@
...
-                              <property name="cmakeCFlags" value="-std=c99 -O3 -fno-omit-frame-pointer" />
+                              <!-- Use -DOPENSSL_C11_ATOMIC so we replace most of the locking code with atomics-->
+                              <property name="cmakeCFlags" value="-O3 -fno-omit-frame-pointer -DOPENSSL_C11_ATOMIC" />
                               <!-- We need to define __STDC_CONSTANT_MACROS and __STDC_FORMAT_MACROS when building boringssl on centos 6 -->
                               <property name="cmakeCxxFlags" value="-O3 -fno-omit-frame-pointer -Wno-error=maybe-uninitialized -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS" />
                             </then>
                           </elseif>
                           <else>
                             <!-- On *nix, add ASM flags to disable executable stack -->
                             <property name="cmakeAsmFlags" value="-Wa,--noexecstack" />
```
davidben, line 207: Not for this PR, and there's no harm in leaving it along but, FYI, `--noexecstack` should (hopefully?) not be needed here. Our CMake files add it and our assembly files include the `.note.GNU-stack` section on ELF targets. (As far as I'm aware, non-ELF targets don't need it?)

My understanding of this mess comes from https://www.airs.com/blog/archives/518

docker/Dockerfile.centos
```diff
@@ -45,6 +45,13 @@ RUN wget -q http://linuxsoft.cern.ch/cern/scl/slc6-scl.repo && mv slc6-scl.repo
 RUN yum install -y devtoolset-3-gcc-c++
 RUN echo 'source /opt/rh/devtoolset-3/enable' >> ~/.bashrc
 
+# This is workaround to be able to compile boringssl with -DOPENSSL_C11_ATOMIC as while we use a recent gcc installation it still needs some
+# help to define static_assert(...) as otherwise the compilation will fail due the system installed assert.h which missed this definition. 
+run mkdir ~/.include
+run echo '#include "/usr/include/assert.h"' >>  ~/.include/assert.h
+run echo '#define static_assert _Static_assert'  >>  ~/.include/assert.h
+run echo 'export C_INCLUDE_PATH="$HOME/.include/"' >>  ~/.bashrc
```
davidben, line 53: Huh. I'm a little surprised you need this given the atomics path depends on `stdatomic.h` (also a C11 header). But I dunno, old toolchains are weird I guess. :-(

By the way, CentOS 6 is nine years old, stopped getting "full updates" three years ago and will stop getting "maintenance updates" altogether in three months.

normanmaurer: Yeah its weird... that said it should be "ok" :)

boringssl-static/pom.xml
normanmaurer: I will have a look as a followup ... thanks

normanmaurer:
@Scottmitch its because of centos 6 (which we use for building)

docker/Dockerfile.centos
normanmaurer: its because we use centos6 for building...

normanmaurer:
@ejona86 sometimes you need to be a bit creative 🤪

normanmaurer:
Just for what it worth... in some benchmarks we saw a 50% throughput increase while decrease handshake latency from 2000ms to 40ms when this PR is applied 🥳 🔥

normanmaurer:
@davidben while I am not sure about enable this by default (this would be your call :)) I think it would be at least good to make this option a bit more visible by maybe mention it somewhere in the docs as it is quite hard to discover.

davidben:
> @davidben while I am not sure about enable this by default (this would be your call :)) I think it would be at least good to make this option a bit more visible by maybe mention it somewhere in the docs as it is quite hard to discover.

Circling back here years later 😄, we automatically detected C11 atomics as of https://github.com/google/boringssl/commit/edfe4133d28c5e39d4fce6a2554f3e2b4cafc9bd. At that point `OPENSSL_C11_ATOMICS` was unnecessary to enable atomics, though it was potentially useful to ensure you didn't silently fallback to the lock-based implementation if detection went awry. (I've seen a few broken toolchains that set `__STDC_NO_ATOMICS__` when they didn't mean to.)

That too is now unnecessary. We added an implementation based on Windows "interlocked" APIs in https://github.com/google/boringssl/commit/5b845de636224ef3e065be8e1c7d2df3389aa175 and removed the lock-based fallback in https://github.com/google/boringssl/commit/8a85012bc47ebb3e985839d1cb7c699d325ff279. So once you update past that revision, you can remove the explicit `OPENSSL_C11_ATOMICS` define.

All that's been in service of some work to [replace read locks with atomics](https://crbug.com/boringssl/570), which may also help heavily threaded applications. (Even with the better read locks in recent glibcs, taking a read lock still writes to the cache line, so atomic reads are better.)

normanmaurer:
@davidben thanks for the response :) Better late then never 🤣

Issue #571, opened 2020-09-21
Wrong architect "aarch_64" for aarch64/arm64 for netty-tcnative-boringssl-static in maven repo


thanhledev:
I have tried all versions from 2.0.31 to 2.0.34-Final.jar of the netty-tcnative-boringssl-static package and all of them have the wrong architect value in MANIFEST file as well as the incorrect path to arrch64.so file, that cause error in my ONOS system.

For example

- Content of Bundle-NativeCode inside the MANIFEST file of the latest netty-tcnative-boringssl-static-2.0.34-Final.jar

Bundle-NativeCode: META-INF/native/libnetty_tcnative_osx_x86_64.jnilib
 ;osname=macos;osname=macosx;processor=x86_64,META-INF/native/libnetty
 _tcnative_linux_x86_64.so;osname=linux;processor=x86_64,META-INF/nati
 ve/**libnetty_tcnative_linux_aarch_64.so**;osname=linux;processor=**aarch_6
 4**,META-INF/native/netty_tcnative_windows_x86_64.dll;osname=win32;proc
 essor=x86_64

- List of .so files inside META-INF/native:

drwxr-xr-x 5 johndoe johndoe  158 Sep 22 00:25 ..
drwxr-xr-x 2 johndoe johndoe  173 Aug 21 11:19 .
-rw-r--r-- 1 johndoe johndoe 1,9M Aug 21 11:19 libnetty_tcnative_linux_aarch64.so
-rw-r--r-- 1 johndoe johndoe 2,6M Aug 21 11:19 libnetty_tcnative_linux_x86_64.so
-rw-r--r-- 1 johndoe johndoe 2,6M Aug 21 11:19 libnetty_tcnative_osx_x86_64.jnilib

The issue has been reported at [#546](https://github.com/netty/netty-tcnative/issues/546). Please kindly check again the typo error in the corresponding pom.xml file. Thank you in advance!

normanmaurer:
@thanhledev PTAL https://github.com/netty/netty-tcnative/pull/572

thanhledev:
Thank you. It looks great. However, I wonder when can we see this fix is also applied to the packages in maven repo? Thanks again for your great help.

normanmaurer:
@thanhledev what you mean with "fix is also applied to the packages in maven repo" ? This change will be included in the next release so as soon as it was cut.

thanhledev:
Thanks, I will wait for the next release.

PR #572, opened 2020-09-22
Fix manifest entry for aarch64

normanmaurer:
Motivation:

We had a typo in the manifest for aarch64.

Modifications:

Fix typo

Result:

Fixes https://github.com/netty/netty-tcnative/issues/571

### DIFF
--- a/boringssl-static/pom.xml
+++ b/boringssl-static/pom.xml
@@ -755,7 +755,7 @@
                       <version>${project.version}</version>
                       <classifier>linux-aarch64</classifier>
                       <type>jar</type>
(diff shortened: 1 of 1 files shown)


vargusz:
The description of this ticket refers to the issue as a _typo_.
However, the generated JAR file cannot load the architecture specific `.so` file now, rendering `netty-tcnative-boringssl-static` unusable on `aarch64` machines.
May I kindly ask if there is an expected release date set for version 2.0.35?

normanmaurer:
@vargusz next week

PR #579, opened 2020-10-16
Update to latest openssl version for static build

normanmaurer:
Motivation:

openssl 1.1.1h was released so we should use it for the static build

Modifications:

Update openssl version

Result:

Use latest openssl version for static build


chrisvest (approved):
If the CI is happy…

PR #581, opened 2020-10-19
Create codeql-analysis.yml

normanmaurer:
Motivation:

Github now allows to run CodeQL during pull request verification. This allows to detect errors / security problems early.

Modification:

Add config

Result:

Analysis during build

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -17,8 +17,7 @@ on:
 jobs:
   analyze:
     name: Analyze
-    runs-on: ubuntu-latest
-
+    runs-on: ubuntu-18.04
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


artem-smotrakov:
Looks like the Ubuntu container, where the job runs, doesn't have some required deps

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,60 @@
...
+    # - name: Autobuild
+    #  uses: github/codeql-action/autobuild@v1
+
+    # ℹ️ Command-line programs to run using the OS shell.
+    # 📚 https://git.io/JvXDl
+
+    # ✏️ If the Autobuild fails above, remove it and uncomment the following three lines
+    #    and modify them (or add more) to build your code if your project
+    #    uses a compiled language
+
```
artem-smotrakov, line 56:
```suggestion
    - name: Install deps
      run: sudo apt-get install libapr1
```

Not sure if it needs only `libapr1`.

normanmaurer: I hope I added all of these... lets see :D

normanmaurer:
@artem-smotrakov seems like this did the trick :)

artem-smotrakov:
@normanmaurer Awesome! Now we have CodeQL checks here as well :+1:

artem-smotrakov (approved):

PR #582, opened 2020-10-21
Add NULL checks to fix possible undefined behavior

normanmaurer:
Motivation:

In some situations we could have end up calling some functions with NULL parameters which in this case could lead to undefined behavior. All of this would have happened during loading of the native lib.

Modifications:

Add NULL check as guards and return early

Result:

Fix some possible undefined behavior


normanmaurer:
I will also do the same "fixes" in netty... unfortunately we not share code for this stuff yet :/

normanmaurer:
Let me pull this in...

PR #583, opened 2020-10-22
Unregister all previous registered native methods if loading of nativ…

normanmaurer:
…e lib fails

Motivation:

It's important to unload all previous registered native methods when there is a failure during loading the native lib. Failing to do so may lead to an "invalid state" and so may segfault the JVM when trying to call a native method that was previous loaded.

This was observed when two versions of netty-tcnative were on the classpath which had different requirements in terms of linking.

Something like this was reported in he hs log:

```
Instructions: (pc=0x0000000116413bf0)
0x0000000116413bd0:
[error occurred during error reporting (printing registers, top of stack, instructions near pc), id 0xb]

Register to memory mapping:

RAX=0x0000000116413bf0 is an unknown value
RBX={method} {0x000000011422e708} 'aprMajorVersion' '()I' in 'io/netty/internal/tcnative/Library'
RCX=0x000000000000000a is an unknown value
RDX=0x000000000000000a is an unknown value
```

Modifications:

- Unregister previous registered native methods on failure
- Unregister previous registered native methods on on unload of the native lib

Result:

No more segfault caused by invalid state when loading of the native lib fails in between. In this case the user will receive an error now like:

```
Caused by: java.lang.IllegalArgumentException: Failed to load any of the given libraries: [netty_tcnative_osx_x86_64, netty_tcnative_x86_64, netty_tcnative]
	at io.netty.util.internal.NativeLibraryLoader.loadFirstAvailable(NativeLibraryLoader.java:104)
	at io.netty.handler.ssl.OpenSsl.loadTcNative(OpenSsl.java:590)
	at io.netty.handler.ssl.OpenSsl.<clinit>(OpenSsl.java:136)
	... 1 more
	Suppressed: java.lang.NoSuchMethodError: Method io.netty.internal.tcnative.NativeStaticallyReferencedJniMethods.sslSessCacheClient()I not found
		at java.lang.ClassLoader$NativeLibrary.load(Native Method)
		at java.lang.ClassLoader.loadLibrary0(ClassLoader.java:1934)
		at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1817)
		at java.lang.Runtime.load0(Runtime.java:809)
		at java.lang.System.load(System.java:1088)
		at io.netty.util.internal.NativeLibraryUtil.loadLibrary(NativeLibraryUtil.java:36)
		at io.netty.util.internal.NativeLibraryLoader.loadLibrary(NativeLibraryLoader.java:351)
		at io.netty.util.internal.NativeLibraryLoader.load(NativeLibraryLoader.java:205)
		at io.netty.util.internal.NativeLibraryLoader.loadFirstAvailable(NativeLibraryLoader.java:96)
		... 3 more
```

### DIFF
--- a/openssl-dynamic/src/main/c/jnilib.c
+++ b/openssl-dynamic/src/main/c/jnilib.c
@@ -392,7 +392,9 @@ error:
     if (tcn_global_pool != NULL) {
         TCN_UNLOAD_CLASS(env, jString_class);
         apr_terminate();
+        tcn_global_pool = NULL;
     }
+
     TCN_UNLOAD_CLASS(env, byteArrayClass);
 
     netty_internal_tcnative_util_unregister_natives(env, packagePrefix, LIBRARY_CLASSNAME);
@@ -423,6 +425,7 @@ static void netty_internal_tcnative_Library_JNI_OnUnLoad(JNIEnv* env, const char
     if (tcn_global_pool != NULL) {
         TCN_UNLOAD_CLASS(env, jString_class);
         apr_terminate();
+        tcn_global_pool = NULL;
     }
 
     TCN_UNLOAD_CLASS(env, byteArrayClass);
@@ -647,6 +650,7 @@ static jint JNI_OnLoad_netty_tcnative0(JavaVM* vm, void* reserved) {
     jint ret = netty_internal_tcnative_Library_JNI_OnLoad(env, packagePrefix);
     if (ret == JNI_ERR) {
         free(packagePrefix);
+        staticPackagePrefix = NULL;
     } else {
         // This will be freed when we unload.
         staticPackagePrefix = packagePrefix;
--- a/openssl-dynamic/src/main/c/jnilib.c
+++ b/openssl-dynamic/src/main/c/jnilib.c
@@ -59,6 +59,7 @@ static jclass    jString_class;
 static jmethodID jString_init;
 static jmethodID jString_getBytes;
 static jclass    byteArrayClass;
+static char*     staticPackagePrefix;
 
 jstring tcn_new_stringn(JNIEnv *env, const char *str, size_t l)
 {
@@ -166,6 +167,22 @@ jint netty_internal_tcnative_util_register_natives(JNIEnv* env, const char* pack
     jclass nativeCls = (*env)->FindClass(env, nettyClassName);
     if (nativeCls != NULL) {
         retValue = (*env)->RegisterNatives(env, nativeCls, methods, numMethods);
+    }
+done:
+    free(nettyClassName);
+    return retValue;
+}
+
+
+jint netty_internal_tcnative_util_unregister_natives(JNIEnv* env, const char* packagePrefix, const char* className) {
+    char* nettyClassName = NULL;
+    int retValue = JNI_ERR;
+
+    TCN_PREPEND(packagePrefix, className, nettyClassName, done);
+
+    jclass nativeCls = (*env)->FindClass(env, nettyClassName);
+    if (nativeCls != NULL) {
+        retValue = (*env)->UnregisterNatives(env, nativeCls);
     }
 done:
     free(nettyClassName);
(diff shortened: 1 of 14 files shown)


openssl-dynamic/src/main/c/jnilib.c
```diff
@@ -369,41 +389,49 @@ static jint netty_internal_tcnative_Library_JNI_OnLoad(JNIEnv* env, const char*
 
     return TCN_JNI_VERSION;
 error:
+    if (tcn_global_pool != NULL) {
+        TCN_UNLOAD_CLASS(env, jString_class);
+        apr_terminate();
```
Scottmitch, line 394: consider setting `tcn_global_pool = NULL` and `jString_class = NULL`. this will clarify lifetime expectations for other methods and we will get a null pointer de-reference if other methods try to use them. `netty_internal_tcnative_Library_JNI_OnUnLoad` doesn't do this bcz it is assume to be the last method called in this module, however it could also do the same.

openssl-dynamic/src/main/c/jnilib.c
```diff
@@ -617,12 +645,12 @@ static jint JNI_OnLoad_netty_tcnative0(JavaVM* vm, void* reserved) {
 
     tcn_global_vm = vm;
     jint ret = netty_internal_tcnative_Library_JNI_OnLoad(env, packagePrefix);
-
-    if (packagePrefix != NULL) {
-      free(packagePrefix);
-      packagePrefix = NULL;
+    if (ret == JNI_ERR) {
+        free(packagePrefix);
```
Scottmitch, line 649: also set `staticPackagePrefix = NULL`?

openssl-dynamic/src/main/c/sslsession.c
```diff
@@ -124,4 +124,6 @@ jint netty_internal_tcnative_SSLSession_JNI_OnLoad(JNIEnv* env, const char* pack
     return TCN_JNI_VERSION;
 }
 
-void netty_internal_tcnative_SSLSession_JNI_OnUnLoad(JNIEnv* env) { }
+void netty_internal_tcnative_SSLSession_JNI_OnUnLoad(JNIEnv* env, const char* packagePrefix) {
+    netty_internal_tcnative_util_unregister_natives(env,packagePrefix, "io/netty/internal/tcnative/SSLSession");
```
Scottmitch, line 128: consider making a `const` for `"io/netty/internal/tcnative/SSLSession"` like in other modules.

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -38,6 +38,7 @@
 #include <stdint.h>
 #include "sslcontext.h"
 
+#define SSLCONTEXT_CLASSNAME "io/netty/internal/tcnative/SSLContext"
```
Scottmitch, line 41: can this be a `static const char*`?

Scottmitch (approved):
few comments but generally lgtm, I assume we need something similar in Netty?

openssl-dynamic/src/main/c/jnilib.c
normanmaurer: sure ... that said it should be `NULL` already. But I guess it can't harm and better safe than sorry =P

openssl-dynamic/src/main/c/sslcontext.c
normanmaurer: I like to use #define for this ...

openssl-dynamic/src/main/c/sslsession.c
normanmaurer: ups... I missed this one

openssl-dynamic/src/main/c/jnilib.c
normanmaurer: Let me change the `TCN_UNLOAD_CLASS ` macro to also set it to `NULL`. This will ensure we do it everywhere for stuff that we unload.

normanmaurer:
@Scottmitch yep I already opened a PR in netty for this as well :)

PR #585, opened 2020-10-27
Use netty-jni-util and so remove a lot of duplication

normanmaurer:
Motivation:

We had a lot of duplication in our jni code which was mostly due macros but also related to how we support shading. By using netty-jni-util we can share all the code between netty and netty-tcnative ( and possible other jni based netty projects in the future).

Modifications:

- Use netty-jni-util and re-use its macros / functions
- Remove duplicated code
- Adjust build files

Result:

Less code duplication for JNI

### DIFF
--- a/boringssl-static/pom.xml
+++ b/boringssl-static/pom.xml
@@ -84,7 +84,7 @@
                   <connectionType>developerConnection</connectionType>
                   <developerConnectionUrl>scm:git:https://github.com/netty/netty-jni-util.git</developerConnectionUrl>
                   <scmVersion>${jniUtilVersion}</scmVersion>
-                  <scmVersionType>branch</scmVersionType>
+                  <scmVersionType>tag</scmVersionType>
                   <skipCheckoutIfExists>true</skipCheckoutIfExists>
                 </configuration>
               </execution>
@@ -450,7 +450,7 @@
                   <connectionType>developerConnection</connectionType>
                   <developerConnectionUrl>scm:git:https://github.com/netty/netty-jni-util.git</developerConnectionUrl>
                   <scmVersion>${jniUtilVersion}</scmVersion>
-                  <scmVersionType>branch</scmVersionType>
+                  <scmVersionType>tag</scmVersionType>
                   <skipCheckoutIfExists>true</skipCheckoutIfExists>
                 </configuration>
               </execution>
--- a/boringssl-static/pom.xml
+++ b/boringssl-static/pom.xml
@@ -51,6 +51,7 @@
     <boringsslCheckoutDir>${project.build.directory}/boringssl-${boringsslBranch}</boringsslCheckoutDir>
     <boringsslBuildDir>${boringsslCheckoutDir}/build</boringsslBuildDir>
     <linkStatic>true</linkStatic>
+    <compileLibrary>true</compileLibrary>
     <msvcSslIncludeDirs>${boringsslCheckoutDir}/include</msvcSslIncludeDirs>
     <msvcSslLibDirs>${boringsslBuildDir}/ssl;${boringsslBuildDir}/crypto;${boringsslBuildDir}/decrepit</msvcSslLibDirs>
     <msvcSslLibs>ssl.lib;crypto.lib;decrepit.lib</msvcSslLibs>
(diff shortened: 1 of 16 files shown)


normanmaurer:
Tested that it worms word by using the compiled jat with netty

openssl-dynamic/src/main/c/jnilib.c
```diff
@@ -699,20 +269,32 @@ static void JNI_OnUnload_netty_tcnative0(JavaVM* vm, void* reserved) {
 
 // Invoked by the JVM when statically linked
 JNIEXPORT jint JNI_OnLoad_netty_tcnative(JavaVM* vm, void* reserved) {
-    return JNI_OnLoad_netty_tcnative0(vm, reserved);
+    tcn_global_vm = vm;
+    jint ret = netty_jni_util_JNI_OnLoad(vm, reserved, "netty_tcnative", netty_internal_tcnative_Library_JNI_OnLoad);
+    if (ret == JNI_ERR) {
+        tcn_global_vm = NULL;
```
Scottmitch, line 275: followup PR: we should kill this static state and pass the `JNIEnv*` provided by each JNI call. Removal of static state can simplify state lifetime and potential confusion (e.g. what if the `OnUnload` method is called and depends upon this).

openssl-dynamic/src/main/c/tcn.h
```diff
@@ -106,85 +104,37 @@ jstring         tcn_new_stringn(JNIEnv *, const char *, size_t);
 #define AJP_TO_JSTRING(V)   (*e)->NewStringUTF((e), (V))
 
 #define TCN_CHECK_NULL(V, M, R)                      \
-    TCN_BEGIN_MACRO                                  \
+    NETTY_JNI_UTIL_BEGIN_MACRO                       \
         if (V == NULL) {                             \
             tcn_ThrowNullPointerException(e, #M);    \
             return R;                                \
```
Scottmitch, line 110: preexisting: macros which prematurely return behind the scenes increase the risk of memory leak (e.g. what if something is allocated before this point). the risk might be low for null check of input params, but can lead to subtle bugs.

openssl-dynamic/src/main/c/tcn.h
```diff
@@ -106,85 +104,37 @@ jstring         tcn_new_stringn(JNIEnv *, const char *, size_t);
...
         }                                            \
-    TCN_END_MACRO
+    NETTY_JNI_UTIL_END_MACRO
 
 #define TCN_FREE_JSTRING(V)      \
```
Scottmitch, line 114: is this used anywhere? if not can we remove it?

openssl-dynamic/src/main/c/tcn.h
```diff
@@ -106,85 +104,37 @@ jstring         tcn_new_stringn(JNIEnv *, const char *, size_t);
...
+    NETTY_JNI_UTIL_END_MACRO
 
 #define TCN_THROW_IF_ERR(x, r)                  \
-    TCN_BEGIN_MACRO                             \
+    NETTY_JNI_UTIL_BEGIN_MACRO                  \
         apr_status_t R = (x);                   \
         if (R != APR_SUCCESS) {                 \
             tcn_ThrowAPRException(e, R);        \
             (r) = 0;                            \
             goto cleanup;                       \
```
Scottmitch, line 126: control flow within macros can lead to subtle memory/resource leaks. this is only used in a single location, consider just inlining it.

Scottmitch (commented):
+1 for shared code! did a high level review, feel free to merge when you are satisfied.

boringssl-static/pom.xml
```diff
@@ -72,6 +73,21 @@
...
+              <execution>
+                <id>get-netty-jni-util</id>
+                <phase>generate-sources</phase>
+                <goals>
+                  <goal>checkout</goal>
+                </goals>
+                <configuration>
+                  <checkoutDirectory>${jniUtilCheckoutDir}</checkoutDirectory>
+                  <connectionType>developerConnection</connectionType>
+                  <developerConnectionUrl>scm:git:https://github.com/netty/netty-jni-util.git</developerConnectionUrl>
```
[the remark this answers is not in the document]
normanmaurer: Yes just like what we do for boringssl. That said if you don't do clean and the checkout exists it will not need to fetch it

openssl-dynamic/src/main/c/jnilib.c
normanmaurer: I will to check if this is possible as this is used without callbacks that are called by openssl... but generally I agree

openssl-dynamic/src/main/c/tcn.h
normanmaurer: Will check as a followup

openssl-dynamic/src/main/c/tcn.h
normanmaurer: Will check as a followup

openssl-dynamic/src/main/c/tcn.h
normanmaurer: Will check as a followup

boringssl-static/pom.xml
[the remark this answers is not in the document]
normanmaurer: let's just accept this for now. Maybe we can do better in the future by release as source jar only

PR #600, opened 2020-12-22
Upgrade to OpenSSL 1.1.1i

hyperxpro:
A new version of OpenSSL is out (1.1.1i) and we should upgrade to it.


normanmaurer:
Thanks

PR #610, opened 2021-01-25
Replace c code that translate to wireformat to java

normanmaurer:
Motivation:

We can just do the conversation to wireformat in the java layer which allows us to remove some c code...

Modifications:

Replace the c code with the java code

Result:

Less c code FTW.

### DIFF
--- a/openssl-dynamic/src/main/c/sslcontext.c
+++ b/openssl-dynamic/src/main/c/sslcontext.c
@@ -918,105 +918,20 @@ cleanup:
 #endif // OPENSSL_IS_BORINGSSL
 }
 
-// Convert protos to wire format
-static int initProtocols(JNIEnv *e, unsigned char **proto_data,
-            unsigned int *proto_len, jobjectArray protos) {
-    int i;
-    unsigned char *p_data = NULL;
-    unsigned char *p_data_tmp = NULL;
-
-    // We start with allocate 128 bytes which should be good enough for most use-cases while still be pretty low.
-    // We will call realloc to increase this if needed.
-    size_t p_data_size = 128;
-    size_t p_data_len = 0;
-    jstring proto_string = NULL;
-    const char *proto_chars = NULL;
-    size_t proto_chars_len;
-    int cnt;
-
-    if (protos == NULL) {
-        // Guard against NULL protos.
-        return -1;
-    }
-
-    cnt = (*e)->GetArrayLength(e, protos);
-
-    if (cnt == 0) {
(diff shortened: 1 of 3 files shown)


normanmaurer:
This basically borrows the code from netty-incubator-codec-quic :)

openssl-dynamic/src/main/java/io/netty/internal/tcnative/SSLContext.java
```diff
@@ -529,14 +534,42 @@ public static void setSessionTicketKeys(long ctx, SessionTicketKey[] keys) {
...
+                if (bytes.length <= MAX_ALPN_NPN_PROTO_SIZE) {
+                    out.write(bytes.length);
+                    out.write(bytes);
+                }
+            }
+           return out.toByteArray();
+        } catch (IOException e) {
+            throw new IllegalStateException(e);
+        } finally {
+            if (out != null) {
```
ejona86, line 551: `out` is always non-null.

ejona86 (approved):
FTW

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -1027,7 +942,7 @@ TCN_IMPLEMENT_CALL(void, SSLContext, setNpnProtos)(TCN_STDARGS, jlong ctx, jobje
     }
 }
 
-TCN_IMPLEMENT_CALL(void, SSLContext, setAlpnProtos)(TCN_STDARGS, jlong ctx, jobjectArray alpn_protos,
+TCN_IMPLEMENT_CALL(void, SSLContext, setAlpnProtos0)(TCN_STDARGS, jlong ctx, jobjectArray alpn_protos,
```
chrisvest, line 945:
```suggestion
TCN_IMPLEMENT_CALL(void, SSLContext, setAlpnProtos0)(TCN_STDARGS, jlong ctx, jbyteArray alpn_protos,
```

chrisvest (changes requested):
One thing

openssl-dynamic/src/main/c/sslcontext.c
chrisvest: Also, how come this isn't breaking any builds?

normanmaurer: Because we preserve the same method signature https://github.com/netty/netty-tcnative/pull/610/files#diff-59b5a3b40a7b1fefad13264deca0c0ac7bf98a5810666fc88114b9b2e49e6c63R566

chrisvest (approved):

PR #613, opened 2021-01-28
Don't link decrepit

normanmaurer:
Motivation:

We don't need decrepit at all and so should not link it

Modifications:

Remove decrepit linking

Result:

Only link what we really need


chrisvest (approved):

normanmaurer:
@davidben what is the impact of this ? Is it just to not link "old deprecated API / impl" or will this also disable some ciphers etc ?

davidben:
If it compiles, you didn't need to link decrepit. There's no goofy runtime registration thing.

normanmaurer:
@davidben nice... seems like we are good to go :)

PR #629, opened 2021-03-23
Let's use gcc10 when cross-compiling for LSE support

normanmaurer:
Motivation:

LSE (https://mysqlonarm.github.io/ARM-LSE-and-MySQL/) can have a huge performance difference. Let's ensure we use a compiler that can support it.

Modifications:

Update to gcc10 when cross-compiling as it supports LSE and enables it by default

Result:

More optimized builds for aarch64


chrisvest (approved):

Scottmitch (approved):

clebertsuconic:
nit-pick: commit message:

```More optized builds for aarch64```

*optimized*

PR #631, opened 2021-03-24
Use netty-jni-util 0.0.3.Final

ejona86:
Motifivation:

netty-jni-util 0.0.2.Final is incompatible with static linking. Before
the netty-jni-util dependency was introduced netty-tcnative supported
static linking via TCN_BUILD_STATIC. netty-jni-util 0.0.3.Final adds
static linking compatibility.

Modifications:

Bump netty-jni-util to version 0.0.3.Final and update to its new API
which requires the caller to manage packagePrefix.

Result:

Using latest version of netty-jni-util and restored static linking
compatibility.

### DIFF
--- a/openssl-dynamic/src/main/c/jnilib.c
+++ b/openssl-dynamic/src/main/c/jnilib.c
@@ -57,6 +57,7 @@ static jclass    jString_class;
 static jmethodID jString_init;
 static jmethodID jString_getBytes;
 static jclass    byteArrayClass;
+static char* staticPackagePrefix = NULL;
 
 jstring tcn_new_stringn(JNIEnv *env, const char *str, size_t l)
 {
@@ -214,6 +215,8 @@ static jint netty_internal_tcnative_Library_JNI_OnLoad(JNIEnv* env, const char* 
                    "getBytes", "()[B", error);
 
     NETTY_JNI_UTIL_LOAD_CLASS(env, byteArrayClass, "[B", error);
+    if (packagePrefix)
+        staticPackagePrefix = strdup(packagePrefix);
 
     return NETTY_JNI_UTIL_JNI_VERSION;
 error:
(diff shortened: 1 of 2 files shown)


openssl-dynamic/src/main/c/jnilib.c
```diff
@@ -214,6 +215,8 @@ static jint netty_internal_tcnative_Library_JNI_OnLoad(JNIEnv* env, const char*
                    "getBytes", "()[B", error);
 
     NETTY_JNI_UTIL_LOAD_CLASS(env, byteArrayClass, "[B", error);
+    if (packagePrefix)
```
normanmaurer, line 218:
```suggestion
    if (packagePrefix) {
```

openssl-dynamic/src/main/c/jnilib.c
```diff
@@ -214,6 +215,8 @@ static jint netty_internal_tcnative_Library_JNI_OnLoad(JNIEnv* env, const char*
...
+        staticPackagePrefix = strdup(packagePrefix);
```
normanmaurer, line 220:
```suggestion
}
```

normanmaurer (changes requested):

openssl-dynamic/src/main/c/jnilib.c
```diff
@@ -214,7 +215,9 @@ static jint netty_internal_tcnative_Library_JNI_OnLoad(JNIEnv* env, const char*
                    "getBytes", "()[B", error);
 
     NETTY_JNI_UTIL_LOAD_CLASS(env, byteArrayClass, "[B", error);
-
+    if (packagePrefix) {
+        staticPackagePrefix = strdup(packagePrefix);
+}
```
normanmaurer, line 220:
```suggestion
    }
```

normanmaurer (changes requested):

PR #650, opened 2021-06-15
Introduce the ability to run key-signing in an fully async fashion.

normanmaurer:
Motivation:

Sometimes we may want to run tasks in a fully async fashion. For example if we need to do a network operation to full fill the task.

This commit only makes use of the real async support for key-signing but we could also do the same for other tasks.

Modifications:

- Add AsyncTask that allows to run a task in an async fashion and notify the callback once done.
- Add AsyncSSLPrivateKeyMethod that makes use of the AsyncTask and so allow for a fully async implementation for key signing.

Result:

More flexible

### DIFF
--- a/openssl-dynamic/src/main/c/sslcontext.c
+++ b/openssl-dynamic/src/main/c/sslcontext.c
@@ -59,6 +59,8 @@ static jmethodID sslPrivateKeyMethodSignTask_init;
 static jclass    sslPrivateKeyMethodDecryptTask_class;
 static jmethodID sslPrivateKeyMethodDecryptTask_init;
 
+static const char* staticPackagePrefix = NULL;
+
 extern apr_pool_t *tcn_global_pool;
 
 static apr_status_t ssl_context_cleanup(void *data)
@@ -2240,11 +2242,14 @@ const SSL_PRIVATE_KEY_METHOD private_key_method = {
 #endif // OPENSSL_IS_BORINGSSL
 
 
-TCN_IMPLEMENT_CALL(void, SSLContext, setPrivateKeyMethod)(TCN_STDARGS, jlong ctx, jobject method) {
+TCN_IMPLEMENT_CALL(void, SSLContext, setPrivateKeyMethod0)(TCN_STDARGS, jlong ctx, jobject method) {
     tcn_ssl_ctxt_t *c = J2P(ctx, tcn_ssl_ctxt_t *);
 
     TCN_CHECK_NULL(c, ctx, /* void */);
 #ifdef OPENSSL_IS_BORINGSSL
+    char* name = NULL;
+    char* combinedName = NULL;
+
     jobject oldMethod = c->ssl_private_key_method;
     if (method == NULL) {
         c->ssl_private_key_method = NULL;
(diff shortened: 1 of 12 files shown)


openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -2259,13 +2264,21 @@ TCN_IMPLEMENT_CALL(void, SSLContext, setPrivateKeyMethod)(TCN_STDARGS, jlong ctx
             return;
         }
 
-        jmethodID signMethod = (*e)->GetMethodID(e, method_class, "sign", "(JI[B)[B");
+        NETTY_JNI_UTIL_PREPEND(staticPackagePrefix, "io/netty/internal/tcnative/ResultCallback;)V", name, error);
+        NETTY_JNI_UTIL_PREPEND("(JI[BL", name, combinedName, error);
+        TCN_REASSIGN(name, combinedName);
```
trustin, line 2269: Out of curiosity - do we have an automated test that makes sure we handle shaded classes correctly, so we don't make a mistake?

openssl-dynamic/src/main/java/io/netty/internal/tcnative/AsyncSSLPrivateKeyMethodAdapter.java
```diff
@@ -0,0 +1,51 @@
...
+    @Override
+    public void sign(long ssl, int signatureAlgorithm, byte[] input, ResultCallback<byte[]> resultCallback) {
+        final byte[] result;
+        try {
+            result = method.sign(ssl, signatureAlgorithm, input);
+        } catch (Throwable cause) {
+            resultCallback.onError(ssl, cause);
+            return;
+        }
+        resultCallback.onSuccess(ssl, result);
```
trustin, line 37: Maybe move this into the try block so we don't have to return in the catch block?

openssl-dynamic/src/main/java/io/netty/internal/tcnative/AsyncSSLPrivateKeyMethodAdapter.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Override
+    public void decrypt(long ssl, byte[] input, ResultCallback<byte[]> resultCallback) {
+        final byte[] result;
+        try {
+            result = method.decrypt(ssl, input);
+        } catch (Throwable cause) {
+            resultCallback.onError(ssl, cause);
+            return;
```
trustin, line 47: Ditto

openssl-dynamic/src/main/java/io/netty/internal/tcnative/CertificateCallbackTask.java
```diff
@@ -35,15 +35,15 @@
 
     // See https://www.openssl.org/docs/man1.0.2/man3/SSL_set_cert_cb.html.
     @Override
-    protected int runTask(long ssl) {
+    protected void runTask(long ssl, TaskCallback taskCallback) {
```
trustin, line 38: Not related with this PR, but do we have a runtime check for TCNative version, so our users get more friendly exception message in case of version mismatch?

openssl-dynamic/src/main/java/io/netty/internal/tcnative/SSLContext.java
```diff
@@ -649,15 +649,33 @@ public static void setAlpnProtos(long ctx, String[] alpnProtos, int selectorFail
...
      * @param ctx context to use
      * @param method method to use for the given context.
      */
-    public static native void setPrivateKeyMethod(long ctx, SSLPrivateKeyMethod method);
+    public static void setPrivateKeyMethod(long ctx, final SSLPrivateKeyMethod method) {
+        setPrivateKeyMethod(ctx, new AsyncSSLPrivateKeyMethodAdapter(method));
+    }
+
+    /**
+     * Set the {@link AsyncSSLPrivateKeyMethod} to use for the given {@link SSLContext}.
```
trustin, line 666:
```suggestion
     * Sets the {@link AsyncSSLPrivateKeyMethod} to use for the given {@link SSLContext}.
```

openssl-dynamic/src/main/java/io/netty/internal/tcnative/SSLContext.java
```diff
@@ -649,15 +649,33 @@ public static void setAlpnProtos(long ctx, String[] alpnProtos, int selectorFail
...
+     * This allows to offload private key operations if needed.
+     *
+     * This method is currently only supported when {@code BoringSSL} is used.
```
trustin, line 669: Should we log a warning once when this method is called without BoringSSL?

openssl-dynamic/src/main/c/sslcontext.c
normanmaurer: yes we have it in netty as part of the `testsuite-shading`

openssl-dynamic/src/main/java/io/netty/internal/tcnative/CertificateCallbackTask.java
normanmaurer: @trustin yes we have in netty itself.

openssl-dynamic/src/main/java/io/netty/internal/tcnative/SSLContext.java
normanmaurer: We already this in netty :)

openssl-dynamic/src/main/java/io/netty/internal/tcnative/AsyncSSLPrivateKeyMethodAdapter.java
normanmaurer: I did this by design as I dint want to have `resultCallback.onSuccess` call `onError` in case itself throws.

openssl-dynamic/src/main/java/io/netty/internal/tcnative/AsyncSSLPrivateKeyMethodAdapter.java
normanmaurer: See above

normanmaurer:
Let me pull this in

trustin (commented):
Belated LGTM

PR #654, opened 2021-08-17
Allow to set renegiotion mode when BoringSSL is used

normanmaurer:
Motivation:

renegiotion is not supported by BoringSSL by default but sometimes when using on the client-side you might want to support it for CLIENT-AUTH.
See also https://github.com/apple/swift-nio-ssl/pull/111

Modifications:

Add native / java code to be able to enable support

Result:

Part of https://github.com/netty/netty/issues/11529


slandelle:
Awesome, thanks a lot!
I'm still on PTO this week. I'll be able to test snapshots when I'm back.

chrisvest (approved):

PR #661, opened 2021-09-06
Allow to set the curves used

normanmaurer:
Motivation:

Sometimes it is useful to restrict the curves that are used.

Modifications:

Expose SSL_CTX_set1_curves_list via JNI

Result:

Be able to configure the curves that are used for a context

### DIFF
--- a/openssl-dynamic/src/main/java/io/netty/internal/tcnative/SSLContext.java
+++ b/openssl-dynamic/src/main/java/io/netty/internal/tcnative/SSLContext.java
@@ -694,4 +694,31 @@ public final class SSLContext {
      * @return {@code true} if successful, {@code false} otherwise.
      */
     public static native boolean setNumTickets(long ctx, int tickets);
+
+    /**
+     * Sets the curves to use.
+     *
+     * See <a href="https://www.openssl.org/docs/man1.1.0/man3/SSL_CTX_set1_curves_list.html">SSL_CTX_set1_curves_list</a>.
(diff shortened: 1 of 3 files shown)


openssl-dynamic/src/main/java/io/netty/internal/tcnative/SSLContext.java
```diff
@@ -694,4 +694,31 @@ public static void setPrivateKeyMethod(long ctx, AsyncSSLPrivateKeyMethod method
...
+        if (curves.length == 0) {
+            throw new IllegalArgumentException();
+        }
+        StringBuilder sb = new StringBuilder();
+        for (String curve: curves) {
+            sb.append(curve);
+            // Curves are separated by : as explained in the manpage.
+            sb.append(':');
+        }
+        sb.setLength(sb.length() - 1);
```
chrisvest, line 719: `StringJoiner` (new in Java 8) can simplify this.

chrisvest (approved):
Had a suggestion, but leaving it up to you if it's worth it or not.

openssl-dynamic/src/main/java/io/netty/internal/tcnative/SSLContext.java
normanmaurer: I think it is not worth it to depend on java8 just for that

chrisvest (approved):
Looks like an ordeal to get old openssl versions supported

PR #662, opened 2021-09-08
Add support for mac m1

normanmaurer:
Motivation:

We should support m1 as more and more users are switching to it.

Modifications:

- Add profile for boringssl-static to cross compile for m1
- Adjust finish release script to also include m1

Result:

Mac M1 is supported

### DIFF
--- a/boringssl-static/pom.xml
+++ b/boringssl-static/pom.xml
@@ -1196,7 +1196,6 @@
                   <libDirectory>${nativeLibOnlyDir}</libDirectory>
                   <forceAutogen>${forceAutogen}</forceAutogen>
                   <forceConfigure>${forceConfigure}</forceConfigure>
-                  <windowsBuildTool>msbuild</windowsBuildTool>
                   <!-- <verbose>true</verbose> -->
                   <configureArgs>
                     <configureArg>--with-ssl=no</configureArg>
--- a/boringssl-static/pom.xml
+++ b/boringssl-static/pom.xml
@@ -733,6 +733,14 @@
                       <groupId>io.netty</groupId>
                       <artifactId>netty-tcnative-boringssl-static</artifactId>
                       <version>${project.version}</version>
+                      <classifier>osx-aarch_64</classifier>
+                      <type>jar</type>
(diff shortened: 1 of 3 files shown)


hyperxpro:
Does compiling require an actual M1 chipset? I mean, will you be manually compiling binary on your M1 machine?

normanmaurer:
@hyperxpro no its cross-compiling on intel.

boringssl-static/pom.xml
```diff
@@ -962,6 +974,248 @@
...
+    <!-- Profile to cross-compile for mac m1 -->
+    <profile>
+      <id>osx-m1-cross-compile</id>
+
+      <properties>
+        <linkStatic>true</linkStatic>
+        <osxCrossCompile>true</osxCrossCompile>
+        <macOsxDeploymentTarget/>
+        <cmakeOsxDeploymentTarget/>
+        <cflags>-target arm64-apple-macos11 -Werror -fno-omit-frame-pointer -fvisibility=hidden -Wunused -Wno-unused-value -O3</cflags>
```
chrisvest, line 986: Not in this PR, but at some point I think it would be good to have `-Werror` be controlled by a property, so people can turn it off if they have a different compiler or compiler version.

boringssl-static/pom.xml
```diff
@@ -962,6 +974,248 @@
...
+                </goals>
+                <phase>compile</phase>
+                <configuration>
+                  <name>netty_tcnative</name>
+                  <nativeSourceDirectory>${generatedSourcesDir}/c</nativeSourceDirectory>
+                  <customPackageDirectory>${generatedSourcesDir}/native-package</customPackageDirectory>
+                  <libDirectory>${nativeLibOnlyDir}</libDirectory>
+                  <forceAutogen>${forceAutogen}</forceAutogen>
+                  <forceConfigure>${forceConfigure}</forceConfigure>
+                  <windowsBuildTool>msbuild</windowsBuildTool>
```
chrisvest, line 1199: Is the `windowsBuildTool` needed?

normanmaurer: nope. this can be removed... copy and paste left-over

Issue #657, opened 2021-09-01
tcnative > 2.0.35 Fails to start in osgi


karthickm512:
netty-tcnative bundles above 2.0.35 version fail to start in Osgi environment citing 'osgi r3 bundle not supported'.

The reason is, the Manifest of tcnative does not have any import/export declaration while tcnative version 2.0.35 and less have import/export defined.

Please add this info so that latest versions can be used,

normanmaurer:
Do you want to do a PR?

karthickm512:
sorry whats a PR? I am a consumer of tcnative osgified jar

NiteshKant:
PR is an acronym for Pull Request. In other words, can you make the changes you are requesting here? We are happy to review the changes for you.

karthickm512:
Sorry I have no coding knowledge, so it would be better if I don't code :) And any Karaf users should be facing this issue for months now.

==Following is what is found in Manifest of 2.0.35.Final==
Manifest-Version: 1.0
Ant-Version: Apache Ant 1.8.2
Created-By: Apache Maven Bundle Plugin
Implementation-Title: Netty/TomcatNative [OpenSSL - Dynamic]
Bundle-Description: A Mavenized fork of Tomcat Native which incorporat
 es various patches. This artifact is dynamically linked    to OpenSSL
  and Apache APR.
Automatic-Module-Name: io.netty.tcnative.openssl.dynamic
Bundle-License: http://www.apache.org/licenses/LICENSE-2.0
Bundle-SymbolicName: io.netty.tcnative
Implementation-Version: 2.0.35.Final
Archiver-Version: Plexus Archiver
Built-By: root
Bnd-LastModified: 1607414176986
Bundle-ManifestVersion: 2
Implementation-Vendor-Id: io.netty
Bundle-DocURL: http://netty.io/
Bundle-Vendor: The Netty Project
**Import-Package**: org.eclipse.jetty.alpn;version="[1,2)";resolution:=opt
 ional,org.eclipse.jetty.npn;version="[1,2)";resolution:=optional,sun.
 nio.ch;resolution:=optional
Tool: Bnd-2.1.0.20130426-122213
Implementation-Vendor: The Netty Project
**Export-Package**: io.netty.internal.tcnative;version="2.0.35.Final"
Bundle-Name: Netty/TomcatNative [OpenSSL - Dynamic]
Bundle-Version: 2.0.35.Final
[...]

PR #664, opened 2021-09-14
Fix OSGI support

normanmaurer:
Motivation:

ef6dd68503eaefdad800a7b905a1ad21a7e1408f did break OSGI support as we did not correctly include the generated MANIFEST

Modifications:

Fix configuration of plugins to include the correct MANIFEST file which has all the entries for OSGI

Result:

Fixes https://github.com/netty/netty-tcnative/issues/657

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -398,7 +398,27 @@
           </execution>
         </executions>
       </plugin>
-
+      <plugin>
+        <artifactId>maven-jar-plugin</artifactId>
+        <version>3.0.2</version>
+        <executions>
+          <execution>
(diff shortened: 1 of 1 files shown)


chrisvest (commented):
I guess we don't have automated tests in tcnative to cover this? I mean, I can't easily tell if this XML is correct or not, and we missed it in the previous PR.

normanmaurer:
@chrisvest nope and I am far away from being a OSGI expert... I just checked that it included that entries that were included before 🤷

chrisvest (approved):

PR #670, opened 2021-10-01
Use curl to download cmake

normanmaurer:
Motivation:

Trying to use wget to download cmake gives certificate errors. Using curl fixes it

Modifications:

Switch to curl to download cmake

Result:

Be able to build docker image again


hyperxpro:
Let's Encrypt's Root CA expired on 30-Sept-2021. So maybe this is the side-effect of it,

Issue #614, opened 2021-01-28
Consider adding support for certificate compression (TLSv1.3)


normanmaurer:
https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#Certificate-compression

CamW:
I am interested in working on this. Is it something you would like to add support for?

If so, it seems to me that dynamically linking and loading native compression libraries might be a good approach. To avoid the bloat of statically linking and the call to java for compression.
Any thoughts on that?

normanmaurer:
@CamW I think it would be better to do the compression in a java callback to miminize trouble :)

PR #689, opened 2022-01-06
Support RFC 8879 Certificate Compression for TLSv1.3 (#614)

CamW:
Motivation:

Certificates in TLSv1.3 can be compressed but Netty does not yet support this. BoringSSL does support it and this change will use that support to provide an API which Netty can use to support certificate compression.

Modifications:

- `CertificateCompressionAlgo.java` class to allow compression implementations to be built in Java.
- Native `addCertificateCompressionAlgorithm0` which allows registration of compression algorithms with BoringSSL.
- JNI work to maintain Java compression algorithm callbacks and call them when BoringSSL needs a certificate compressed or decompressed.

Result:

New API, which Netty can use to provide support for certificate compression.

### DIFF
--- a/openssl-classes/src/main/java/io/netty/internal/tcnative/CertificateCompressionAlgo.java
+++ b/openssl-classes/src/main/java/io/netty/internal/tcnative/CertificateCompressionAlgo.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2022 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package io.netty.internal.tcnative;
+
+/**
+ * Provides compression/decompression implementations for TLS Certificate Compression
+ * (<a href="https://tools.ietf.org/html/rfc8879">RFC 8879</a>).
+ */
+public interface CertificateCompressionAlgo {
+    int TLS_EXT_CERT_COMPRESSION_ZLIB   = NativeStaticallyReferencedJniMethods.tlsExtCertCompressionZlib();
+    int TLS_EXT_CERT_COMPRESSION_BROTLI = NativeStaticallyReferencedJniMethods.tlsExtCertCompressionBrotli();
+    int TLS_EXT_CERT_COMPRESSION_ZSTD   = NativeStaticallyReferencedJniMethods.tlsExtCertCompressionZstd();
+
+    /**
+     * Compress the given input with the specified algorithm and return the compressed bytes.
+     *
+     * @param ssl           the SSL instance
+     * @param input         the uncompressed form of the certificate
+     * @return              the compressed form of the certificate
+     * @throws Exception    thrown if an error occurs while compressing
+     */
+    byte[] compress(long ssl, byte[] input) throws Exception;
+
+    /**
+     * Decompress the given input with the specified algorithm and return the decompressed bytes.
+     *
+     * <h3>Implementation
+     * <a href="https://tools.ietf.org/html/rfc8879#section-5">Security Considerations</a></h3>
+     * <p>Implementations SHOULD bound the memory usage when decompressing the CompressedCertificate message.</p>
+     * <p>
+     * Implementations MUST limit the size of the resulting decompressed chain to the specified {@code uncompressedLen},
+     * and they MUST abort the connection (throw an exception) if the size of the output of the decompression
+     * function exceeds that limit.
+     * </p>
+     *
+     * @param ssl               the SSL instance
+     * @param uncompressedLen   the expected length of the uncompressed certificate
+     * @param input             the compressed form of the certificate
+     * @return                  the decompressed form of the certificate
+     * @throws Exception        thrown if an error occurs while decompressing or output
+     * size exceeds {@code uncompressedLen}
+     */
+    byte[] decompress(long ssl, int uncompressedLen, byte[] input) throws Exception;
+
+    /**
+     * Return the ID for the compression algorithm provided for by a given implementation.
+     *
+     * @return compression algorithm ID as specified by RFC8879
+     * <PRE>
+     * {@link CertificateCompressionAlgo#TLS_EXT_CERT_COMPRESSION_ZLIB}
+     * {@link CertificateCompressionAlgo#TLS_EXT_CERT_COMPRESSION_BROTLI}
+     * {@link CertificateCompressionAlgo#TLS_EXT_CERT_COMPRESSION_ZSTD}
+     * </PRE>
+     */
+    int algorithmId();
+
+}
--- a/openssl-dynamic/src/main/c/cert_compress.c
+++ b/openssl-dynamic/src/main/c/cert_compress.c
@@ -1,0 +1,182 @@
+/*
+ * Copyright 2022 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "tcn.h"
+#include "ssl_private.h"
+#ifdef OPENSSL_IS_BORINGSSL
+#include "cert_compress.h"
+
+static int compress(jobject compression_algorithm, jmethodID compress_method, SSL* ssl, CBB* out,
+    const uint8_t* in, size_t in_len) {
+
+    int ret = 0;
+    JNIEnv *e = NULL;
+    jbyteArray inputArray = NULL;
+
+    if (compression_algorithm == NULL || compress_method == NULL) {
+        goto complete;
+    }
+    if (tcn_get_java_env(&e) != JNI_OK) {
+        goto complete;
+    }
+    if ((inputArray = (*e)->NewByteArray(e, in_len)) == NULL) {
+        goto complete;
+    }
+
+    (*e)->SetByteArrayRegion(e, inputArray, 0, in_len, (jbyte*) in);
+
+    jbyteArray resultArray = (*e)->CallObjectMethod(e, compression_algorithm, compress_method,
+                    P2J(ssl), inputArray);
+
+    if ((*e)->ExceptionCheck(e) != JNI_FALSE) {
+        (*e)->ExceptionClear(e);
+        goto complete; // Exception while calling into Java
+    }
+    if (resultArray == NULL) {
+        goto complete; // Received NULL array from call to Java
+    }
+
+    int resultLen = (*e)->GetArrayLength(e, resultArray);
+    uint8_t* outData = NULL;
+    if (!CBB_reserve(out, &outData, resultLen)) {
+        goto complete; // Unable to reserve space for compressed data
+    }
+    jbyte* resultData = (*e)->GetByteArrayElements(e, resultArray, NULL);
+    memcpy(outData, resultData, resultLen);
+    if (!CBB_did_write(out, resultLen)) {
+        goto complete; // Unable to advance bytes written to CBB
+    }
+    (*e)->ReleaseByteArrayElements(e, resultArray, resultData, JNI_ABORT);
+    ret = 1; // Success
+    goto complete;
+
+complete:
+    // Free up any allocated memory and return.
+    if (inputArray != NULL) {
+        (*e)->DeleteLocalRef(e, inputArray);
+    }
+    return ret;
+}
+
+static int decompress(jobject compression_algorithm, jmethodID decompress_method, SSL* ssl, CRYPTO_BUFFER** out,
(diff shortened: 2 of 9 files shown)


CamW:
It looks like testing is usually done on the netty repo for this sort of thing. I've got a commit for the integration into netty and testing available https://github.com/CamW/netty/commit/ef81b86429f3712432cb7e1b25b417b8205e51ce
That'll only work once this is available though.

normanmaurer:
@CamW there are build failures on windows:

```
NFO] 
23808
[INFO] "D:\a\netty-tcnative\netty-tcnative\boringssl-static\target\native-build\vs2010.vcxproj" (default target) (1) ->
23809
[INFO] (Link target) -> 
23810
[INFO]   sslcontext.obj : error LNK2001: unresolved external symbol brotli_decompress_java [D:\a\netty-tcnative\netty-tcnative\boringssl-static\target\native-build\vs2010.vcxproj]
23811
[INFO]   sslcontext.obj : error LNK2001: unresolved external symbol zlib_decompress_java [D:\a\netty-tcnative\netty-tcnative\boringssl-static\target\native-build\vs2010.vcxproj]
23812
[INFO]   sslcontext.obj : error LNK2001: unresolved external symbol zstd_compress_java [D:\a\netty-tcnative\netty-tcnative\boringssl-static\target\native-build\vs2010.vcxproj]
23813
[INFO]   sslcontext.obj : error LNK2001: unresolved external symbol zstd_decompress_java [D:\a\netty-tcnative\netty-tcnative\boringssl-static\target\native-build\vs2010.vcxproj]
23814
[INFO]   sslcontext.obj : error LNK2001: unresolved external symbol zlib_compress_java [D:\a\netty-tcnative\netty-tcnative\boringssl-static\target\native-build\vs2010.vcxproj]
23815
[INFO]   sslcontext.obj : error LNK2001: unresolved external symbol brotli_compress_java [D:\a\netty-tcnative\netty-tcnative\boringssl-static\target\native-build\vs2010.vcxproj]
23816
[INFO]   D:\a\netty-tcnative\netty-tcnative\boringssl-static\target\native-build\/target/x64-release/lib\netty_tcnative.dll : fatal error LNK1120: 6 unresolved externals [D:\a\netty-tcnative\netty-tcnative\boringssl-static\target\native-build\vs2010.vcxproj]
23817
[INFO] 
23818
[INFO]     40 Warning(s)
```

openssl-dynamic/src/main/c/cert_compress.c
```diff
@@ -0,0 +1,182 @@
...
+
+    int resultLen = (*e)->GetArrayLength(e, resultArray);
+    uint8_t* outData = NULL;
+    if (!CBB_reserve(out, &outData, resultLen)) {
+        goto complete; // Unable to reserve space for compressed data
+    }
+    jbyte* resultData = (*e)->GetByteArrayElements(e, resultArray, NULL);
+    memcpy(outData, resultData, resultLen);
+    if (!CBB_did_write(out, resultLen)) {
+        goto complete; // Unable to advance bytes written to CBB
```
normanmaurer, line 60: In this case we still need to call:

```
(*e)->ReleaseByteArrayElements(e, resultArray, resultData, JNI_ABORT);
```

openssl-dynamic/src/main/c/cert_compress.c
```diff
@@ -0,0 +1,182 @@
...
+    }
+    (*e)->ReleaseByteArrayElements(e, resultArray, resultData, JNI_ABORT);
+    ret = 1; // Success
+    goto complete;
+
+complete:
+    // Free up any allocated memory and return.
+    if (inputArray != NULL) {
+        (*e)->DeleteLocalRef(e, inputArray);
```
normanmaurer, line 69: honestly I think we don't need to delete the local ref explicit.

openssl-dynamic/src/main/c/cert_compress.c
```diff
@@ -0,0 +1,182 @@
...
+    }
+    jbyte* resultData = (*e)->GetByteArrayElements(e, resultArray, NULL);
+    memcpy(outData, resultData, uncompressed_len);
+    (*e)->ReleaseByteArrayElements(e, resultArray, resultData, JNI_ABORT);
+    ret = 1; // Success
+
+complete:
+    // Free up any allocated memory and return.
+    if (inputArray != NULL) {
+        (*e)->DeleteLocalRef(e, inputArray);
```
normanmaurer, line 122: same comment as above

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -2647,6 +2666,88 @@ TCN_IMPLEMENT_CALL(jboolean, SSLContext, setCurvesList0)(TCN_STDARGS, jlong ctx,
...
 }
 
+TCN_IMPLEMENT_CALL(jint, SSLContext, addCertificateCompressionAlgorithm0)(TCN_STDARGS, jlong ctx, jint direction, jint algorithmId, jobject algorithm) {
+    tcn_ssl_ctxt_t *c = J2P(ctx, tcn_ssl_ctxt_t *);
+    TCN_CHECK_NULL(c, ctx, 0);
+    if (algorithm == NULL) {
+        tcn_ThrowIllegalArgumentException(e, "Compression algorithm may not be null");
+    }
+    if (!(direction & SSL_CERT_COMPRESSION_DIRECTION_COMPRESS) && !(direction & SSL_CERT_COMPRESSION_DIRECTION_DECOMPRESS)) {
+        tcn_ThrowIllegalArgumentException(e, "Invalid direction specified");
```
normanmaurer, line 2676: We need to return early.

```suggestion
        tcn_ThrowIllegalArgumentException(e, "Invalid direction specified");
        return 0;
```

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -2647,6 +2666,88 @@ TCN_IMPLEMENT_CALL(jboolean, SSLContext, setCurvesList0)(TCN_STDARGS, jlong ctx,
     return ret == 1 ? JNI_TRUE : JNI_FALSE;
 }
 
+TCN_IMPLEMENT_CALL(jint, SSLContext, addCertificateCompressionAlgorithm0)(TCN_STDARGS, jlong ctx, jint direction, jint algorithmId, jobject algorithm) {
+    tcn_ssl_ctxt_t *c = J2P(ctx, tcn_ssl_ctxt_t *);
+    TCN_CHECK_NULL(c, ctx, 0);
+    if (algorithm == NULL) {
+        tcn_ThrowIllegalArgumentException(e, "Compression algorithm may not be null");
```
normanmaurer, line 2673: We need to return early.

```suggestion
        tcn_ThrowIllegalArgumentException(e, "Compression algorithm may not be null");
        return 0;
```

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -2647,6 +2666,88 @@ TCN_IMPLEMENT_CALL(jboolean, SSLContext, setCurvesList0)(TCN_STDARGS, jlong ctx,
...
+                direction & SSL_CERT_COMPRESSION_DIRECTION_COMPRESS ? zstd_compress_java : NULL,
+                direction & SSL_CERT_COMPRESSION_DIRECTION_DECOMPRESS ? zstd_decompress_java : NULL);
+            if (result) {
+                c->ssl_cert_compression_zstd_algorithm = algoRef;
+                c->ssl_cert_compression_zstd_compress_method = compressMethod;
+                c->ssl_cert_compression_zstd_decompress_method = decompressMethod;
+            }
+            break;
+        default:
+            tcn_ThrowException(e, "Unrecognized certificate compression algorithm");
```
normanmaurer, line 2738: Otherwise we will put a second exception on the stack below. 
```suggestion
             (*e)->DeleteGlobalRef(e, algoRef);
            tcn_ThrowException(e, "Unrecognized certificate compression algorithm");
            return 0;
```

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -2647,6 +2666,88 @@ TCN_IMPLEMENT_CALL(jboolean, SSLContext, setCurvesList0)(TCN_STDARGS, jlong ctx,
...
+    }
+
+    int result = 0;
+    switch (algorithmId) {
+        case TLSEXT_cert_compression_zlib:
+            result = SSL_CTX_add_cert_compression_alg(c->ctx, algorithmId,
+                direction & SSL_CERT_COMPRESSION_DIRECTION_COMPRESS ? zlib_compress_java : NULL,
+                direction & SSL_CERT_COMPRESSION_DIRECTION_DECOMPRESS ? zlib_decompress_java : NULL);
+            if (result) {
+                c->ssl_cert_compression_zlib_algorithm = algoRef;
```
normanmaurer, line 2712: We also need to guard against the fact that the user might call the method multiple times with the same `algorithmId`. At the moment this will case a memory leak.

```suggestion
                if (c->ssl_cert_compression_zlib_algorithm != NULL) {
                    (*e)->DeleteGlobalRef(e, c->ssl_cert_compression_zlib_algorithm);
                }
                c->ssl_cert_compression_zlib_algorithm = algoRef;
```

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -2647,6 +2666,88 @@ TCN_IMPLEMENT_CALL(jboolean, SSLContext, setCurvesList0)(TCN_STDARGS, jlong ctx,
...
+                c->ssl_cert_compression_zlib_compress_method = compressMethod;
+                c->ssl_cert_compression_zlib_decompress_method = decompressMethod;
+            }
+            break;
+        case TLSEXT_cert_compression_brotli:
+            result = SSL_CTX_add_cert_compression_alg(c->ctx, algorithmId,
+                direction & SSL_CERT_COMPRESSION_DIRECTION_COMPRESS ? brotli_compress_java : NULL,
+                direction & SSL_CERT_COMPRESSION_DIRECTION_DECOMPRESS ? brotli_decompress_java : NULL);
+            if (result) {
+                c->ssl_cert_compression_brotli_algorithm = algoRef;
```
normanmaurer, line 2722: Same as above
```suggestion
                if (c->ssl_cert_compression_brotli_algorithm != NULL) {
                    (*e)->DeleteGlobalRef(e, c->ssl_cert_compression_brotli_algorithm);
                }
                c->ssl_cert_compression_brotli_algorithm = algoRef;
```

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -2647,6 +2666,88 @@ TCN_IMPLEMENT_CALL(jboolean, SSLContext, setCurvesList0)(TCN_STDARGS, jlong ctx,
...
+                c->ssl_cert_compression_brotli_compress_method = compressMethod;
+                c->ssl_cert_compression_brotli_decompress_method = decompressMethod;
+            }
+            break;
+        case TLSEXT_cert_compression_zstd:
+            result = SSL_CTX_add_cert_compression_alg(c->ctx, algorithmId,
+                direction & SSL_CERT_COMPRESSION_DIRECTION_COMPRESS ? zstd_compress_java : NULL,
+                direction & SSL_CERT_COMPRESSION_DIRECTION_DECOMPRESS ? zstd_decompress_java : NULL);
+            if (result) {
+                c->ssl_cert_compression_zstd_algorithm = algoRef;
```
normanmaurer, line 2732: Same comment as above
```suggestion
                if (c->ssl_cert_compression_zstd_algorithm != NULL) {
                    (*e)->DeleteGlobalRef(e, c->ssl_cert_compression_zstd_algorithm);
                }
                c->ssl_cert_compression_zstd_algorithm = algoRef;
```

normanmaurer (changes requested):
Great work!

openssl-dynamic/src/main/c/sslcontext.c
CamW: I have a test that shows that only the first call, per sslContext, will succeed so this shouldn't happen but better to be safe, I think :-)

openssl-dynamic/src/main/c/sslcontext.c
CamW: fixed

openssl-dynamic/src/main/c/sslcontext.c
CamW: fixed

openssl-dynamic/src/main/c/cert_compress.c
CamW: removed

openssl-dynamic/src/main/c/cert_compress.c
CamW: removed

CamW:
Thanks @normanmaurer for the review, it should be good now. I've tried but not got it building on windows, should just be that extra file. If that doesn't fix it, I'll get a local windows build working.

normanmaurer (changes requested):

openssl-classes/src/main/java/io/netty/internal/tcnative/CertificateCompressionAlgo.java
```diff
@@ -0,0 +1,70 @@
...
+     * <a href="https://tools.ietf.org/html/rfc8879#section-5">Security Considerations</a></h3>
+     * <p>Implementations SHOULD bound the memory usage when decompressing the CompressedCertificate message.</p>
+     * <p>
+     * Implementations MUST limit the size of the resulting decompressed chain to the specified {@code uncompressedLen},
+     * and they MUST abort the connection (throw an exception) if the size of the output of the decompression
+     * function exceeds that limit.
+     * </p>
+     *
+     * @param ssl               the SSL instance
+     * @param uncompressedLen   the expected length of the uncompressed certificate
```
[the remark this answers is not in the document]
CamW: @normanmaurer I wouldn't say that this was an upper bound, no. It is precisely the size of the uncompressed cert. Instead of sending the cert, a peer would send a `CompressedCertificate` message which is as follows:

```
struct {
            CertificateCompressionAlgorithm algorithm;
            uint24 uncompressed_length;
            opaque compressed_certificate_message<1..2^24-1>;
       } CompressedCertificate;
```

I say in the comment that it's the "expected" length because a malicious peer could send a compression bomb to cause a poor implementation to over-allocate space.

normanmaurer:
Great work @CamW ! 

Can you open a PR against netty which consume this ?

PR #690, opened 2022-01-13
Explicit define the target when cross compile

normanmaurer:
Motivation:

We explicitly need to set the target when we cross compile as otherwise it might fail

Modifications:

- Explicit set the target
- Define properties in the same place

Result:

No problem when cross compile


normanmaurer:
See also https://forum.dlang.org/post/qvnqqa$ulf$1@digitalmars.com

PR #747, opened 2022-09-15
Add profile to cross compile for intel on macOs

normanmaurer:
Motivation:

As m1/m2 becomes more used we should also have a profile to cross compile for intel.

Modifications:

- Add profile to cross-compile for intel when running on m1/m2.
- Adjust finish_release.sh script to use correct cross-compile profile based on the architecture.

Result:

Be able to cross compile for intel and also running the release on m1/m2

### DIFF
--- a/boringssl-static/pom.xml
+++ b/boringssl-static/pom.xml
@@ -1133,11 +1133,11 @@
         <cppflags>-target ${target} -DHAVE_OPENSSL -I${boringsslSourceDir}/include</cppflags>
         <ldflags>-arch arm64 -L${boringsslHome}/ssl -L${boringsslHome}/crypto -lssl -lcrypto</ldflags>
         <!-- use aarch_64 as this is also what os.detected.arch will use on an aarch64 system -->
-        <nativeJarFile>${project.build.directory}/${project.build.finalName}-${os.detected.name}-osx_64.jar</nativeJarFile>
-        <nativeLibOsParts>${os.detected.name}_osx_64</nativeLibOsParts>
-        <jniClassifier>${os.detected.name}-osx_64</jniClassifier>
+        <nativeJarFile>${project.build.directory}/${project.build.finalName}-${os.detected.name}-osx_x86_64.jar</nativeJarFile>
+        <nativeLibOsParts>${os.detected.name}_osx_x86_64</nativeLibOsParts>
+        <jniClassifier>${os.detected.name}-osx_x86_64</jniClassifier>
         <jniArch>osx_64</jniArch>
-        <javaModuleNameClassifier>${os.detected.name}.osx_64</javaModuleNameClassifier>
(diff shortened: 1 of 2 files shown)


boringssl-static/pom.xml
```diff
@@ -1113,6 +1113,254 @@
...
+        <cmakeCxxFlags>-O3 -fno-omit-frame-pointer -target ${target} -Wno-error=range-loop-analysis</cmakeCxxFlags>
+        <cflags>-target ${target} -Werror -fno-omit-frame-pointer -fvisibility=hidden -Wunused -Wno-unused-value -O3</cflags>
+        <cppflags>-target ${target} -DHAVE_OPENSSL -I${boringsslSourceDir}/include</cppflags>
+        <ldflags>-arch arm64 -L${boringsslHome}/ssl -L${boringsslHome}/crypto -lssl -lcrypto</ldflags>
+        <!-- use aarch_64 as this is also what os.detected.arch will use on an aarch64 system -->
+        <nativeJarFile>${project.build.directory}/${project.build.finalName}-${os.detected.name}-osx_64.jar</nativeJarFile>
+        <nativeLibOsParts>${os.detected.name}_osx_64</nativeLibOsParts>
+        <jniClassifier>${os.detected.name}-osx_64</jniClassifier>
+        <jniArch>osx_64</jniArch>
+        <javaModuleNameClassifier>${os.detected.name}.osx_64</javaModuleNameClassifier>
```
chrisvest, line 1140: What is the `osx_64` part used in these lines? Should it be `osx_x86_64` or something like that?

scripts/finish_release.sh
```diff
@@ -20,13 +20,19 @@ if git tag | grep -q "$2" ; then
     exit 1
 fi
 
+CROSS_COMPILE_PROFILE="mac-m1-cross-compile"
+if [ "$ARCH" == "arm" ]; then
```
chrisvest, line 24: On my machine, `$ARCH` is not defined. I do, however, have the following:

```
% uname -p
arm
```

scripts/finish_release.sh
```diff
@@ -20,13 +20,19 @@ if git tag | grep -q "$2" ; then
...
+    PROFILE="mac-intel-cross-compile"
```
chrisvest, line 25: The variable assigned here should be `CROSS_COMPILE_PROFILE`

chrisvest (changes requested):

boringssl-static/pom.xml
normanmaurer: typo

scripts/finish_release.sh
normanmaurer: missed to copy some stuff... fixed

chrisvest (approved):

PR #753, opened 2022-11-15
Remove weak symbol when building with BoringSSL

davidben:
(I wasn't able to figure out the build, but I'm hoping CI will catch if I messed anything up.)

Motivation:

Current revisions of BoringSSL have upgraded the `unsigned` parameter to a `size_t` one. (Folks requested that we start to work through -Wshorten-64-to-32 warnings.) This is broadly API-compatible but breaks netty-tcnative's weak symbol machinery because it copies the functionp prototype.

Modification:

Move the remaining functions BoringSSL always has into the OPENSSL_IS_BORINGSSL block and disable the weak symbol checks.

Result:

netty-tcnative builds with BoringSSL.

Further work can probably be done here to clean this up, depending on what the project wants. First of all, it's 2022 and you can probably just stop supporting OpenSSL 1.0.1. Then much of this goes away anyway.

Beyond that, BoringSSL usually does feature detection with either OPENSSL_VERSION_NUMBER (we define a roughly analogous OpenSSL version) or our BORINGSSL_API_VERSION symbol. If you went with a strategy like that, it'd be a bit easier for us, though I don't know if it'd be a headache for other cases. Or perhaps we can come up with a suitable abstraction here.

### DIFF
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -1255,7 +1255,7 @@ TCN_IMPLEMENT_CALL(jstring, SSL, getAlpnSelected)(TCN_STDARGS,
                                                          jlong ssl /* SSL * */) {
     // Use weak linking with GCC as this will alow us to run the same packaged version with multiple
     // version of openssl.
-    #if defined(__GNUC__) || defined(__GNUG__)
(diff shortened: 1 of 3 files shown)


normanmaurer:
Thanks a lot @davidben !

PR #754, opened 2022-11-19
Fix BoringSSL osx cross-compilation incorrect binary output

zekronium:
With the current maven configuration, when compiling from apple-silicon to intel produces arch64 binaries instead of x86_64 (via profile `mac-intel-cross-compile`)

```
❯ file libnetty_tcnative.jnilib
libnetty_tcnative.jnilib: Mach-O 64-bit dynamically linked shared library arm64
```

This PR replaces the wrong parameters in the compilation which now produces the correct target binaries
```
❯ file libnetty_tcnative.jnilib
libnetty_tcnative.jnilib: Mach-O 64-bit dynamically linked shared library x86_64
```


normanmaurer (approved):

normanmaurer:
@zekronium thanks a lot!

PR #757, opened 2023-01-03
Expose SSL3_RT_MAX_ENCRYPTED_LENGTH via constant

normanmaurer:
Motivation:

We need to expose SSL3_RT_MAX_ENCRYPTED_LENGTH so we can use it to detemine the capacity of src ByteBuffer for unwrap operations.

Modifications:

Add method and constant that exposed SSL3_RT_MAX_ENCRYPTED_LENGTH

Result:

Be able to correctly size the capacity of the ByteBuffer


normanmaurer:
This is related to https://github.com/netty/netty/issues/13073

chrisvest (approved):

Issue #703, opened 2022-02-17
Link against to the libcrypt.so.2


floating-cat:
OS: Linux
Library: netty-tcnative-*-static

I just noticed that I am unable to run a Netty app in ArchLinux because Arch's glibc package has removed the libcrypt.so.1: https://bbs.archlinux.org/viewtopic.php?id=274160
So I need to install the `libxcrypt-compat` to run it.
From my checking, this is the same in the Fedora, and I think there would be more distro to remove it in the future.

To build a `libcrypt.so.1` from [besser82/libxcrypt](https://github.com/besser82/libxcrypt), the package need to not use the `--disable-obsolete-api` flag when building which is not used by some distros, and we need to install the `libxcrypt-compat` additionally.

Would it better if the `netty-tcnative-*-static` link against to the libcrypt.so.2 and recommend the users to install the `libxcrypt` package instead if they don't have the `libcrypt.so.2` installed?

Thanks.

Some reference links:
https://github.com/pypa/manylinux/issues/305#issuecomment-489332594
https://fedoraproject.org/wiki/Changes/Replace_glibc_libcrypt_with_libxcrypt

arloor:
I ran into this problem on RHEL9 too.

```shell
Caused by: java.lang.UnsatisfiedLinkError: /tmp/libnetty_tcnative_linux_x86_641905760841177269486.so: libcrypt.so.1: cannot open shared object file: No such file or directory
        at jdk.internal.loader.NativeLibraries.load(Native Method) ~[?:?]
        at jdk.internal.loader.NativeLibraries$NativeLibraryImpl.open(NativeLibraries.java:388) ~[?:?]
        at jdk.internal.loader.NativeLibraries.loadLibrary(NativeLibraries.java:232) ~[?:?]
        at jdk.internal.loader.NativeLibraries.loadLibrary(NativeLibraries.java:174) ~[?:?]
        at java.lang.ClassLoader.loadLibrary(ClassLoader.java:2389) ~[?:?]
        at java.lang.Runtime.load0(Runtime.java:755) ~[?:?]
        at java.lang.System.load(System.java:1953) ~[?:?]
        at io.netty.util.internal.NativeLibraryUtil.loadLibrary(NativeLibraryUtil.java:36) ~[forwardproxy-1.0-jar-with-dependencies.jar:?]
        ... 23 more
```

PR #758, opened 2023-01-09
Patch APR to remove crypt dependency

normanmaurer:
Motivation:

We dont use any APR features that uses libcrypt so we can safely not link against it. This resolves various issues as some distributions otherwise need the extra installation of libxcrypt-compat etc.

Modifications:

- Add patchfile for APR configure.in to remove linking against libcrypt
- Patch APR during build

Result:

Fixes https://github.com/netty/netty-tcnative/issues/703

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -64,7 +64,7 @@ jobs:
     #    uses a compiled language
 
     - name: Install tools / libraries 
-      run: sudo apt-get update && sudo apt-get -y install autoconf automake libtool make tar libapr1-dev libssl-dev cmake perl ninja-build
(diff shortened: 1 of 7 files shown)


normanmaurer:
```
 ldd boringssl-static/target/native-lib-only/META-INF/native/linux64/libnetty_tcnative.so
ldd: warning: you do not have execution permission for `boringssl-static/target/native-lib-only/META-INF/native/linux64/libnetty_tcnative.so'
	linux-vdso.so.1 (0x00007ffc2ef80000)
	libc.so.6 => /lib64/libc.so.6 (0x00007fa555a00000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fa5560de000)
```

chrisvest (approved):

PR #762, opened 2023-01-17
Remove compile scope junit dependencies from openssl-dynamic submodule

### LINKED ISSUE #761: JUnit dependencies included transitively in compile scope when depending on netty-tcnative
JUnit dependencies are being incorrectly transitively included when depending on netty-tcnative.

For example, consider the following pom.xml:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">

    <modelVersion>4.0.0</modelVersion>

    <groupId>com.example</groupId>
    <artifactId>example</artifactId>
    <version>1.0.0-SNAPSHOT</version>
    <packaging>pom</packaging>

    <properties>
        <netty.version>4.1.87.Final</netty.version>
    </properties>

    <dependencyManagement>
        <dependencies>
            <dependency>
                <groupId>io.netty</groupId>
                <artifactId>netty-bom</artifactId>
                <version>${netty.version}</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>
        </dependencies>
    </dependencyManagement>

    <dependencies>
        <dependency>
            <groupId>io.netty</groupId>
            <artifactId>netty-tcnative</artifactId>
            <classifier>linux-x86_64-fedora</classifier>
        </dependency>
    </dependencies>
</project>
```

Running `mvn dependency:tree` shows the junit dependencies being incorrectly included...

```
[INFO] --- maven-dependency-plugin:2.8:tree (default-cli) @ example ---
[INFO] com.example:example:pom:1.0.0-SNAPSHOT
[INFO] \- io.netty:netty-tcnative:jar:linux-x86_64-fedora:2.0.56.Final:compile
[INFO]    +- io.netty:netty-tcnative-classes:jar:2.0.56.Final:compile
[INFO]    +- org.junit.jupiter:junit-jupiter-api:jar:5.9.0:compile
[INFO]    |  +- org.opentest4j:opentest4j:jar:1.2.0:compile
[INFO]    |  +- org.junit.platform:junit-platform-commons:jar:1.9.0:compile
[INFO]    |  \- org.apiguardian:apiguardian-api:jar:1.1.2:compile
[INFO]    \- org.junit.jupiter:junit-jupiter-engine:jar:5.9.0:compile
[INFO]       \- org.junit.platform:junit-platform-engine:jar:1.9.0:compile
```

I expect to only see...

```
[INFO] --- maven-dependency-plugin:2.8:tree (default-cli) @ example ---
[INFO] com.example:example:pom:1.0.0-SNAPSHOT
[INFO] \- io.netty:netty-tcnative:jar:linux-x86_64-fedora:2.0.56.Final:compile
[INFO]    +- io.netty:netty-tcnative-classes:jar:2.0.56.Final:compile
```

I believe the netty-tcnative poms should include `<scope>test</scope>` anywhere the junit dependencies are declared (e.g. [here](https://github.com/netty/netty-tcnative/blob/77047924fb9ee6f86935f45528ca7c215cbd8dec/openssl-dynamic/pom.xml#L164-L174)), rather than just in dependency management (e.g. [here](https://github.com/netty/netty-tcnative/blob/77047924fb9ee6f86935f45528ca7c215cbd8dec/pom.xml#L475-L486)).

maven 3.8.7
netty 4.1.87.Final
netty-tcnative 2.0.56.Final

philsttr:
Motivation:

So junit dependencies are not transitively included when users depend on netty-tcnative in compile scope.

Modification:

Removed the junit dependencies from the openssl-dynamic submodule, so that they are inherited from netty-tcnative-parent with the proper test scope.

netty-tcnative-parent declares the junit dependencies as test scope in its dependencies.  But the openssl-dynamic submodule was overriding the scope to be the default (compile) scope.

Result:

Junit dependencies are not transitively included when users depend on netty-tcnative in compile scope.

Fixes #761

### DIFF
--- a/openssl-dynamic/pom.xml
+++ b/openssl-dynamic/pom.xml
@@ -161,16 +161,6 @@
       <artifactId>netty-tcnative-classes</artifactId>
       <version>${project.version}</version>
     </dependency>
-    <dependency>
-      <groupId>org.junit.jupiter</groupId>
-      <artifactId>junit-jupiter-api</artifactId>
-      <version>${junit.version}</version>
-    </dependency>
-    <dependency>
-      <groupId>org.junit.jupiter</groupId>
-      <artifactId>junit-jupiter-engine</artifactId>
-      <version>${junit.version}</version>
-    </dependency>
   </dependencies>
 
   <profiles>


openssl-dynamic/pom.xml
```diff
@@ -161,16 +161,6 @@
...
-    <dependency>
-      <groupId>org.junit.jupiter</groupId>
-      <artifactId>junit-jupiter-api</artifactId>
-      <version>${junit.version}</version>
-    </dependency>
-    <dependency>
-      <groupId>org.junit.jupiter</groupId>
-      <artifactId>junit-jupiter-engine</artifactId>
-      <version>${junit.version}</version>
-    </dependency>
```
normanmaurer, line 173: hmm.. shouldn't these be still included here but have the right scope ?

normanmaurer: I think the correct solution would be to have these just declared in a dependency management in the parent and with the right scope here.

philsttr: Note that there are a couple other dependencies declared in the parent as well (netty-build-common and netty-jni-util).  Do you want to leave them in dependencies in the parent, or move them to dependencyManagement in the parent and dependencies in each submodule?

normanmaurer: lets just merge this one for now and I will think about the rest

PR #763, opened 2023-01-17
Change libeay32 and ssleay32 to libcrypto and libssl

iliark:
Unable to build openssl-dynamic because libeay32 and ssleay32 names are not in use now. Names were changed to libcrypto and libssl according to this comment: https://github.com/openssl/openssl/issues/10332

After changes have been made build was done successfully.


normanmaurer:
@iliark thanks a lot

PR #767, opened 2023-02-02
Bump up os-maven-plugin to 1.7.1

Panxuefeng-loongson:
Motivation:

os-maven-plugin 1.6.2 does not recognize LoongArch information, os-maven-plugin supports LoongArch since 1.7.1

Modification:

Update pom.xml

Result:

LoongArch cpu information can be correctly identified


normanmaurer:
Thanks

PR #779, opened 2023-03-20
Set read-only permissions on GitHub workflows

### LINKED ISSUE #778: Set read-only token permissions for GitHub Action workflows
By default, GitHub workflows run with write-all permissions. This is dangerous, since it opens the project up to supply-chain attacks. [GitHub itself](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-secrets) recommends ensuring all workflows run with minimal permissions.

netty-tcnative's workflows are mainly used for testing. As such, they do not need such broad permissions.

This issue can be solved in two ways:
- add top-level read-only permissions to all workflows, and then give additional job-level permissions as necessary; and/or
- set the default token permissions to read-only in the repo settings.

I'll be sending a PR along with this issue that sets the top-level permissions as necessary.

If you instead (or also) wish to modify the default token permissions:
1. Open the repo settings
2. Go to [Actions > General](https://github.com/netty/netty-tcnative/settings/actions)
3. Under "Workflow permissions", set them to "Read repository contents and packages permissions"

If you wish to only modify the default token permissions, know that you'll need to modify any workflows that require additional permissions.

---

**Disclosure:** My name is Pedro and I work with Google and the [Open Source Security Foundation (OpenSSF)](https://www.openssf.org/) to improve the supply-chain security of the open-source ecosystem.

pnacht:
Fixes #778

As per the linked issue, this workflow sets read-only top-level permissions on all workflows.

It is my understanding that the jobs in `release.yml` all require write permissions (they all set `git config`, so I assume commits will be written and pushed), so those have been given at the job level. This ensures that if a new job is added in the future that doesn't require write permissions, it'll only have read access.

If I misunderstood something and other workflows require additional permissions or some jobs in `release.yml` don't require write access, let me know and I'll fix the PR.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -10,6 +10,9 @@ on:
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 env:
   MAVEN_OPTS: -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false -Dmaven.wagon.http.retryhandler.count=5 -Dmaven.wagon.httpconnectionManager.ttlSeconds=240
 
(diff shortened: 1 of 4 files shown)


normanmaurer:
@pnacht merged... thanks!

PR #783, opened 2023-04-21
Replace wget usage with curl

normanmaurer:
Motivation:

Lets be consistent and just use curl

Modifications:

Replace wget with curl

Result:

Cleanup


trustin:
This changeset seems pretty dangerous? As least we should verify the checksum and PGP signature...

normanmaurer:
I will do a followup... I need to figure out how to convince Debian to use another trust store

Issue #788, opened 2023-05-04
Memory leak on SslProvider instantiation with netty-tcnative-boringssl-static


notdryft:
We observed a memory leak that happens by switching from `netty-tcnative-boringssl-static` version `2.0.56.Final` to `2.0.57.Final` (or later, we tested up to `2.0.60.Final`).

Scenario is as follow:

* Create a `SslContextBuilder` once

```java
SslContextBuilder sslContextBuilder = SslContextBuilder.forClient.sslProvider(SslProvider.OPENSSL_REFCNT)
sslContextBuilder.ciphers(null, IdentityCipherSuiteFilter.INSTANCE_DEFAULTING_TO_SUPPORTED_CIPHERS)
```

For us the issue happened with both `SslProvider.OPENSSL_REFCNT` and `SslProvider.OPENSSL`.

Then, inside a loop:

* `SslContext sslContext = sslContextBuilder.build()`
* `ReferenceCountUtil.release(sslContext)`

Do that millions of times and you can see the following happen, left is `2.0.56.Final` and after the memory drop, `2.0.57.Final`:

<img width="721" alt="Screenshot 2023-05-04 at 15 12 10" src="https://user-images.githubusercontent.com/2608594/236214755-9473dc61-9745-4c07-a908-ce230e77aa19.png">

In our test case, `-Xms` and `-Xmx` were both `1G` so memory shouldn't be expected to grow that much more as everything is recycled right after instantiation.

Sample project that reproduces the issue provided as an attachment: [test-netty-tcnative-leak.zip](https://github.com/netty/netty-tcnative/files/11397904/test-netty-tcnative-leak.zip)

Command to run the project: `mvn compile exec:java -Dexec.mainClass=Main`

notdryft:
Forgot to mention something. When trying to pin down the issue, we used `-Dio.netty.leakDetection.level=paranoid` but it printed nothing!

slandelle:
Caused by #759 ?

normanmaurer:
Please upgrade to 2.0.60.Final Am 04.05.2023 um 16:35 schrieb Stephane Landelle ***@***.***>:﻿
Caused by #759 ?

slandelle:
@normanmaurer 

> or later, we tested up to 2.0.60.Final

Included. 2.0.60 included is the version we currently have in production and it suffers from the same issue. We were able to figure out the regression was introduced in  2.0.57.

slandelle:
[...]

PR #790, opened 2023-05-05
Fix memory leak when creating and destroying a lot of context

normanmaurer:
Motivation:

During introducing https://github.com/netty/netty-tcnative/pull/759 I missed to add a free call which did introduce a small memory leak. This was not noticable for most users as creating and destroying of SSL context is usually very rare. That said some users might still do this and so notice the leak.

Modifications:

- Add missing free(...) call to match calloc(...)
- Move code of one function to the other as it was the only calling function. This simplifies the code and makes it less error-prone

Result:

Fixes https://github.com/netty/netty-tcnative/issues/788

### DIFF
--- a/openssl-dynamic/src/main/c/sslcontext.c
+++ b/openssl-dynamic/src/main/c/sslcontext.c
@@ -62,133 +62,6 @@ static jmethodID sslPrivateKeyMethodDecryptTask_init;
 
 static const char* staticPackagePrefix = NULL;
 
-
-static void ssl_context_cleanup(tcn_ssl_ctxt_t *c)
-{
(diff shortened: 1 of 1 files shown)


franz1981 (approved):
@normanmaurer you can give a shot to the async profiler native memory leak branch!

notdryft:
We're currently testing it against the async profiler nativemem mode ourselves!

normanmaurer:
I could reproduce a leak anymore with this pulled in. Once you confirm it as well I will do a release

franz1981:
Great @notdryft let me know if that worked too!

PR #796, opened 2023-06-29
SSL_CVERIFY_REQUIRED to follow API docs

Scottmitch:
Motivation:
The OpenSSL docs specify that the `SSL_VERIFY_FAIL_IF_NO_PEER_CERT` flag must be masked with `SSL_VERIFY_PEER`. Currently netty hard codes a value of `2` for required which is just the value of `SSL_VERIFY_FAIL_IF_NO_PEER_CERT`.

[1] https://www.openssl.org/docs/man1.0.2/man3/SSL_CTX_set_verify.html
> Server mode: if the client did not return a certificate, the TLS/SSL
handshake is immediately terminated with a "handshake failure" alert. This flag must be used together with SSL_VERIFY_PEER.


Scottmitch:
nvm we are translating the values correctly under the covers: https://github.com/netty/netty-tcnative/blob/main/openssl-dynamic/src/main/c/sslutils.c#L642-L666

PR #803, opened 2023-07-20
Fix compilation on some configurations

morvael:
Adds -Wno-error=stringop-overflow flag to BoringSSL compilation so that it starts working again on specific OS/architecture configuration.


normanmaurer:
@morvael thanks a lot!

PR #807, opened 2023-07-27
Fix linux build for openssl-static

hasnain-db:
It looks like https://github.com/netty/netty-tcnative/commit/0828ffc73499269bbc6ffd8bf5ad3c98b0a67b9d broke the build on linux (my apologies!).

The openssl build now puts the libraries into `/lib64` instead of `/lib` so the following steps wouldn't include the native code in the jar. This change asks the build to put it into the old location (similar to the Mac build, which is what I tested in the last PR).

I tested this on a linux machine and verified that the build is now ~similarly sized to before the previous commit (and that it works)


normanmaurer (approved):

normanmaurer:
@hasnain-db thanks a lot!

PR #819, opened 2023-09-04
Hash-pin GitHub Actions and use dependabot to keep them updated

### LINKED ISSUE #818: Hash-pin workflow GitHub Actions
Hey, it's Pedro (see #779 and #781) and I'm back with a new security suggestion!

When developing with CI workflows, it's common to version-pin dependencies (i.e. `actions/checkout@v3`). However, version tags are mutable, so a malicious attacker could overwrite a version tag to point to a malicious or vulnerable commit instead.

Pinning workflow dependencies by hash ensures the dependency is immutable and its behavior is guaranteed.

These hashes can be automatically updated by dependabot. Whenever a new version of an Action is released, you'll receive a PR updating both its hash and the version comment (see [this repo](https://github.com/pnacht/logr/pulls) as an example).

I'll send a PR pinning the Actions along with this issue.

pnacht:
Fixes #818.

This PR hash-pins all GitHub Actions to ensure their behavior is as expected, protecting the project from supply-chain attacks.

These hashes (and version comments) will be kept up-to-date by dependabot.

`ci-release.yml` uses `crazy-max/ghaction-import-gpg`. It was at v3, but the Action is now at v5. The only breaking change was to one argument name, so I've taken the liberty of making that bump.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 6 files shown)


pnacht:
Let me know if this is something the project is interested in. If not, feel free to close!

normanmaurer:
Sorry we ignored this for too long.. Will do a new PR to fix it.

PR #820, opened 2023-09-05
Remove obsolete flag when building BoringSSL

normanmaurer:
Motivation:

-DOPENSSL_C11_ATOMIC is not needed anymore for BoringSSL as it uses atomics by default now.

Modifications:

Remove -DOPENSSL_C11_ATOMIC

Result:

Cleanup


normanmaurer:
@davidben I am correct here ?

davidben:
Yeah, you shouldn't need it with a recent BoringSSL. As a minor nitpick, the we require `static_assert` not specifically for atomics, but in general now. (For non-MSVC compilers, we require C11 these days. MSVC will probably follow soon.)

PR #836, opened 2023-10-22
fix(boringssl): fix incorrect build profile for osx intel cross compile

zekronium:
Motivation:

boringssl-static profile `mac-intel-cross-compile` incorrectly sets arch resulting in `osx-osx-x86_64` build classifier.
It also wrongfully sets the native lib name to `libnetty_tcnative_osx_osx_64`, making `netty-handler` unable to find it

If the intention of this profile was to build cross platform (from linux to mac), then `os.detected.name` can be overridden to still produce the similar result.
```xml
<os.detected.name>osx</os.detected.name>
```

Modifications:
Adjust arch to `x86_64` in maven profile for boringssl static

Result:
`./mvn -Pmac-intel-cross-compile -am -pl boringssl-static clean package` builds correctly on an apple silicon system


normanmaurer (approved):

normanmaurer:
@zekronium thanks a lot !

Issue #853, opened 2024-02-21
sys_siglist: symbol not found on alpine


codefromthecrypt:
I've noticed several issues discussing compat issues with alpine, and not sure if this last error is a problem or not.

If someone can analyze this in the context of netty-tcnative (vs from a higher layer user like grpc or armeria), would be fantastic. **biggest question: is ` sys_siglist: symbol not found` going to cause problems in prod, or is this just fuzz**

Here's the repro, using the latest release on an image with alpine 3.19.1 (used by zipkin with armeria)

```bash
$ docker run --rm -it --entrypoint /bin/sh ghcr.io/openzipkin/java:21.0.2_p13
/java # uname -a
Linux e71082a2a46f 6.5.0-15-generic #15-Ubuntu SMP PREEMPT_DYNAMIC Tue Jan  9 22:39:36 UTC 2024 aarch64 Linux
/java # wget -q https://repo1.maven.org/maven2/io/netty/netty-tcnative-boringssl-static/2.0.63.Final/netty-tcnative-boringssl-static-2.0.63.Final-linux-aarch_64.jar
/java # jar -xf netty-tcnative-boringssl-static-2.0.63.Final-linux-aarch_64.jar META-INF/native
/java # apk info gcompat 2>&-
gcompat-1.1.0-r4 description:
GNU C Library compatibility layer for musl

gcompat-1.1.0-r4 webpage:
https://git.adelielinux.org/adelie/gcompat

gcompat-1.1.0-r4 installed size:
148 KiB

/java # ldd META-INF/native/libnetty_tcnative_linux_aarch_64.so 
	/lib/ld-musl-aarch64.so.1 (0xffffb148c000)
	librt.so.1 => /lib/ld-musl-aarch64.so.1 (0xffffb148c000)
	libcrypt.so.1 => /lib/libcrypt.so.1 (0xffffb125a000)
	libdl.so.2 => /lib/ld-musl-aarch64.so.1 (0xffffb148c000)
	libc.so.6 => /lib/ld-musl-aarch64.so.1 (0xffffb148c000)
	libucontext.so.1 => /lib/libucontext.so.1 (0xffffb1239000)
	libobstack.so.1 => /usr/lib/libobstack.so.1 (0xffffb1218000)
Error relocating META-INF/native/libnetty_tcnative_linux_aarch_64.so: sys_siglist: symbol not found
```

[...]

PR #854, opened 2024-02-26
Disable usage of sys_siglist function to workaround issues on alpine

normanmaurer:
Motivation:

When we try to use our static compiled version on alpine it fails due missing sys_siglist. We can workaround this by letting apr use its own implementation.

Modifications:

Use ac_cv_have_decl_sys_siglist=no when compiling APR.

Result:

netty-tcnative-boringssl-static should be useable on alpine. Fixes https://github.com/netty/netty-tcnative/issues/853


ejona86 (approved):

normanmaurer:
ok problem is still there... need to investigate further

PR #861, opened 2024-03-14
Add key log callback option to SSLContext

chrisvest:
Motivation:
Wireshark can decrypt TLS sessions during packet capture, if the session keys (etc.) are available from an SSL key log file. This log file format has become a de facto industry standard, and BoringSSL (and maybe the others too, didn't check) has a callback mechanism for delivering log lines in this format.

Modification:
Add `KeyLogCallback` interface and an `SSLContext.setKeyLogCallback` method, which integrators can easily implement the SSLKEYLOGFILE feature, or equivalent, on top of.

Result:
It is now possible to configure netty-tcnative in a way that TLS sessions can be decrypted at packet-capture time by Wireshark, making it easier to investigate and debug problems with TLS.

### DIFF
--- a/openssl-dynamic/src/main/c/sslcontext.c
+++ b/openssl-dynamic/src/main/c/sslcontext.c
@@ -2552,7 +2552,7 @@ static void keylog_cb(const SSL* ssl, const char *line) {
 
     JNIEnv *e = NULL;
     if (tcn_get_java_env(&e) != JNI_OK) {
-        // There's nothing we cna do with the JNIEnv*.
+        // There's nothing we can do with the JNIEnv*.
         return;
     }
 
--- a/openssl-dynamic/src/main/c/sslcontext.c
+++ b/openssl-dynamic/src/main/c/sslcontext.c
@@ -100,6 +100,13 @@ static apr_status_t ssl_context_cleanup(void *data)
             }
             c->ssl_cert_compression_zstd_algorithm = NULL;
         }
+        if (c->keylog_callback != NULL) {
+            if (e != NULL) {
+                (*e)->DeleteGlobalRef(e, c->keylog_callback);
+            }
+            c->keylog_callback = NULL;
+        }
+        c->keylog_callback_method = NULL;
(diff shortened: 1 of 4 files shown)


chrisvest:
Not only a de facto standard, but perhaps on its way to becoming a formalized standard as well: https://datatracker.ietf.org/doc/draft-ietf-tls-keylogfile/

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -2531,6 +2538,86 @@ TCN_IMPLEMENT_CALL(void, SSLContext, setSniHostnameMatcher)(TCN_STDARGS, jlong c
...
+
+    tcn_ssl_state_t *state = tcn_SSL_get_app_state(ssl);
+    if (state == NULL || state->ctx == NULL) {
+        // There's nothing we can do without tcn_ssl_state_t.
+        return;
+    }
+
+    JNIEnv *e = NULL;
+    if (tcn_get_java_env(&e) != JNI_OK) {
+        // There's nothing we cna do with the JNIEnv*.
```
normanmaurer, line 2555:
```suggestion
        // There's nothing we can do with the JNIEnv*.
```

normanmaurer (approved):
Generally speaking looks good. Just two nits.

hyperxpro:
I will write an example for this and try to couple it with PcapWriteHandler.

makkarpov:
For now, it seems like there is no public API to access the required context pointer.

`long ctx` is a private field in `io.netty.handler.ssl.ReferenceCountedOpenSslContext`. Deprecated-but-public `sslCtxPointer()` is not an option, since function expects pointer to `tcn_ssl_ctxt_t`, not `SSL_CTX`. Attempt to use it will result in a segfault.

It looks like a reflective access to the `private long ctx` field is required to actually use this API. Luckily, Netty is an automatic module, and therefore is open by default. So this actually works without any `--add-opens` in modern JDKs. But still looks like a hack.

But otherwise the API works, thank you for implementing this.

PR #867, opened 2024-04-12
Add OpenSSL 3.x support to openssl-dynamic

klaren:
### Motivation:
The `openssl-dynamic` module only compiles against OpenSSL 1.x.

### Modifications:
The most significant change is that one no longer needs to provide DH parameters manually. By calling `SSL_CTX_set_dh_auto()`, OpenSSL will use the built-in parameters that match the key pair size. E.g. RSA 2048 will use DH 2048 and so on.

The property `jdk.tls.ephemeralDHKeySize`, which was used to affect the size of the DH parameters before, will be ignored when using the dynamically linked OpenSSL. Instead, the system OpenSSL library's config will be used.

Another change is how FIPS is managed because OpenSSL 3 introduces a "Provider" concept. The `FIPS_mode_set` is removed and is now replaced with `EVP_default_properties_enable_fips` which instructs OpenSSL to load the FIPS compliant provider, if no provider is explicitly requested.

### Result:
The `openssl-dynamic` module can now successfully compile and run with OpenSSL 3.x.

-------------------------

This only makes the necessary changes to the code. There is still the question of how this should be published(if at all?) since it has to be a separate artifact from the 1.x version. Some alternatives I see are
* Completely drop 1.x support. This would also unnest all the pre-processing spaghetti in the code.
* Publish under a separate classifier.
* Some hybrid of the above by creating a new `openssl3-dynamic` module.

### DIFF
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -58,6 +58,12 @@ static UI_METHOD *ui_method = NULL;
 #endif // OPENSSL_NO_ENGINE
 
 
+#if defined(OPENSSL_FIPS) && (OPENSSL_VERSION_NUMBER < 0x30000000L)
+#define tcn_enable_fips(to)   FIPS_mode_set((to))
+#else
+#define tcn_enable_fips(to)   EVP_default_properties_enable_fips(NULL, (to))
+#endif
+
 #if OPENSSL_VERSION_NUMBER < 0x10100000L || (defined(LIBRESSL_VERSION_NUMBER) && LIBRESSL_VERSION_NUMBER < 0x2090000fL)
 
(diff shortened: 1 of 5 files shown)


normanmaurer:
@klaren thanks a lot for the contribution... Did you sign our ICLA yet, https://netty.io/s/icla ?

klaren:
@normanmaurer yes, I signed it just before I created the PR.

normanmaurer:
@klaren can you please check the build failures: https://github.com/netty/netty-tcnative/actions/runs/8658366916/job/23968334532?pr=867 ?

klaren:
@normanmaurer Minor typo, should be fixed now.

normanmaurer:
Thanks a lot @klaren

PR #869, opened 2024-04-26
Upgrade Debian builds to version 12

klaren:
Motivation:
Debian 7 went EOL in May 2018. This shipped with OpenSSL 1.1.1 which went EOL on 11th September 2023.

Modification:
Update the debian docker image to version 12 and minor cleanups of the docker files.

Result:
The dynamically linked `netty-tcnative-linux-x86_64` artifact now links with supported libraries.


klaren:
```shell
> ldd libnetty_tcnative_linux_x86_64.so
	linux-vdso.so.1 (0x00007fffe54ef000)
	libssl.so.3 => /lib/x86_64-linux-gnu/libssl.so.3 (0x000073b8784ca000)
	libcrypto.so.3 => /lib/x86_64-linux-gnu/libcrypto.so.3 (0x000073b877e00000)
	libapr-1.so.0 => /lib/x86_64-linux-gnu/libapr-1.so.0 (0x000073b87848e000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x000073b877a00000)
	libuuid.so.1 => /lib/x86_64-linux-gnu/libuuid.so.1 (0x000073b878484000)
	/lib64/ld-linux-x86-64.so.2 (0x000073b8785a5000)
```

slandelle:
If this move (breaking change) was accepted, the plain crazy build chain we at Gatling created for [brotli4j](https://github.com/hyperxpro/Brotli4j) to retain compatibility with Debian 7 could also be way simplified.
cc @ericmolle @hyperxpro

klaren:
@normanmaurer I think this could be considered a breaking change, so I'm not sure whether you want this or not. But since everything is EOL maybe it's okay. We needed this so I'm just sharing what I have.

I also have a RHEL9 compatible one(built on almalinux since CentOS is discontinued) if you are interested. That one is a bit more tricky since Red Hat will backport security fixes for OpenSSL 1.x until 2029 for the people with big wallets.

Issue #888, opened 2024-10-23
libcrypt: arm64 / x86_64 different dependencies on EL9


flozano:
EDIT: Likely same as #722
- netty 4.1.114
- tcnative 2.0.66
- AlmaLinux EL9
- OpenJDK 21.0.5 (tested with AWS and Temurin)
- both archs x64 and arm64

```
$ ls netty*
netty-buffer-4.1.114.Final.jar                                     netty-resolver-dns-4.1.114.Final.jar
netty-codec-4.1.114.Final.jar                                      netty-tcnative-boringssl-static-2.0.66.Final.jar
netty-codec-dns-4.1.114.Final.jar                                  netty-tcnative-boringssl-static-2.0.66.Final-linux-aarch_64.jar
netty-codec-haproxy-4.1.114.Final.jar                              netty-tcnative-boringssl-static-2.0.66.Final-linux-x86_64.jar
netty-codec-http2-4.1.114.Final.jar                                netty-tcnative-boringssl-static-2.0.66.Final-osx-aarch_64.jar
netty-codec-http-4.1.114.Final.jar                                 netty-tcnative-boringssl-static-2.0.66.Final-osx-x86_64.jar
netty-codec-socks-4.1.114.Final.jar                                netty-tcnative-boringssl-static-2.0.66.Final-windows-x86_64.jar
netty-codec-xml-4.1.114.Final.jar                                  netty-tcnative-classes-2.0.66.Final.jar
netty-common-4.1.114.Final.jar                                     netty-transport-4.1.114.Final.jar
netty-common-999.2-SNAPSHOT.jar                                    netty-transport-classes-epoll-4.1.114.Final.jar
netty-handler-4.1.114.Final.jar                                    netty-transport-classes-kqueue-4.1.114.Final.jar
netty-handler-proxy-4.1.114.Final.jar                              netty-transport-native-epoll-4.1.114.Final.jar
netty-incubator-codec-classes-quic-0.0.68.Final.jar                netty-transport-native-epoll-4.1.114.Final-linux-aarch_64.jar
netty-incubator-codec-native-quic-0.0.68.Final.jar                 netty-transport-native-epoll-4.1.114.Final-linux-x86_64.jar
netty-incubator-codec-native-quic-0.0.68.Final-linux-aarch_64.jar  netty-transport-native-kqueue-4.1.114.Final.jar
[...]

PR #891, opened 2024-10-24
Patch APR in docker image thats is used to cross-compile to remove li…

normanmaurer:
…bcrypt dependency.

Motivation:

We don't need to depend on libcrypt when cross-compiling. We already don't depend on it when we don't cross-compile

Modifications:

Patch APR as part of the docker image creation and so remove libcrypt dependency

Result:

Fixes https://github.com/netty/netty-tcnative/issues/888

### DIFF
--- a/docker/Dockerfile.cross_compile_aarch64
+++ b/docker/Dockerfile.cross_compile_aarch64
@@ -45,7 +45,7 @@ RUN set -x && \
   tar xvf apr-$APR_VERSION.tar.gz && \
   pushd apr-$APR_VERSION && \
   patch -p0 $SOURCE_DIR/apr_crypt.patch && \
-  buildconf && \
+  ./buildconf && \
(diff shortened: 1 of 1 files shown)


normanmaurer:
Verified there is no libcrypt linking anymore with this PR:

```
root@71991d92312f:/tmp/build-pr-centos7-aarch64-jars/boringssl-static/target# ldd META-INF/native/libnetty_tcnative_linux_aarch_64.so
	linux-vdso.so.1 (0x0000ffff8518a000)
	librt.so.1 => /lib/aarch64-linux-gnu/librt.so.1 (0x0000ffff84f34000)
	libdl.so.2 => /lib/aarch64-linux-gnu/libdl.so.2 (0x0000ffff84f20000)
	libc.so.6 => /lib/aarch64-linux-gnu/libc.so.6 (0x0000ffff84dad000)
	libpthread.so.0 => /lib/aarch64-linux-gnu/libpthread.so.0 (0x0000ffff84d7c000)
	/lib/ld-linux-aarch64.so.1 (0x0000ffff8515a000)
```
cc @slandelle

flozano:
now it also depends on pthread like x86_64 non-crosscompiled 👯

slandelle:
Awesome, thanks a lot Norman!

flozano:
thank you!!!

PR #913, opened 2025-02-24
Add method to allow to set curves / groups per SSL instance

normanmaurer:
Motivation:

At the moment its only possible to set the curves / groups per SSL_CTX instance. Sometimes we need to pick them based on the SSL instance.

Modifications:

Add methods to allow setting curves / groups

Result:

Be able to specify curves / groups per SSL instance

### DIFF
--- a/openssl-classes/src/main/java/io/netty/internal/tcnative/SSL.java
+++ b/openssl-classes/src/main/java/io/netty/internal/tcnative/SSL.java
@@ -566,7 +566,7 @@ public final class SSL {
     /**
      * Sets the curves to use.
      *
-     * See <a href="https://www.openssl.org/docs/man1.1.0/man3/SSL_set1_curves_list.html">SSL_set1_curves_list</a>.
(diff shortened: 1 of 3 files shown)


openssl-classes/src/main/java/io/netty/internal/tcnative/SSL.java
```diff
@@ -562,6 +562,53 @@ public static boolean setCipherSuites(long ssl, String ciphers)
      */
     public static native boolean setCipherSuites(long ssl, String ciphers, boolean tlsv13)
             throws Exception;
+
+    /**
+     * Sets the curves to use.
+     *
+     * See <a href="https://www.openssl.org/docs/man1.1.0/man3/SSL_set1_curves_list.html">SSL_set1_curves_list</a>.
```
chrisvest, line 569: I get 404 on these links.

chrisvest (approved):

PR #915, opened 2025-03-18
Use branch 'main' of boringssl

charlesconnell:
The BoringSSL project's default branch is named `main`, not `master`. They have a `master` branch but it contains out-of-date code. This will pull in the latest code.


charlesconnell:
I have signed the contributor license agreement

charlesconnell:
I suspect that the build failure is because GCC 9 is too old to build the latest BoringSSL code, but I haven't proven that yet

charlesconnell:
I've update the PR description and the associated ticket because I misunderstood the situation at first. The BoringSSL commit used in netty-tcnative is not that old, but still would benefit from being updated.

The build failure relates to a phenomenon described [here](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=51242#c32). Versions of GCC from 9.3 onward correctly do not produce warnings. However, I can't find an easy way to install that a GCC that new into CentOS 6.

normanmaurer:
@charlesconnell this was superseded by https://github.com/netty/netty-tcnative/pull/921. Thanks!

PR #924, opened 2025-05-13
Fix SSL.getSigAlgs() on BoringSSL for TLS 1.3

### LINKED ISSUE #923: Improper algorithm names returned by BoringSSL with TLS 1.3
`SSL.getSigAlgs()` currently returns incorrect values when backed by BoringSSL for TLS 1.3. Due to the second argument to [SSL_get_signature_algorithm_name](https://github.com/netty/netty-tcnative/blob/b552068c6cb3c18ac09b95a201c43acd9ac94a79/openssl-dynamic/src/main/c/ssl.c#L2586), the curve name is appended to the returned algorithm name.  This is different from all other cases (OpenSSL, or BoringSSL with TLS 1.2). This can cause downstream dependencies (such as [Netty's SignatureAlgorithmConverter](https://github.com/netty/netty/blob/4.2/handler/src/main/java/io/netty/handler/ssl/SignatureAlgorithmConverter.java)) to fail.

Example: The call `SSL.getSigAlgs()` will return `ecdsa_sha256_secp256r1` when in the BoringSSL with TLS 1.3 case and `ecdsa_sha256` in all other cases.

sfc-gh-grubin:
Always pass `false` as the second parameter to
`SSL_get_signature_algorithm_name()`. This forces BoringSSL to omit the curve name and return algorithm names in the style expected by callers of `SSL.getSigAlgs()`. Previously this parameter was only false for TLS 1.2.

This change does *not* have associated tests because I was unable to find any.

This is to fix #923

### DIFF
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -2583,7 +2583,7 @@ TCN_IMPLEMENT_CALL(jobjectArray, SSL, getSigAlgs)(TCN_STDARGS, jlong ssl) {
     }
 
     for (i = 0; i < num_peer_sigalgs; i++) {
-        if ((alg = SSL_get_signature_algorithm_name(peer_sigalgs[i], SSL_version(ssl_) != TLS1_2_VERSION)) == NULL) {
+        if ((alg = SSL_get_signature_algorithm_name(peer_sigalgs[i], false)) == NULL) {
             // The signature algorithm is not known to BoringSSL, skip it.
             continue;
         }


normanmaurer:
@chrisvest PTAL as well

sfc-gh-grubin:
Here are the relevant docs:
https://github.com/google/boringssl/blob/2a514a51baebd5a232fc64f7b082f7a8b28cd29d/include/openssl/ssl.h#L1154-L1158

```
// SSL_get_signature_algorithm_name returns a human-readable name for |sigalg|,
// or NULL if unknown. If |include_curve| is one, the curve for ECDSA algorithms
// is included as in TLS 1.3. Otherwise, it is excluded as in TLS 1.2.
OPENSSL_EXPORT const char *SSL_get_signature_algorithm_name(uint16_t sigalg,
                                                            int include_curve);
```

normanmaurer:
@sfc-gh-grubin did you sign our ICLA ? https://netty.io/s/icla

normanmaurer:
@sfc-gh-grubin tests is done in netty and we will add a test there.

sfc-gh-grubin:
A few quick clarifying questions: 

As per the [developer guide](https://netty.io/wiki/developer-guide.html), a CLA signature may not be needed. Do you need one?
> ... unless your contribution is trivial such as a single line change or a typo fix.

If you do need a CLA signed, is the individual CLA sufficient or do we need a corporate CLA signed?

Thank you.

sfc-gh-grubin:
> > A few quick clarifying questions:
> > As per the [developer guide](https://netty.io/wiki/developer-guide.html), a CLA signature may not be needed. Do you need one?
> 
> Actually you are right this patch is so small that we can skip it.

Let's skip it then.

chrisvest (approved):
Looks good!

normanmaurer:
@sfc-gh-grubin thanks!

PR #932, opened 2025-06-15
Obtain correct cipher in authentication_method(...)

normanmaurer:
Motivation:

This function returns the first cipher from the preference order, but is not necessarily the actual cipher that was negotiated.

Modifications:

- Use SSL_get_current_cipher or the struct directly to obtain the used cipher

Result:

authentication_method(...) returns the correct method.


normanmaurer:
This came up in https://github.com/netty/netty-tcnative/pull/929

skmcgrail (approved):

PR #936, opened 2025-07-16
Add missing <openssl/sha.h> include

davidben:
ssl_private.h depends on SHA_DIGEST_LENGTH, which is defined in <openssl/sha.h>. netty-tcnative is currently relying on the header getting transitively included elsewhere.

BoringSSL is looking to reduce transitive includes of <openssl/sha.h> to help some systems better audit uses of legacy algorithms like SHA-1. This change keeps netty-tcnative compatible with those changes.


normanmaurer:
Thanks a lot

PR #953, opened 2026-02-12
Open up SslCredential API to AWS-LC

jmcrawford45:
Motivation:

Allow users of AWS-LC to leverage the new SslCredential API.

Modifications:

Update docs and feature flags to allow SslCredential calls.

Result:

Users of AWS-LC can leverage the new SslCredential API.

Implements https://github.com/netty/netty/pull/15919#discussion_r2788573973


chrisvest:
The AWS-LC build failed.
@skmcgrail is the credentials API supported in AWS-LC?

normanmaurer:
I think we can close this for now... it is still not supported

PR #968, opened 2026-04-11
Add new method authenticationMethodsPacked to avoid string and array allocation

doom369:
Currently, `authenticationMethods()` allocates multiple strings and a string array during initial connection. This could be avoided with a new method, where we pack all auth methods into a single int. This will also simplify code on the Netty side (2 switch statements could be removed in `OpenSslKeyMaterialManager.setKeyMaterialServerSide`).

image

### DIFF
--- a/openssl-classes/src/main/java/io/netty/internal/tcnative/SSL.java
+++ b/openssl-classes/src/main/java/io/netty/internal/tcnative/SSL.java
@@ -673,6 +673,32 @@ public final class SSL {
     public static native String[] authenticationMethods(long ssl);
 
     /**
+     * Return the authentication methods packed into an {@code int}, avoiding {@code String[]} allocation.
+     * Each bit corresponds to an OpenSSL {@code SSL_a*} authentication constant (see sslutils.c):
+     * <p>
+     * Common (produced by all code paths):
+     * <ul>
+     *     <li>{@code 0x0001} (SSL_aRSA)   - RSA authentication (RSA, DHE_RSA, ECDHE_RSA)</li>
+     *     <li>{@code 0x0002} (SSL_aDSS)   - DSS authentication (DHE_DSS)</li>
+     *     <li>{@code 0x0004} (SSL_aNULL)  - no auth / anonymous (DH_anon, ECDH_anon)</li>
+     *     <li>{@code 0x0040} (SSL_aECDSA) - ECDSA authentication (ECDHE_ECDSA)</li>
+     * </ul>
+     * Legacy (only possible on OpenSSL &lt; 1.1.0):
+     * <ul>
+     *     <li>{@code 0x0008} (SSL_aDH)     - Fixed DH auth</li>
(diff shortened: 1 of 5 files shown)


openssl-classes/src/main/java/io/netty/internal/tcnative/SSL.java
```diff
@@ -672,6 +672,32 @@ public static void setTlsExtHostName(long ssl, String hostname) {
...
+    /**
+     * Return the authentication methods packed into an {@code int}, avoiding {@code String[]} allocation.
+     * Each bit corresponds to an OpenSSL {@code SSL_a*} authentication constant (see sslutils.c):
+     * <p>
+     * Common (produced by all code paths):
+     * <ul>
+     *     <li>{@code 0x0001} (SSL_aRSA)   - RSA authentication (RSA, DHE_RSA, ECDHE_RSA)</li>
+     *     <li>{@code 0x0002} (SSL_aDSS)   - DSS authentication (DHE_DSS)</li>
+     *     <li>{@code 0x0004} (SSL_aNULL)  - no auth / anonymous (DH_anon, ECDH_anon)</li>
+     *     <li>{@code 0x0040} (SSL_aECDSA) - ECDSA authentication (ECDHE_ECDSA)</li>
```
chrisvest, line 684: I'm wondering is if we should expose these constants through static fields in `SSL`, initialized from `NativeStaticallyReferencedJniMethods` calls.

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -1998,6 +1998,107 @@ TCN_IMPLEMENT_CALL(jobjectArray, SSL, authenticationMethods)(TCN_STDARGS, jlong
     return array;
 }
 
+/* Bits for algorithm_auth, same as defined in sslutils.c */
+#define TCN_SSL_aRSA    0x00000001L
+#define TCN_SSL_aDSS    0x00000002L
+#define TCN_SSL_aNULL   0x00000004L
+#define TCN_SSL_aECDSA  0x00000040L
+#define TCN_SSL_aALL    (TCN_SSL_aRSA | TCN_SSL_aDSS | TCN_SSL_aNULL | TCN_SSL_aECDSA)
```
chrisvest, line 2006: Not great to define them again. Maybe they can be moved to `ssl_private.h`?

openssl-classes/src/main/java/io/netty/internal/tcnative/SSL.java
doom369: It's a good idea actually, minimizes the chance for error.

openssl-dynamic/src/main/c/ssl.c
doom369: I'll check

normanmaurer:
@chrisvest PTAL again

chrisvest:
@doom369 Can you make the commits yourself instead of having Claude do it? We want to be able to keep track of where changes come from.

PR #971, opened 2026-04-13
Clear exception if happens during keylog_cb

normanmaurer:
Motivation:

If our java method throws in the keylog function we need to clear the exception as otherwise we might end up corrupting the JNI state

Modifications:

- Cheack if exception was thrown and if so clear it

Result:

No more risk of corrupted JNI state

### DIFF
--- a/openssl-dynamic/src/main/c/sslcontext.c
+++ b/openssl-dynamic/src/main/c/sslcontext.c
@@ -2670,6 +2670,10 @@ static void keylog_cb(const SSL* ssl, const char *line) {
     // Execute the java callback
     (*e)->CallVoidMethod(e, state->ctx->keylog_callback, state->ctx->keylog_callback_method,
                 P2J(ssl), outputLine);
+    // Clear the exception if any was thrown as otherwise we might corrupt the JNI state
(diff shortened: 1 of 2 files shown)


openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -2672,6 +2670,10 @@ static void keylog_cb(const SSL* ssl, const char *line) {
     // Execute the java callback
     (*e)->CallVoidMethod(e, state->ctx->keylog_callback, state->ctx->keylog_callback_method,
                 P2J(ssl), outputLine);
+    // Clear the exception if any was thrown as otherwise we might corrupt the JNI state
+    if ((*e)->ExceptionCheck(e) == JNI_TRUE) {
+        (*e)->ExceptionClear(e);
```
chrisvest, line 2675: This silently swallows the exception. We should update the javadocs in `KeyLogCallback` to note that the `handle` method should never throw, and if they do anyway the exception will be swallowed.

normanmaurer: done

PR #978, opened 2026-04-24
Revert "Correctly build boringssl-static for linux aarch64 (#975)"

normanmaurer:
This reverts commit 4939940102c1727467c7248b9a6c95ed03db45d8.


chrisvest:
What happened?

normanmaurer:
At the end this was not needed so we can revert it again

PR #980, opened 2026-05-07
Drop checks on SSL_CREDENTIAL support for older BoringSSL

### LINKED ISSUE #979: Weak symbols declarations are incompatible with latest BoringSSL
We started upgrading some of our functions to take `const` pointers, notably `SSL_CTX_add1_credential` and `SSL_add1_credential`. Normally turning an input pointer from `T*` to `const T*` is still source-compatible, but it looks like netty-tcnative redeclares some of BoringSSL's symbols as weak symbols:

https://github.com/netty/netty-tcnative/blob/25e07f6efbdad2e1051e15ee3bd13eff9c5be22a/openssl-dynamic/src/main/c/ssl_private.h#L526-L543

To avoid this, I suggest either:

* Drop support for 2023-era BoringSSL's? That's pretty old. (PS: This [documentation](https://boringssl.googlesource.com/boringssl/+/refs/heads/main/crypto/fipsmodule/FIPS.md#update-stream) might apply here.)

* If you _do_ still need to support old BoringSSLs, we have this `BORINGSSL_API_VERSION` marker.  Looking back at https://boringssl-review.googlesource.com/c/boringssl/+/66690, `BORINGSSL_API_VERSION >= 32` will check when we first added `SSL_CREDENTIAL`. If you want other APIs too, you might need to go a little further. But in general, the process is:
   * Find when we added the API
   * Find the next `BORINGSSL_API_VERSION` bump. If we foresaw that someone might want to condition on it, often we'll bump in the same CL. Sometimes we'll bump it sa litlt elater.

dingxiangfei2009:
Fix #979.

Given that `SSL_CREDENTIAL` has been in BoringSSL for a long time, we could move on and make these functions mandatory for BoringSSL-based builds.

### DIFF
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -2726,6 +2726,7 @@ TCN_IMPLEMENT_CALL(jlong, SSL, getSelectedCredential)(TCN_STDARGS, jlong ssl) {
     return (jlong)(intptr_t)credential;
 #else
     tcn_ThrowUnsupportedOperationException(e, "SSL_CREDENTIAL API not available.");
+    return 0;
 #endif
 }
 
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -2681,11 +2681,9 @@ TCN_IMPLEMENT_CALL(void, SSL, setRenegotiateMode)(TCN_STDARGS, jlong ssl, jint m
 }
 
 TCN_IMPLEMENT_CALL(void, SSL, addCredential)(TCN_STDARGS, jlong ssl, jlong cred) {
-    if (!check_credential_api(e)) return;
+#ifdef OPENSSL_IS_BORINGSSL
     SSL *ssl_ = J2P(ssl, SSL *);
     TCN_CHECK_NULL(ssl_, ssl, /* void */);
-
-#ifdef OPENSSL_IS_BORINGSSL
     SSL_CREDENTIAL* credential = (SSL_CREDENTIAL*)(intptr_t)cred;
     TCN_CHECK_NULL(credential, credential, /* void */);
 
@@ -2693,22 +2691,22 @@ TCN_IMPLEMENT_CALL(void, SSL, addCredential)(TCN_STDARGS, jlong ssl, jlong cred)
     if (result == 0) {
         tcn_Throw(e, "Failed to add credential to SSL");
     }
+#else
+    tcn_ThrowUnsupportedOperationException(e, "SSL_CREDENTIAL API not available.");
 #endif
 }
 
 TCN_IMPLEMENT_CALL(jlong, SSL, getSelectedCredential)(TCN_STDARGS, jlong ssl) {
-    if (!check_credential_api(e)) return 0;
+#ifdef OPENSSL_IS_BORINGSSL
     SSL *ssl_ = J2P(ssl, SSL *);
     TCN_CHECK_NULL(ssl_, ssl, 0);
-
-#ifdef OPENSSL_IS_BORINGSSL
     const SSL_CREDENTIAL* credential = SSL_get0_selected_credential(ssl_);
     if (credential == NULL) {
         return 0;
     }
     return (jlong)(intptr_t)credential;
 #else
-    return 0;  // Unreachable - check_credential_api throws
+    tcn_ThrowUnsupportedOperationException(e, "SSL_CREDENTIAL API not available.");
 #endif
 }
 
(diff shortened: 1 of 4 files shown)


normanmaurer (approved):

normanmaurer:
@dingxiangfei2009 did you sign our ICLA yet ? https://netty.io/s/icla

dingxiangfei2009:
Ah, I wasn't aware of CLA. I just completed the form.

dingxiangfei2009:
Oh yes, I must cc @davidben

openssl-dynamic/src/main/c/sslcredential.c
```diff
@@ -294,12 +302,11 @@ TCN_IMPLEMENT_CALL(jlong, SSLCredential, newDelegated)(TCN_STDARGS) {
     }
     return (jlong)(intptr_t)credential;
 #else
-    return 0;  // Unreachable - check_credential_api throws
+    tcn_ThrowUnsupportedOperationException(e, "SSL_CREDENTIAL API not available.");
 #endif
 }
 
 TCN_IMPLEMENT_CALL(void, SSLCredential, setDelegatedCredential)(TCN_STDARGS, jlong cred, jbyteArray dc) {
```
davidben, line 311: Not a question for this CL but OOC, do you all use delegated credentials? The original user of it apparently no longer cares about it, so I was looking at whether we even need to keep those APIs. (No big deal if you all care about it. It's not hugely in the way. Mostly curious.)

normanmaurer: No we don't use it...

normanmaurer (approved):

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -2700,34 +2700,32 @@ TCN_IMPLEMENT_CALL(void, SSL, setRenegotiateMode)(TCN_STDARGS, jlong ssl, jint m
...
-
-#ifdef OPENSSL_IS_BORINGSSL
     const SSL_CREDENTIAL* credential = SSL_get0_selected_credential(ssl_);
     if (credential == NULL) {
         return 0;
     }
     return (jlong)(intptr_t)credential;
 #else
-    return 0;  // Unreachable - check_credential_api throws
+    tcn_ThrowUnsupportedOperationException(e, "SSL_CREDENTIAL API not available.");
```
normanmaurer, line 2728:
```suggestion
    tcn_ThrowUnsupportedOperationException(e, "SSL_CREDENTIAL API not available.");
    return 0;
```

openssl-dynamic/src/main/c/sslcredential.c
```diff
@@ -43,38 +43,38 @@ static void throw_openssl_error(JNIEnv* env, const char* msg) {
...
 // Core SSL_CREDENTIAL functions
 TCN_IMPLEMENT_CALL(jlong, SSLCredential, newX509)(TCN_STDARGS) {
-    if (!check_credential_api(e)) return 0;
 #ifdef OPENSSL_IS_BORINGSSL
     SSL_CREDENTIAL* cred = SSL_CREDENTIAL_new_x509();
     TCN_CHECK_NULL(cred, credential, 0);
     return (jlong)(intptr_t)cred;
 #else
-    return 0;  // Unreachable - check_credential_api throws
+    tcn_ThrowUnsupportedOperationException(e, "SSL_CREDENTIAL API not available.");
```
normanmaurer, line 51:
```suggestion
    tcn_ThrowUnsupportedOperationException(e, "SSL_CREDENTIAL API not available.");
    return 0;
```

openssl-dynamic/src/main/c/sslcredential.c
```diff
@@ -294,12 +302,11 @@ TCN_IMPLEMENT_CALL(jlong, SSLCredential, newDelegated)(TCN_STDARGS) {
     }
     return (jlong)(intptr_t)credential;
 #else
-    return 0;  // Unreachable - check_credential_api throws
+    tcn_ThrowUnsupportedOperationException(e, "SSL_CREDENTIAL API not available.");
```
normanmaurer, line 306:
```suggestion
    tcn_ThrowUnsupportedOperationException(e, "SSL_CREDENTIAL API not available.");
    return 0;
```

openssl-dynamic/src/main/c/sslcredential.c
davidben: Ah great. How do you feel about removing the calls from netty-tcnative? Is that something we can just do? Would we want to leave the Java-side functions around as no-ops, or can those also be unwound? This would be `newDelegated` and `setDelegatedCredential`.

chrisvest: Since these are very recent additions, I _think_ it's fine to turn them into no-ops and deprecate them. Or better yet, make the method throw an exception.

PR #981, opened 2026-05-14
Restore linux-aarch64 artifacts to Maven Central publish

colleenkinross:
The linux-aarch64 artifacts were dropped starting in 2.0.72 when PR #926 migrated from nexus-staging-maven-plugin to central-publishing-maven-plugin.

Unlike other platforms for which we can build openssl-dynamic and boringssl-static in a single command, the aarch64 cross-compile requires two separate commands because openssl-dynamic must be built against a pre-compiled ARM version of APR and OpenSSL and boringssl-static doesn't have this dependency. The old plugin accepted -DaltStagingDirectory to write staging output outside of target/, so it survived the clean phase in the second command. The new plugin defaults to target/central-staging, which the clean phase in the second command deletes along with the rest of the root project's target/ directory.

Fix by passing -DoutputDirectory=/root/local-staging to both publish calls, mirroring the original -DaltStagingDirectory behavior, then copying the combined output into target/central-staging for the CI upload step.

PR #984, opened 2026-06-03
aarch64 release: openssl-dynamic and boringssl-static into single command

colleenkinross:
The two-command approach caused `clean` in the second build to wipe target/central-staging, losing the openssl-dynamic artifacts. 

Original fix attempt was https://github.com/netty/netty-tcnative/pull/981, however this did not work because the output directory wasn't *delete-able* which blocked the clean from the plugin. The attempt to resolve this was to create a subdirectory. After that change it became clear that the output directory was not used for artifacts, instead it was where the central_bundle.zip was placed which caused another release issue.

In the first fix, it was thought that we could not separate the commands because openssl-dynamic must be built against a pre-compiled ARM version of APR. However, this original reason for separate commands (conflicting -DaprArmHome values for each module) was removed in #906, so combining them is now safe. There are still differences in the flags passed, but these are ignored by boringssl, so should not cause an issue (-DopensslArmHome, -DaprArmHome).


normanmaurer:
Merged thanks…... also did cut the release. So I hope all is good now .

colleenkinross:
it worked!!!! :) thank you very much @normanmaurer !

PR #986, opened 2026-06-05
linux aarch_64 openssl-dynamic link to openssl3.x

colleenkinross:
Previously the aarch64 cross-compile was pinned to OpenSSL 1.1.1k, causing the linux-aarch_64-fedora JAR to require libssl.so.1.1 at runtime instead of libssl.so.3. This upgrades all three Linux build environments (centos6/x86_64, centos7/aarch64 cross-compile, debian/x86_64) to build OpenSSL 3.6.1 from source, with a single source of truth in pom.xml (opensslVersion property) propagated to CI and docker-compose. I also added cacheing to speed up the time to validate (but it's a clean build in release).

Mac was already updated to build with OpenSSL3.6 here: https://github.com/netty/netty-tcnative/pull/952 

There are several workarounds (listed below) because CentOS 6 and Debian 7 are EOL — updating the base images would eliminate them, which I've left out of scope here.
1. --no-check-certificate on the OpenSSL wget calls: CentOS 6 and Debian 7 ship with outdated CA bundles that cannot verify modern GitHub TLS certificates. This is consistent with the existing pattern already used for other wget downloads in these same Dockerfiles (ninja, GCC, APR).
2. custom build required for OpenSSL3.x since centos6/debian 7 since there's no OpenSSL 3.x binaries packaged for these distros as they are EOL. 
3. no-asm/no-apps when building OpenSSL3.x: 
  - no-asm used to *skip* all assembly generation and use pure C fallbacks which allows the old assemblers to handle it.
  - no-apps: skips building the CLI tools since we only need libssl.so and libcrypto.so
4. Added perl modules needed by OpenSSL3.x build system: perl-IPC-Cmd, perl-Time-Piece
5. Prevent APR from linking against libcrypt.so.1 as it's absent on AL2023+/RHEL9+ (replaced by libcrypt.so.2) and therefore prevents the dynamic linking functionality. We do this by setting ac_cv_search_crypt=no  - mirrors patches/apr_crypt.patch applied by the maven x86_64 build (remove crypt).

## Validation

Three Docker images rebuilt locally and validated:
- `netty-tcnative-centos:default` (x86_64-fedora): `libssl.so.3`, `libcrypto.so.3` ✓
- `netty-tcnative-centos:cross_compile_aarch64`: `libssl.so.3`, `libcrypto.so.3`, ELF64 AArch64 ✓
- `netty-tcnative-debian:default` (x86_64): `libssl.so.3`, `libcrypto.so.3` ✓

## This will break service for OS versions that don't have OpenSSL3.x. I've captured this below:

| Platform | x86_64 | aarch64 | Change |
|---|:---:|:---:|---|
| Amazon Linux 2023 | ✅  | ✅  | **Gains** linking (system has `libssl.so.3`) |
| RHEL 9.x / CentOS Stream 9 | ✅ | ✅ | **Gains** linking |
| Debian 12 / Ubuntu 22.04+ | ✅  | — | **Gains** linking |
| RHEL 8.x | — | ❌  | **Loses** aarch64 linking (system has `libssl.so.1.1`, not `.so.3`) |
| RHEL 8.x x86_64 / Ubuntu 20.04 | ❌  | — | No change (broken before, broken after) |


colleenkinross:
@normanmaurer it turns out the linux docker images were still built against OpenSSL1.x (only mac was upgraded to OpenSSL3.x). This is the change that switches this.

I've done a validation of the jar actually being used by our service to verify linking was to OpenSSL3.x (same way I saw that OpenSSL1 was still being used). 

This is quite urgent. I know it's a significant change. Please let me know as soon as you can whether this looks reasonable and on what time frame you could likely review/cut a release (if there are no big issues). I've tried to be as detailed as possible in the description.

colleenkinross:
@normanmaurer let me know if you want to chat about this :)

chrisvest:
Since we download the openssl tar.gz files without validating the TLS connections, it'd be good to add a steps where we extract the `opensslSha256` from the pom file as well, and check the downloaded tar.gz against it.

Apart from that, I think this looks good.

colleenkinross:
> Since we download the openssl tar.gz files without validating the TLS connections, it'd be good to add a steps where we extract the `opensslSha256` from the pom file as well, and check the downloaded tar.gz against it.
> 
> Apart from that, I think this looks good.

Good verification addition - thanks @chrisvest! I've made the change. If this looks good - when could you cut a new release?

openssl-dynamic/pom.xml
```diff
@@ -370,10 +373,6 @@
                       <property>aprArmHome</property>
                       <message>The folder of APR for aarch64 must be specified by hand. Please try -DaprArmHome=</message>
                     </requireProperty>
-                    <requireProperty>
-                      <property>opensslArmHome</property>
-                      <message>The folder of OpenSSL for aarch64 must be specified by hand. Please try -DopensslArmHome=</message>
-                    </requireProperty>
```
normanmaurer, line 376: Why was this removed ?

colleenkinross: The dockerfile now installs OpenSSL to a predictable path (since it's not present on the distros the package is built against) so I just moved it up and have the pom file compute it's default value. See line 335. The flag is no longer required

normanmaurer:
@chrisvest PTAL again

normanmaurer (approved):

chrisvest (approved):

colleenkinross:
thank you very much! Are you able to do another release @normanmaurer?

colleenkinross:
Hi @normanmaurer - do you have an ETA on the release cut? I'm sorry to keep bugging you. If you know a time it'll let us plan a workaround.

normanmaurer:
Done

PR #987, opened 2026-06-09
Clear certificate chain in setKeyMaterial.

### LINKED ISSUE #985: HelloRetryRequest with TLS 1.3 can result in duplicate intermediate certs when tcnative is used with OpenSSL
Hello!

I hit an issue in our production environment when enabling TLS 1.3 support on our web server. Our server certificate has an intermediate which we respond with in addition to the leaf cert. A subset of clients saw that they would receive duplicates of the same intermediate in the response during handshake. Most clients handle this gracefully but some clients throw certificate validation failures. On further investigation it happens consistently when the server responds with a HelloRetryRequest. This can happen when e.g. the client specifies a named group not supported by the server. TLS 1.2 connections remain unaffected.

I tried with tcnative using OpenSSL 3.x, OpenSSL 1.1.x, and BoringSSL. Out of those, BoringSSL is the only one that does not have this issue.

netty: 4.1.91.Final
tcnative: 2.0.78.Final

I have a minimal reproducer on https://github.com/cfolgere/netty-tls13-duplicate-hrr-issue/blob/main/src/test/java/com/example/nettytls13bug/Tls13HelloRetryDuplicateChainTest.java, tested on my M1. This test shows both the non-HRR and the HRR case and asserts that only two certificates are returned. The HRR case will fail with anything except BoringSSL.

I also have example pcaps on https://github.com/cfolgere/netty-tls13-duplicate-hrr-issue/tree/main/captures for both the successful case (no HRR sent) and the failing case (with HRR). You can clearly see the duplicate intermediate in the HRR case.

I'm opening the issue here as I believe the fix would ultimately be in tcnative.

mawiwill:
Fixes #985.

Netty calls SSL.setKeyMaterial in its server-side certificate callback (https://docs.openssl.org/3.6/man3/SSL_CTX_set_cert_cb/ but the same going back to 1.0.2). OpenSSL invokes certificate callbacks after the server receives a ClientHello.

TLS 1.2 TLS sessions usually had one ClientHello, so this callback would usually run once. TLS 1.3, however, might begin a session with multiple ClientHellos as the result of
HelloRetryRequests. SSL.setKeyMaterial adds chained certificates (intermediates and roots) to the per-connection SSL struct with SSL_add1_chain_certs. This function appends certificates, so every time a ClientHello arrives, setKeyMaterial would re-add the same chain. Finicky clients fail if they receive the same intermediate twice.

This change allows netty-tcnative to talk TLS 1.3 with these clients. It clears the SSL struct's certificate chain (that is, everything after the certificate assigned with SSL_use_certificate) on every setKeyMaterial call. Multiple ClientHellos will result in the same key meterial appearing in Certificate responses.

Note that this should improve correctness for TLS 1.2, as well, since renegotiation for e.g. rekeying would have suffered from the same duplicate intermediates issue.

### DIFF
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -2333,7 +2333,6 @@ TCN_IMPLEMENT_CALL(void, SSL, setKeyMaterial)(TCN_STDARGS, jlong ssl, jlong chai
         return;
     }
 
-
     // The first cert was loaded via SSL_use_certificate so skip it.
     for (i = 1; i < numCerts; ++i) {
 
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -2316,6 +2316,7 @@ TCN_IMPLEMENT_CALL(void, SSL, setKeyMaterial)(TCN_STDARGS, jlong ssl, jlong chai
         }
     }
 
+#if OPENSSL_VERSION_NUMBER >= 0x10002000L || defined(OPENSSL_IS_BORINGSSL) || defined(OPENSSL_IS_AWSLC)
     // Clear the chain; the loop below appends certificates to the
     // chain, and in TLS 1.3, this can be called again after the
     // server sends a HelloRetryRequest. Without clearing the
@@ -2332,6 +2333,7 @@ TCN_IMPLEMENT_CALL(void, SSL, setKeyMaterial)(TCN_STDARGS, jlong ssl, jlong chai
         }
         return;
     }
+#endif // OPENSSL_VERSION_NUMBER >= 0x10002000L || defined(OPENSSL_IS_BORINGSSL) || defined(OPENSSL_IS_AWSLC)
 
     // The first cert was loaded via SSL_use_certificate so skip it.
     for (i = 1; i < numCerts; ++i) {
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -2316,6 +2316,24 @@ TCN_IMPLEMENT_CALL(void, SSL, setKeyMaterial)(TCN_STDARGS, jlong ssl, jlong chai
         }
     }
 
+    // Clear the chain; the loop below appends certificates to the
+    // chain, and in TLS 1.3, this can be called again after the
+    // server sends a HelloRetryRequest. Without clearing the
+    // certificate chain that results in duplicate entries.
+    if (tcn_SSL_clear_chain_certs(ssl_) != 1) {
+        int errCode = ERR_get_error();
+        if (errCode == 0) {
+            tcn_Throw(e, "Could not clear certificate chain (unknown)");
+        } else {
+            ERR_error_string_n(errCode, err, ERR_LEN);
+            ERR_clear_error();
+
+            tcn_Throw(e, "Could not clear certificate chain (%s)", err);
+        }
+        return;
+    }
+
+
     // The first cert was loaded via SSL_use_certificate so skip it.
     for (i = 1; i < numCerts; ++i) {
 
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -2316,7 +2316,7 @@ TCN_IMPLEMENT_CALL(void, SSL, setKeyMaterial)(TCN_STDARGS, jlong ssl, jlong chai
         }
     }
 
-#if OPENSSL_VERSION_NUMBER >= 0x10002000L || defined(OPENSSL_IS_BORINGSSL) || defined(OPENSSL_IS_AWSLC)
+#if OPENSSL_VERSION_NUMBER >= 0x10002000L
     // Clear the chain; the loop below appends certificates to the
     // chain, and in TLS 1.3, this can be called again after the
     // server sends a HelloRetryRequest. Without clearing the
@@ -2333,7 +2333,7 @@ TCN_IMPLEMENT_CALL(void, SSL, setKeyMaterial)(TCN_STDARGS, jlong ssl, jlong chai
         }
         return;
     }
-#endif // OPENSSL_VERSION_NUMBER >= 0x10002000L || defined(OPENSSL_IS_BORINGSSL) || defined(OPENSSL_IS_AWSLC)
+#endif // OPENSSL_VERSION_NUMBER >= 0x10002000L
 
     // The first cert was loaded via SSL_use_certificate so skip it.
     for (i = 1; i < numCerts; ++i) {
--- a/openssl-dynamic/src/main/c/ssl.c
+++ b/openssl-dynamic/src/main/c/ssl.c
@@ -2320,7 +2320,7 @@ TCN_IMPLEMENT_CALL(void, SSL, setKeyMaterial)(TCN_STDARGS, jlong ssl, jlong chai
     // chain, and in TLS 1.3, this can be called again after the
     // server sends a HelloRetryRequest. Without clearing the
     // certificate chain that results in duplicate entries.
-    if (tcn_SSL_clear_chain_certs(ssl_) != 1) {
+    if (SSL_clear_chain_certs(ssl_) != 1) {
         int errCode = ERR_get_error();
         if (errCode == 0) {
             tcn_Throw(e, "Could not clear certificate chain (unknown)");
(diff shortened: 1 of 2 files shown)


openssl-dynamic/src/main/c/ssl_private.h
```diff
@@ -473,6 +473,7 @@ enum ssl_verify_result_t tcn_SSL_cert_custom_verify(SSL* ssl, uint8_t *out_alert
 #if !defined(OPENSSL_IS_BORINGSSL) && !defined(OPENSSL_IS_AWSLC)
 #define tcn_SSL_add1_chain_cert(ssl, x509) SSL_add1_chain_cert(ssl, x509)
 #define tcn_SSL_add0_chain_cert(ssl, x509) SSL_add0_chain_cert(ssl, x509)
+#define tcn_SSL_clear_chain_certs(ssl) SSL_clear_chain_certs(ssl)
```
normanmaurer: Why do we even need this and can't just call `SSL_clear_chain_certs` directly ?

mawiwill: Only to mimic the code that's there already. I'm happy to remove it!

normanmaurer: We only use our own defines if we need to use different APIs depending on which version / lib is used which seems to be not the case here and so we should just use `SSL_clear_chain_certs` directly.

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -2316,6 +2316,24 @@ TCN_IMPLEMENT_CALL(void, SSL, setKeyMaterial)(TCN_STDARGS, jlong ssl, jlong chai
...
+        } else {
+            ERR_error_string_n(errCode, err, ERR_LEN);
+            ERR_clear_error();
+
+            tcn_Throw(e, "Could not clear certificate chain (%s)", err);
+        }
+        return;
+    }
+
+
```
normanmaurer, line 2337:
```suggestion
```

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -2316,6 +2316,24 @@ TCN_IMPLEMENT_CALL(void, SSL, setKeyMaterial)(TCN_STDARGS, jlong ssl, jlong chai
         }
     }
 
+    // Clear the chain; the loop below appends certificates to the
+    // chain, and in TLS 1.3, this can be called again after the
+    // server sends a HelloRetryRequest. Without clearing the
+    // certificate chain that results in duplicate entries.
+    if (tcn_SSL_clear_chain_certs(ssl_) != 1) {
```
normanmaurer:
```suggestion
    if (SSL_clear_chain_certs(ssl_) != 1) {
```

openssl-dynamic/src/main/c/ssl_private.h
normanmaurer:
```suggestion
```

normanmaurer:
@mawiwill did you sign our ICLA yet ? https://netty.io/s/icla

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -2316,6 +2316,23 @@ TCN_IMPLEMENT_CALL(void, SSL, setKeyMaterial)(TCN_STDARGS, jlong ssl, jlong chai
         }
     }
 
+    // Clear the chain; the loop below appends certificates to the
+    // chain, and in TLS 1.3, this can be called again after the
+    // server sends a HelloRetryRequest. Without clearing the
+    // certificate chain that results in duplicate entries.
+    if (SSL_clear_chain_certs(ssl_) != 1) {
```
normanmaurer, line 2324: Seems like this is not supported in older Openssl versions so we need to put ifdef around it. This is fine imho as older openssl versions also not support TLS 1.3 anyway.

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -2316,6 +2316,23 @@ TCN_IMPLEMENT_CALL(void, SSL, setKeyMaterial)(TCN_STDARGS, jlong ssl, jlong chai
...
+            tcn_Throw(e, "Could not clear certificate chain (unknown)");
+        } else {
+            ERR_error_string_n(errCode, err, ERR_LEN);
+            ERR_clear_error();
+
+            tcn_Throw(e, "Could not clear certificate chain (%s)", err);
+        }
+        return;
+    }
+
```
normanmaurer, line 2337:
```suggestion
#if OPENSSL_VERSION_NUMBER >= 0x10002000L || defined(OPENSSL_IS_BORINGSSL) || defined(OPENSSL_IS_AWSLC)
    // Clear the chain; the loop below appends certificates to the
    // chain, and in TLS 1.3, this can be called again after the
    // server sends a HelloRetryRequest. Without clearing the
    // certificate chain that results in duplicate entries.
    if (SSL_clear_chain_certs(ssl_) != 1) {
        int errCode = ERR_get_error();
        if (errCode == 0) {
            tcn_Throw(e, "Could not clear certificate chain (unknown)");
        } else {
            ERR_error_string_n(errCode, err, ERR_LEN);
            ERR_clear_error();

            tcn_Throw(e, "Could not clear certificate chain (%s)", err);
        }
        return;
    }
#endif // OPENSSL_VERSION_NUMBER >= 0x10002000L || defined(OPENSSL_IS_BORINGSSL) || defined(OPENSSL_IS_AWSLC)

```

normanmaurer:
@mawiwill PTAL

mawiwill:
> ICLA

Just signed the ICLA!

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -2316,6 +2316,25 @@ TCN_IMPLEMENT_CALL(void, SSL, setKeyMaterial)(TCN_STDARGS, jlong ssl, jlong chai
         }
     }
 
+#if OPENSSL_VERSION_NUMBER >= 0x10002000L || defined(OPENSSL_IS_BORINGSSL) || defined(OPENSSL_IS_AWSLC)
```
chrisvest: If `defined(OPENSSL_IS_BORINGSSL) || defined(OPENSSL_IS_AWSLC)` is true, then we end up in the if-define on line 2299, instead of in the else-define here. So these must always be false at this location. Harmless, but confusing.

openssl-dynamic/src/main/c/ssl.c
mawiwill: Does netty-tcnative need to support OpenSSL < 1.0.2? The OpenSSL foundation claims that 1.0.1 and earlier are no longer supported, even with a support contract:
https://web.archive.org/web/20260607225346/https://openssl-library.org/policies/releasestrat/index.html

chrisvest (commented):
Just one small nit

openssl-dynamic/src/main/c/ssl.c
normanmaurer: A right... let me fix this

openssl-dynamic/src/main/c/ssl.c
```diff
@@ -2316,6 +2316,25 @@ TCN_IMPLEMENT_CALL(void, SSL, setKeyMaterial)(TCN_STDARGS, jlong ssl, jlong chai
...
+            tcn_Throw(e, "Could not clear certificate chain (unknown)");
+        } else {
+            ERR_error_string_n(errCode, err, ERR_LEN);
+            ERR_clear_error();
+
+            tcn_Throw(e, "Could not clear certificate chain (%s)", err);
+        }
+        return;
+    }
+#endif // OPENSSL_VERSION_NUMBER >= 0x10002000L || defined(OPENSSL_IS_BORINGSSL) || defined(OPENSSL_IS_AWSLC)
```
normanmaurer:
```suggestion
#if OPENSSL_VERSION_NUMBER >= 0x10002000L
    // Clear the chain; the loop below appends certificates to the
    // chain, and in TLS 1.3, this can be called again after the
    // server sends a HelloRetryRequest. Without clearing the
    // certificate chain that results in duplicate entries.
    if (SSL_clear_chain_certs(ssl_) != 1) {
        int errCode = ERR_get_error();
        if (errCode == 0) {
            tcn_Throw(e, "Could not clear certificate chain (unknown)");
        } else {
            ERR_error_string_n(errCode, err, ERR_LEN);
            ERR_clear_error();

            tcn_Throw(e, "Could not clear certificate chain (%s)", err);
        }
        return;
    }
#endif // OPENSSL_VERSION_NUMBER >= 0x10002000L
```

chrisvest (approved):

mawiwill:
> @mawiwill PTAL

This definitely works for me! I have a minor question about support OpenSSL versions, mostly out of curiosity. Thanks for your quick work on the PR, I really appreciate it!

normanmaurer:
@mawiwill thanks a lot!

PR #989, opened 2026-06-26
Fix linux-x86_64 openssl-dynamic to link against OpenSSL 3.x

colleenkinross:
The previous fix (PR #986) built OpenSSL 3.x from source and symlinked /usr/local/ssl -> /opt/openssl-$VERSION, expecting tcnative's configure to find it via the default search path. This didn't work: the native build was still picking up the system OpenSSL 1.0.x.

Fix: pass -DopensslHome explicitly to every build command in the centos6 and debian7 docker-compose files, and add a Maven profile in openssl-dynamic/pom.xml that activates on -DopensslHome and forwards --with-ssl to configure. Remove the now-unnecessary symlink.

Verified the RPATH for each to verify fixed:
  │ linux-x86_64-fedora (centos6)   │ libssl.so.3               │ /opt/openssl-3.6.1/lib       │   ✅   │
  │ linux-aarch_64-fedora (centos7) │ libssl.so.3               │ /opt/openssl-3.6.1-share/lib │   ✅   │
  │ linux-x86_64 (debian7)          │ libssl.so.3               │ /opt/openssl-3.6.1/lib       │   ✅

### DIFF
--- a/docker/Dockerfile.centos6
+++ b/docker/Dockerfile.centos6
@@ -66,8 +66,7 @@ RUN set -x && \
   # no-asm: devtoolset-9 on CentOS 6 cannot reliably compile OpenSSL's hand-tuned x86_64 assembly
   ./Configure linux-x86_64 --prefix=/opt/openssl-$OPENSSL_VERSION --libdir=lib shared no-asm no-apps && \
   make -j1 install_sw) && \
-  rm -rf openssl-$OPENSSL_VERSION openssl-$OPENSSL_VERSION.tar.gz && \
-  ln -sf /opt/openssl-$OPENSSL_VERSION /usr/local/ssl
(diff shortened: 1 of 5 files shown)


colleenkinross:
@chrisvest can you review ASAP and released?

Another emergency i'm sorry.

colleenkinross:
@normanmaurer

super super urgent (launching GA tuesday). let me know if you're able to review and release ASAP

chrisvest (approved):

colleenkinross:
thank you very much @chrisvest !

@normanmaurer could you cut a release? The biggest emergency is mitigated on our side, but we would still like OpenSSL3.x linking across x86-64 as quickly as reasonable. Let me know if you need anything to proceed!

normanmaurer:
@colleenkinross done

colleenkinross:
thank you very much!!!!! @normanmaurer and @chrisvest !!!

PR #993, opened 2026-07-29
Fix data race by setting info callback on SSL instance

kannanjgithub:
Set the SSL info callback on the specific SSL instance rather than the shared SSL context. This avoids concurrent modification/access of the shared context during connection establishment and handshake processing, which was causing data races.

### DIFF
--- a/openssl-dynamic/src/main/c/sslcontext.c
+++ b/openssl-dynamic/src/main/c/sslcontext.c
@@ -182,14 +182,6 @@ static apr_status_t ssl_context_cleanup(void *data)
     return APR_SUCCESS;
 }
 
-void ssl_info_callback(const SSL *ssl, int where, int ret) {
-    tcn_ssl_state_t* state = NULL;
-    if (0 != (where & SSL_CB_HANDSHAKE_START)) {
-        if ((state = tcn_SSL_get_app_state(ssl)) != NULL) {
-            state->handshakeCount++;
-        }
-    }
-}
 
 /* Initialize server context */
 TCN_IMPLEMENT_CALL(jlong, SSLContext, make)(TCN_STDARGS, jint protocol, jint mode)
--- a/openssl-dynamic/src/main/c/sslcontext.c
+++ b/openssl-dynamic/src/main/c/sslcontext.c
@@ -182,7 +182,7 @@ static apr_status_t ssl_context_cleanup(void *data)
     return APR_SUCCESS;
 }
 
-static void ssl_info_callback(const SSL *ssl, int where, int ret) {
+void ssl_info_callback(const SSL *ssl, int where, int ret) {
     tcn_ssl_state_t* state = NULL;
     if (0 != (where & SSL_CB_HANDSHAKE_START)) {
         if ((state = tcn_SSL_get_app_state(ssl)) != NULL) {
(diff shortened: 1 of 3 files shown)


normanmaurer:
@kannanjgithub did you sign our ICLA ? https://netty.io/s/icla

openssl-dynamic/src/main/c/sslcontext.c
```diff
@@ -182,7 +182,7 @@ static apr_status_t ssl_context_cleanup(void *data)
     return APR_SUCCESS;
 }
 
-static void ssl_info_callback(const SSL *ssl, int where, int ret) {
+void ssl_info_callback(const SSL *ssl, int where, int ret) {
```
normanmaurer: consider just keep the static and move the whole function to ssl.c. This way you also don't need any changes in the ssl_private.h file.

normanmaurer:
@kannanjgithub actually I fail to see what is the problem here... Can you explain why this would cause a data race ?

kannanjgithub:
> @kannanjgithub actually I fail to see what is the problem here... Can you explain why this would cause a data race ?

In the previous code, the registration call was placed inside newSSL (in ssl.c), which is executed every time a new connection is established:
```
// Inside newSSL (called for every connection)
SSL_CTX_set_info_callback(c->ctx, ssl_info_callback);
```
Even though different threads are writing the exact same value (the address of ssl_info_callback), they are performing unsynchronized concurrent writes to the same memory location in the shared SSL_CTX structure. Concurrent unsynchronized writes to the same memory location constitute a data race, which can lead to undefined behavior, memory corruption, or CPU cache inconsistency issues. This got flagged by ThreadSanitizer in tcnative code imported into Google.

openssl-dynamic/src/main/c/sslcontext.c
kannanjgithub: Done.

normanmaurer:
> > @kannanjgithub actually I fail to see what is the problem here... Can you explain why this would cause a data race ?
> 
> In the previous code, the registration call was placed inside newSSL (in ssl.c), which is executed every time a new connection is established:
> 
> ```
> // Inside newSSL (called for every connection)
> SSL_CTX_set_info_callback(c->ctx, ssl_info_callback);
> ```
> 
> Even though different threads are writing the exact same value (the address of ssl_info_callback), they are performing unsynchronized concurrent writes to the same memory location in the shared SSL_CTX structure. Concurrent unsynchronized writes to the same memory location constitute a data race, which can lead to undefined behavior, memory corruption, or CPU cache inconsistency issues. This got flagged by ThreadSanitizer in tcnative code imported into Google.

huh... This is called in SSLContext.make(...) so it will only be done once and not on each new SSL instance creation.  That's why I am confused.

normanmaurer:
@kannanjgithub maybe you use a version which does not have https://github.com/netty/netty-tcnative/commit/f47c053af5d297317c636298b160d77304090c4b included yet ?

kannanjgithub:
@normanmaurer Yes, you are exactly right, we don't have this commit imported into Google yet (It is still on netty-tcnative 2.0.75.Final)

normanmaurer:
So can we close this ?

PR #1001, opened 2026-08-28
Fix double DeleteLocalRef call which results in UB

normanmaurer:
Motivation:

A double DeleteLocalRef on the same local reference is present in both the OpenSSL and BoringSSL cert-verify callbacks in sslcontext.c. The NETTY_JNI_UTIL_DELETE_LOCAL macro does not null out the pointer after deleting
, so a second delete on the same stale handle is real UB per the JNI spec — it can corrupt the local-ref table and crash later in an unrelated JNI call.

This regression was introduces by https://github.com/netty/netty-tcnative/commit/990751e766b5d4bf697a4a9fb6b45e66bf415b8d

Modifications:

Remove deletion of local ref when its already done

Result:

No more UB


normanmaurer:
@chrisvest also did add some safety guards in https://github.com/netty/netty-jni-util/commit/5ba8032687bfca41a2b40dd2d70a6cc2f19ad207.

PR #1008, opened 2026-09-08
fix: preserve musl compatibility in FIPS build

pe-bernard:
## Why

[#997](https://github.com/netty/netty-tcnative/pull/997) added the musl compatibility check to the FIPS `native-jar` execution, but did not port the corresponding link and post-link handling from the Linux release profiles. As a result, FIPS artifacts retain glibc loader and runtime dependencies that make them fail the new check.

## Changes

- link the C++ runtime and unwinder archives statically in `fips-boringssl-static`
- remove the architecture-specific `ld-linux` `DT_NEEDED` entry from the final FIPS native library before the musl check
- document that FIPS, x86_64, and aarch64 profiles each duplicate these compatibility settings

## Verification

- `xmllint --noout boringssl-static/pom.xml`
- `git diff --check`
- downstream Linux FIPS builds completed successfully on x86_64 and aarch64, including the native musl compatibility check

Local Maven model validation was not run because this machine has no Java runtime.

### DIFF
--- a/boringssl-static/pom.xml
+++ b/boringssl-static/pom.xml
@@ -313,6 +313,21 @@
                       </then>
                     </if>
 
+                    <!-- Same ld-linux DT_NEEDED removal as the release profiles.
+                         See docs/musl-compatibility.md -->
+                    <if>
+                      <equals arg1="${os.detected.name}" arg2="linux" />
+                      <then>
+                        <exec executable="patchelf" failonerror="true" dir="${nativeLibOnlyDir}/META-INF/native/linux${archBits}/" resolveexecutable="true">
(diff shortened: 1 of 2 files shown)


fredericgermain:
Ah a FIPS variant. I didn't see that one. I'm testing FIPS in CI in many variants rn, using the pinned fips commit in https://boringssl.googlesource.com/boringssl.

There could be quite a lot of subtilities with FIPS and compiler flags.

That's something you'd be interested to have in CI? It could be part of work I did on the CI builds, but it'd need some time to finalise. 

Thanks Pierre-Eloi for your PR!

PR #1009, opened 2026-09-09
Build the FIPS profile in CI, and make it build again

fredericgermain:
Follow-up to #997 and #1008. Draft: the amd64 validation of the final state is still running, see "Verification".

## Motivation

#1008 showed the blind spot: nothing in CI builds the FIPS profile, so a change to the release profiles that is not ported to it only fails on someone's downstream build. Building it in CI turned up more than the missing port: the profile no longer builds on a fresh machine at all (the FIPS tarball bucket is private now), does not link on aarch64, pinned a module that holds no certificate, and called clang-12 "the validated toolchain" when the current certificate names clang 17.

## What is in here (one commit each)

1. **patchelf in the Debian, Arch and openSUSE dev images.** Their `build` service has failed since #997; CI only runs Debian's dynamic-only service.
2. **FIPS profile.** BoringSSL is checked out from git at the pinned sha, which is what [FIPS.md](https://boringssl.googlesource.com/boringssl/+/refs/heads/main/crypto/fipsmodule/FIPS.md) recommends anyway (the tarball at `commondatastorage.googleapis.com/chromium-boringssl-fips` returns 403 now); `CMAKE_POSITION_INDEPENDENT_CODE` is back (dropped in #950, breaks the aarch64 link); ninja builds only `crypto ssl decrepit bssl`; the pin moves to `fips-20240805`, the module of certificate #5244 (validated 2026-04-18), the newest one that holds a certificate; the compiler is a property (`fipsCC`/`fipsCXX`, default clang-17, the version in that certificate's security policy).
3. **musl on modern toolchains.** Built on a current distro the artifact imports glibc-only symbols from code we do not compile with our flags (fortify `__*_chk`, `__libc_single_threaded`, glibc 2.38's `__isoc23_*` redirects, `_dl_find_object`): it loads on Alpine but `ldd` fails. `-U_FORTIFY_SOURCE` on the Linux compile lines and weak fallbacks in `musl_compat.c`, following the existing pattern. No-op on the CentOS release images.
4. **CI.** A Debian 13 image carrying the build environment from #5244's policy (clang 17.0.6, ninja 1.12.1 from the archive; go 1.22.3, cmake 3.29.3 at those versions), legs `debian13-x86_64` and `debian13-x86_64-fips`, and bare-Alpine `musl-verify` legs on both jars. The FIPS one matters most: the module's integrity check and self-tests run in an ELF constructor at `dlopen`, so only a real load proves the post-link patchelf left it intact.

## Verification

- amd64: the same tree, minus the pin change, passed the full PR workflow on my fork (all 17 jobs). The run with the 2024-08-05 pin and matching toolchain is in progress locally; the PR stays a draft until it is green.
- arm64: both profiles built natively on Debian 13, zero musl-check warnings, both jars load on bare Alpine.

## Open points

- Which module to pin is a policy decision. `fips-20251031` (previous pin) and `fips-20260721` both build with this profile via `-DboringsslBranch=<sha>`; neither holds a certificate yet.
- The fallbacks in commit 3 add a few exported weak symbols to the release artifacts, which never reference them. CentOS 6 and 7 builds pass unchanged.
- The Debian 13 image builds on JDK 21 (`--release 8` keeps the class files at 8); trixie has no JDK 8.

## For discussion: is a certified pin the right default?

- The only BoringCrypto module with a certificate today is the 2024-08-05 one (#5244, issued 2026-04-18): source that is two years old, which contradicts ordinary patch hygiene. The three 2025 modules are still in the CMVP queue ("Comment Resolution" on the [modules-in-process list](https://csrc.nist.gov/projects/cryptographic-module-validation-program/modules-in-process/modules-in-process-list)), 10 to 20 months after their version dates.
- The CMVP's stated goal, as relayed at the [OpenSSL Conference in October 2025](https://archive.openssl-conference.org/2025/presentations/Jason_Lawlor_OpenSSL_Presentation_JL_October_5_2025.pdf), was a six-month queue by the end of 2025 and no queue by mid-2026. That is not what the BoringCrypto entries show.
- [FIPS.md](https://boringssl.googlesource.com/boringssl/+/refs/heads/main/crypto/fipsmodule/FIPS.md) now documents an "update stream": `main`, built with `-DFIPS=1`, following the [FedRAMP policy of January 2025](https://www.fedramp.gov/rev5/fips/) that asks vendors to promote update streams over validated module streams. On that stream `FIPS_version()` returns 0.
- So the profile can serve two different needs: a certified pin for whoever must cite a certificate, and the update stream for whoever wants current crypto with FIPS-mode self-tests. This PR makes the certified module the default and keeps the other one a `-DboringsslBranch` away; the reverse default is equally defensible and is worth deciding explicitly.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -40,6 +40,16 @@ jobs:
           - setup: al2023-x86_64-aws_lc
             docker-compose-run: "-f docker/docker-compose.al2023.yaml run build"
             docker-bake-args: "-f docker-compose.al2023.yaml"
+          # boringssl-static on a current toolchain, and the only CI leg that builds the FIPS
+          # profile (it needs clang-17; see docker/Dockerfile.debian13). Neither is a release
+          # artifact. Before these legs the FIPS profile had no CI at all, so a change to the
+          # release profiles that was not ported to it went unnoticed until a downstream build.
+          - setup: debian13-x86_64
+            docker-compose-run: "-f docker/docker-compose.debian-13.yaml run build"
+            docker-bake-args: "-f docker-compose.debian-13.yaml"
+          - setup: debian13-x86_64-fips
+            docker-compose-run: "-f docker/docker-compose.debian-13.yaml run build-fips"
+            docker-bake-args: "-f docker-compose.debian-13.yaml"
 
     name: ${{ matrix.setup }}
     steps:
@@ -203,6 +213,28 @@ jobs:
             variant: gcompat
             extra-pkgs: gcompat
             drop-elftools: "0"
+            build-service: runtime-setup
+            run-service: verify
+          # The Debian 13-built jars: the FIPS artifact is the one that matters (its power-on
+          # self-test and integrity check run inside an ELF constructor at dlopen, so only a
+          # real load proves the patchelf'd library is still intact), and the default-profile
+          # one shows what a modern-gcc build looks like on musl. bare x86_64 only: aarch64
+          # would need an arm64 FIPS build leg, and the gcompat/nolibgcc variants add nothing
+          # the CentOS 6 legs do not already establish.
+          - setup: alpine-x86_64-fips
+            os: ubuntu-24.04
+            jars: build-debian13-x86_64-fips-jars
+            variant: bare
+            extra-pkgs: ""
+            drop-elftools: "1"
+            build-service: runtime-setup
+            run-service: verify
+          - setup: alpine-x86_64-debian13
+            os: ubuntu-24.04
+            jars: build-debian13-x86_64-jars
+            variant: bare
+            extra-pkgs: ""
+            drop-elftools: "1"
             build-service: runtime-setup
             run-service: verify
           - setup: glibc-control-x86_64
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -38,6 +38,16 @@ jobs:
           - setup: al2023-x86_64-aws_lc
             docker-compose-run: "-f docker/docker-compose.al2023.yaml run build"
             docker-bake-args: "-f docker-compose.al2023.yaml"
+          # boringssl-static on a current toolchain, and the only CI leg that builds the FIPS
+          # profile (it needs clang-17; see docker/Dockerfile.debian13). Neither is a release
+          # artifact. Before these legs the FIPS profile had no CI at all, so a change to the
+          # release profiles that was not ported to it went unnoticed until a downstream build.
+          - setup: debian13-x86_64
+            docker-compose-run: "-f docker/docker-compose.debian-13.yaml run build"
+            docker-bake-args: "-f docker-compose.debian-13.yaml"
+          - setup: debian13-x86_64-fips
+            docker-compose-run: "-f docker/docker-compose.debian-13.yaml run build-fips"
+            docker-bake-args: "-f docker-compose.debian-13.yaml"
 
     name: ${{ matrix.setup }}
     steps:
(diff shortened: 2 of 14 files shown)


fredericgermain:
Still WIP, the last validation run is in progress. I'll come back at it tonight. Not critical for a release IMHO, but nice to have a better support for FIPS in latest release

- A few non-critical fixes on top of this: the aarch64 link needs CMAKE_POSITION_INDEPENDENT_CODE back (dropped in #950), and a bare ninja builds all 564 BoringSSL targets where four are used.
- The FIPS source tarballs on commondatastorage.googleapis.com are down (403 on every object, 401 on the bucket), so the profile now clones the git repo at the pinned sha, which is what FIPS.md recommends anyway. Downstream builds only kept working off the download plugin's cache.
- Reverting the pin to the 2024-08-05 module (fips-20240805), the one that actually holds a certificate (#5244, issued April 2026); fips-20251031 is still in the CMVP queue. Same idea for the toolchain: clang 17.0.6, go 1.22.3, ninja 1.12.1, cmake 3.29.3 as named in that certificate's security policy, instead of clang-12. Any newer fips-YYYYMMDD head still builds with -DboringsslBranch=<sha>. Three newer SSL_CREDENTIAL setters had to be gated on BORINGSSL_API_VERSION for that module.
- CI builds it on Debian 13 rather than Ubuntu: its archive carries exactly that clang and ninja, and it matches the other dev images.
- Built on a toolchain that modern, the artifact picks up glibc-only imports (fortify, the glibc 2.38 __isoc23_* redirects, and friends) that make ldd fail on Alpine even though it loads. Handled in musl_compat.c; it applies to anyone building this profile on a current Ubuntu or Debian.

For discussion: a certified pin means shipping two-year-old crypto source, which contradicts normal patch hygiene. The 2025 modules sit at "Comment Resolution" on NIST's in-process list 10 to 20 months after their version dates, and the CMVP's stated goal of a six-month queue by end of 2025 and none by mid-2026 (OpenSSL Conference, Oct 2025) is not happening. FIPS.md now documents an update stream, main with -DFIPS=1, per the FedRAMP policy.
There are also some 2026 fips tag in the git repo, but it feels they didn't bother to submit them yet.

normanmaurer:
@fredericgermain let me know once this is ready

fredericgermain:
Hi @normanmaurer,

Let me try to finish that this week.

fredericgermain:
@normanmaurer this is ready for review.

- Rebased on current main. The Debian 13 compose file follows the ghcr.io build-cache convention from #1015 (tag `debian13`).
- Verified on amd64 for both the normal boringssl-static build and the FIPS-mode one: zero musl-check warnings, and both jars load on bare Alpine, the FIPS one also with gcompat.
- Reference for the FIPS side: Google's [FIPS.md](https://boringssl.googlesource.com/boringssl/+/refs/heads/main/crypto/fipsmodule/FIPS.md).

It took a few turns to get here, so here is the path, in case the end result looks arbitrary:

1. **First I tried to support the validated 2024 module** (`fips-20240805`, certificate #5244), with the exact toolchain from its security policy. It builds, but tcnative calls three `SSL_CREDENTIAL` functions that do not exist in that two-year-old tree, so they had to be compiled out behind `BORINGSSL_API_VERSION` checks. Carrying compatibility patches for an old BoringSSL is not a direction I think we want, so I scrapped it.
2. **Then I followed FIPS.md literally:** build `main` with `-DFIPS=1` and the latest stable tools. A floating branch plus floating tools is too many moving parts for a maintainer to support. I have not bought into all of these security practices myself, and I did not want to hand you a CI leg that breaks whenever Google moves.
3. **So everything is pinned, while still following FIPS.md in spirit:**
   - BoringSSL is pinned to the head of Google's newest FIPS branch, `fips-20260721`. It is from July, recent enough. If Google still cuts these branches, presumably they are meant to be used. It may deserve a chat with their team about the current workflow: I do not think FIPS.md is up to date, its link to the FedRAMP policy is dead.
   - The CI image is Debian stable (13), pinned by image digest and by a snapshot.debian.org timestamp, so CI gets no surprises.
   - Go is the one tool that cannot come from Debian: that BoringSSL branch needs Go 1.25.8 and Debian ships 1.24. It is downloaded at a pinned version with a checksum. One more moving part, unfortunately.

On musl: built on a modern distro the artifact needs a few glibc-only fallbacks in `musl_compat.c`, which I kept to the measured minimum; the CentOS release images are unaffected. I also looked at linking libstdc++ dynamically; static, like the other targets, looks better.

That FIPS build is quite close to something I would consider taking to production. The one thing I am not sure about is whether the compilation differences (clang, FIPS mode, the delocated module) have a bad impact on performance. I plan to run some A/B benchmarks against the normal build, hopefully I find the time.

Two things for discussion:

1. **How official this looks.** The `-fips` artifactId and the `BoringSSL-FIPS-Compliant: true` manifest entry (#945) can read as if Netty as a whole provides FIPS compliance, and a CI job may make it look more official still. What the profile produces is a FIPS-mode build that contains BoringCrypto; the binary holds no certificate. FedRAMP's own policy asks vendors not to use terms like "FIPS compliant". I left the name alone because it changes what downstream builders consume. Let me know in review if I should suffix it with `-boringcrypto` instead, or something else.
2. **I would recommend simplifying the CI.** The `-fedora` variant has no purpose any more: it existed for the OpenSSL 1.0 soname difference, and now that everything links `libssl.so.3` the two Linux artifacts are the same library built on two EOL images. I would probably recommend Debian 13, or another modern distro, as the base builder, with a crosstool-ng toolchain if we want to keep the glibc floor low. Happy to open a separate issue with the details.

The commits are self-contained if you would rather take the Arch and openSUSE patchelf fix separately.

